PR #11, opened 2017-10-06
Updated to use latest swagger in documentation rather than copy

michael-bryson:


docs/pom.xml
```diff
@@ -35,6 +35,20 @@
         <maven.install.skip>true</maven.install.skip>
         <skipNexusStagingDeployMojo>true</skipNexusStagingDeployMojo>
     </properties>
+    
+    <dependencies>
+        <dependency>
+            <groupId>com.github.jobservice</groupId>
+            <artifactId>job-service-contract</artifactId>
+            <version>${project.version}</version>
```
dermot-hardy, line 43: @tonymcveigh, I don't have a problem with this but did you tell me there was a problem with it (i.e. using `${project.version}` in a dependency section)?

dermot-hardy: Thanks build machine.  BTW have you become self-aware?

PR #13, opened 2017-10-13
CAF-3645: DB changes for job dependencies feature

mulhollandc:
I have included new tables and function changes for the new job dependencies feature.
I have needed to back out recent copyright changes as this was causing upgrade issues. I'll raise a separate ticket to re-visit this.

### DIFF
--- a/job-service-db/pom.xml
+++ b/job-service-db/pom.xml
@@ -30,6 +30,10 @@
         <version>2.3.0-SNAPSHOT</version>
     </parent>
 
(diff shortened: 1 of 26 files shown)


glucy:
Hi @mulhollandc I've taken a look over you changes, but I am afraid the final Stored Procedure totally lost me.   If you think it is best to get these changes in so we can start building Java Code, (and test code) around these changes and test them fully that way I am OK with that approach.

PR #37, opened 2017-12-11
CAF-3904: Job Tracking Worker concurrency issues

mulhollandc:
Proposed changes to address concurrency related exceptions reported in the JIRA ticket (see https://jira.autonomy.com/browse/CAF-3904). In summary:
1. The Job Tracking Worker has been changed to support catch/retry when concurrency related exceptions are detected. These can arise if multiple transactions attempt to create tables and indexes at the same time.
2. The ReportProgress DB function has been modified to take an exclusive table lock on the top most dynamic task table to prevent concurrent updates. This fix was needed to protect a different scenario I spotted during additional testing where the a job had been left in Active state even though ALL sub-tasks where completed.
3. Additional DB code was changed to allow postgres to chose a name for indexes created for the dynamic task tables.

### DIFF
--- a/worker-jobtracking/src/main/java/com/hpe/caf/worker/jobtracking/JobTrackingWorkerReporter.java
+++ b/worker-jobtracking/src/main/java/com/hpe/caf/worker/jobtracking/JobTrackingWorkerReporter.java
@@ -103,22 +103,16 @@ public class JobTrackingWorkerReporter implements JobTrackingReporter {
 
                 //  Allow for retries in the event that the source of the error is from concurrent sessions
                 //  attempting table and/or index creation at the same time.
-                if (retryCount++ == maxRetries) {
+                if (retryCount++ < maxRetries &&
+                        (se.getMessage().contains("duplicate key value violates unique constraint") ||
+                                se.getMessage().matches("(?s).*(relation|type).*already exists.*"))) {
+                    LOG.info(MessageFormat.format("Retrying reportJobTaskProgress() call for job task {0}. Retry count {1}.",
+                            jobTaskId, retryCount));
+                } else {
                     LOG.error(MessageFormat.format(errorMessage, jobTaskId, se.getMessage()));
                     throw new JobReportingException(
                             MessageFormat.format(errorMessage, jobTaskId,
                                     se.getMessage()), se);
-                } else {
-                    if (se.getMessage().contains("duplicate key value violates unique constraint") ||
-                            se.getMessage().matches("(?s).*(relation|type).*already exists.*")) {
-                        LOG.info(MessageFormat.format("Retrying reportJobTaskProgress() call for job task {0}. Retry count {1}.",
-                                jobTaskId, retryCount));
-                    } else {
-                        LOG.error(MessageFormat.format(errorMessage, jobTaskId, se.getMessage()));
-                        throw new JobReportingException(
-                                MessageFormat.format(errorMessage, jobTaskId,
-                                        se.getMessage()), se);
-                    }
                 }
             }
         }
--- a/worker-jobtracking/src/main/java/com/hpe/caf/worker/jobtracking/JobTrackingWorkerReporter.java
+++ b/worker-jobtracking/src/main/java/com/hpe/caf/worker/jobtracking/JobTrackingWorkerReporter.java
@@ -89,12 +89,40 @@ public class JobTrackingWorkerReporter implements JobTrackingReporter {
      */
     @Override
     public void reportJobTaskProgress(final String jobTaskId, final int estimatedPercentageCompleted) throws JobReportingException {
-        try (Connection conn = getConnection()) {
-            //TODO - FUTURE: pass estimatedPercentageCompleted to the database function
-            report(conn, jobTaskId, JobStatus.Active);
-        } catch (SQLException se) {
-            throw new JobReportingException(MessageFormat.format("Failed to report the progress of job task {0}. {1}", jobTaskId, se.getMessage()), se);
+
+        int retryCount = 0;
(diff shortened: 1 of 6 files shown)


mulhollandc:
Developer build passing - see http://sou-jenkins2.hpeswlab.net:8080/job/JobService/view/Developer/job/JobService~job-service~CAF-3904_Errors_Reporting_Progress~CI/

worker-jobtracking/src/main/java/com/hpe/caf/worker/jobtracking/JobTrackingWorkerReporter.java
```diff
@@ -89,12 +89,40 @@ public JobTrackingWorkerReporter() throws JobReportingException {
...
+                //TODO - FUTURE: pass estimatedPercentageCompleted to the database function
+                report(conn, jobTaskId, JobStatus.Active);
+                break;
+            } catch (final SQLException se) {
+                final String errorMessage = "Failed to report the progress of job task {0}. {1}";
+
+                //  Allow for retries in the event that the source of the error is from concurrent sessions
+                //  attempting table and/or index creation at the same time.
+                if (retryCount++ == maxRetries) {
+                    LOG.error(MessageFormat.format(errorMessage, jobTaskId, se.getMessage()));
```
dermot-hardy: Will this be logged where the exception is caught?  We don't want to double or triple log the same events.

worker-jobtracking/src/main/java/com/hpe/caf/worker/jobtracking/JobTrackingWorkerReporter.java
```diff
@@ -89,12 +89,40 @@ public JobTrackingWorkerReporter() throws JobReportingException {
...
     @Override
     public void reportJobTaskProgress(final String jobTaskId, final int estimatedPercentageCompleted) throws JobReportingException {
-        try (Connection conn = getConnection()) {
-            //TODO - FUTURE: pass estimatedPercentageCompleted to the database function
-            report(conn, jobTaskId, JobStatus.Active);
-        } catch (SQLException se) {
-            throw new JobReportingException(MessageFormat.format("Failed to report the progress of job task {0}. {1}", jobTaskId, se.getMessage()), se);
+
+        int retryCount = 0;
+        final int maxRetries = 5;
```
dermot-hardy: This seems arbitrary.  I know we originally thought we would only need one retry and it would always work the second time.  Why did that turn out not to be the case again?

worker-jobtracking/src/main/java/com/hpe/caf/worker/jobtracking/JobTrackingWorkerReporter.java
```diff
@@ -89,12 +89,40 @@ public JobTrackingWorkerReporter() throws JobReportingException {
...
+                } else {
+                    if (se.getMessage().contains("duplicate key value violates unique constraint") ||
+                            se.getMessage().matches("(?s).*(relation|type).*already exists.*")) {
+                        LOG.info(MessageFormat.format("Retrying reportJobTaskProgress() call for job task {0}. Retry count {1}.",
+                                jobTaskId, retryCount));
+                    } else {
+                        LOG.error(MessageFormat.format(errorMessage, jobTaskId, se.getMessage()));
+                        throw new JobReportingException(
+                                MessageFormat.format(errorMessage, jobTaskId,
+                                        se.getMessage()), se);
```
dermot-hardy: This block is just the same as the block above, so you should merge the `if` statements - i.e. make the `if` something like:

    if (retryCount++ < maxRetries && (se.getMessage().contains(...) || se.getMessage().matches(...))) {
        // Retry
        ...
    } else {
        // Fail
        ...
    }

dermot-hardy (changes requested):

worker-jobtracking/src/main/java/com/hpe/caf/worker/jobtracking/JobTrackingWorkerReporter.java
mulhollandc: A combination of a couple of things I think. The use of a long JobId (e.g. '7cf26494bc00400_1B2XGydNQ5K_wv1XLsWmW') and many nested sub-tasks results in Postgres truncating the index name. This combined with additional set of Job Tracking Workers caused the previous retry count of 1 to be exceeded. I have no control over when the key violation occurs in the first place as it all depends on what each transaction is doing at that time. I have only managed to repro this once since it happened but a 2nd retry happen to resolve it. I have simply bumped this up in case it occurs again. If you can't think of an alternative solution given there is no predictability when this will happen, then let me know.

worker-jobtracking/src/main/java/com/hpe/caf/worker/jobtracking/JobTrackingWorkerReporter.java
mulhollandc: This is the only place the exception is caught. The error message will only be logged whenever either the retry count has exceeded OR the error message contains something other than what I expect.

worker-jobtracking/src/main/java/com/hpe/caf/worker/jobtracking/JobTrackingWorkerReporter.java
mulhollandc: Yes you are absolutely correct. I'll address this.

mulhollandc:
@dermot-hardy Code review change made.

mulhollandc:
@dermot-hardy Final (hopefully) changes made.

dermot-hardy (approved):

PR #42, opened 2018-01-12
CAF-3920: Corrected issue with Docker container not reporting as unhealthy

CJComac:
https://jira.autonomy.com/browse/CAF-3920

Returning 500 error in the healthcheck response if contact to RabbitMQ or the DB resulted in an error

### DIFF
--- a/job-service-admin/src/main/java/com/hpe/caf/services/admin/HealthCheck.java
+++ b/job-service-admin/src/main/java/com/hpe/caf/services/admin/HealthCheck.java
@@ -42,8 +42,6 @@ public class HealthCheck extends HttpServlet
 
     private final static Logger LOG = LoggerFactory.getLogger(HealthCheck.class);
 
-    private static boolean unhealthyWorker = false;
-
     @Override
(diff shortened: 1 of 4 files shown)


job-service-admin/src/main/java/com/hpe/caf/services/admin/HealthCheck.java
```diff
@@ -42,10 +42,11 @@
 
     private final static Logger LOG = LoggerFactory.getLogger(HealthCheck.class);
 
+    private static boolean unhealthyWorker = false;
```
dermot-hardy: As discussed, use return values so that you don't need this `static`.

dermot-hardy (changes requested):

job-service-admin/src/main/java/com/hpe/caf/services/admin/HealthCheck.java
CJComac: Adjusted in the latest commit.

dermot-hardy (approved):

PR #44, opened 2018-02-16
SCMOD-3285: Updated base images

Anthony-McGreevy:


Anthony-McGreevy:
@dermot-hardy Merging this PR as it is such an insignificant change.

PR #46, opened 2018-03-08
Updated email addresses and removed Conal from team pages

michael-bryson:


docs/_config.yml
```diff
@@ -1,5 +1,5 @@
 title: Job Service
-email: caf@hpe.com
+email: job-service@microfocus.com
```
dermot-hardy: Create a PDL for this:
https://myitsupport.ext.hpe.com/SEITSArticle?ArticleNumber=000001160

Use 'jobservice@hpe.com' (I don't think we can create Micro Focus PDLs yet) and make sure to enable it for external access from the internet (using ATM - https://atm.atl.hp.com/main.asp).

dermot-hardy (changes requested):

docs/_config.yml
michael-bryson: PDL jobservice@hpe.com has been created.

PR #54, opened 2018-06-21
SCMOD-4372: Updated to produce no response

Anthony-McGreevy:
https://jira.autonomy.com/browse/SCMOD-4372


dermot-hardy (approved):
Change looks good but we need a Jira for this

dermot-hardy:
We're probably going to want to update default queue name and configuration in tandem with this.

Anthony-McGreevy:
SCMOD-4372 has been raised to cover this issue.

Anthony-McGreevy:
workers default output queue is the workers name plus -err or just worker-err if the workers name cannot be found.

dermot-hardy (approved):

PR #65, opened 2018-08-09
SCMOD-4348: Deadlock fix

Anthony-McGreevy:
https://jira.autonomy.com/browse/SCMOD-4348


Anthony-McGreevy:
@dermot-hardy  I'm not entirely sure the changelog updates are strictly required because I can see they haven't been updated for the last few database procedure updates but I figured they should be maintained.

dermot-hardy (approved):

PR #66, opened 2018-08-13
For review

Anthony-McGreevy:


Anthony-McGreevy:
@dermot-hardy This is the branch I am working with currently, my SQL changes are here, they won't be merged in this branch, it is only being used for testing.

PR #67, opened 2018-08-13
SCMOD-4843: Updated to create all parent tables and insert rows on completion

Anthony-McGreevy:
https://jira.autonomy.com/browse/SCMOD-4843


Anthony-McGreevy:
@dermot-hardy I have split out all of the work on the pgsql code into this commit.

dermot-hardy (approved):

PR #69, opened 2018-08-14
SCMOD-4658: Install job-service database as part of container startup

omkumar-mariappan:
https://jira.autonomy.com/browse/SCMOD-4658


job-service-db/pom.xml
```diff
@@ -34,13 +34,15 @@
         <dependency>
             <groupId>com.github.cafapi.util</groupId>
             <artifactId>util-liquibase-installer</artifactId>
+            <version>1.16.0-SNAPSHOT</version>
```
Anthony-McGreevy, line 37: Is there something from this version that was needed? This version must have previously been managed from a parent project using dependency management, does this need changed instead? Same with the plugin version below.

Anthony-McGreevy (approved):

job-service-db/pom.xml
dermot-hardy: I agreed - better to update whatever `caf-common` is referenced.

dermot-hardy (changes requested):

job-service-db/pom.xml
omkumar-mariappan: Plugin version was needed to pull the 1.16.0-SNAPSHOP Version. Without plugin change, it was pulling the 1.15.0-1.  Refactored the code in `util-liquibase-installer` to take full connection URL, that's the way we are passing from marathon template. For that to work, I had to make code change.  Otherwise we had to make code other repos to make it work with 1.15.0-1

omkumar-mariappan:
@dermot-hardy JobService does nt have direct reference to `caf-common`,  parent of this repo is `caf-common-parent`.. do you know how caf-common gets referenced?. If i change it to 1.16.0-SNAPSHOT, then i dont need to make this pom file change

dermot-hardy (approved):

dermot-hardy:
It will be coming in through the worker-framework reference.  I'm fine with it the way it is btw.

omkumar-mariappan:
Ok, i can see it now. I will keep the current changes in the pom. And add some comments saying that when we upgrade workerframework, we can remove the line from the pom

dermot-hardy (approved):

PR #76, opened 2018-10-26
SCMOD-4489: Cleanup of commented code

rahul-kulkarni73:
https://autjira.microfocus.com/browse/SCMOD-4489

### DIFF
--- a/job-service-acceptance-tests/src/integration-test/java/com/hpe/caf/jobservice/acceptance/JobServiceEndToEndIT.java
+++ b/job-service-acceptance-tests/src/integration-test/java/com/hpe/caf/jobservice/acceptance/JobServiceEndToEndIT.java
@@ -130,11 +130,6 @@ public class JobServiceEndToEndIT {
         testItemAssetIds = generateWorkerBatch();
     }
 
-//    @Test
-//    public void testJobCompletionWithTaskDataObject() throws Exception {
-//        testJobCompletion(true);
-//    }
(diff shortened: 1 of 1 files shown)


job-service-acceptance-tests/src/integration-test/java/com/hpe/caf/jobservice/acceptance/JobServiceEndToEndIT.java
```diff
@@ -130,11 +130,6 @@ public void testSetup() throws Exception {
         testItemAssetIds = generateWorkerBatch();
     }
 
-//    @Test
-//    public void testJobCompletionWithTaskDataObject() throws Exception {
-//        testJobCompletion(true);
-//    }
-
```
michael-bryson: I did not previously comment on this section so why are you now deleting this? Is this something that should be enabled in future?

michael-bryson (changes requested):

job-service-acceptance-tests/src/integration-test/java/com/hpe/caf/jobservice/acceptance/JobServiceEndToEndIT.java
rahul-kulkarni73: Added it back, was there before. So not sure if its going to be used.

michael-bryson (approved):

PR #78, opened 2019-01-16
SCMOD-5725: Added Kubernetes quick start deployment

michael-bryson:
Also removed the docker-compose deployment files.

https://autjira.microfocus.com/browse/SCMOD-5725


deploy/src/main/resources/production-kubernetes-prerequisites/README.md
```diff
@@ -0,0 +1,32 @@
+# Production Kubernetes Prerequisites
+
+The Production Kubernetes Prerequisites deployment supports the deployment of the Job Service Database on Kubernetes. This folder contains the kubernetes environment and template files that are required to deploy the Job Service Database and RabbitMQ.
+
+## Service Configuration
+
+### Kubernetes Template
+The `jobservice-prerequisites-deployment.yaml` template file describes the marathon deployment information required for starting the Job Service Database and RabbitMQ. The template file uses property substitution to get values for configurable properties **required** for service deployment. These properties are configured in the kubernetes environment file `kubernetes.env`.
```
dermot-hardy: "describes the **marathon** deployment information"

deploy/src/main/resources/production-kubernetes-prerequisites/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+- `JOB_SERVICE_DB_PASSWORD`: The password for the Postgres database.
+
+## Service Deployment
+
+To deploy the persistent volume, issue the following command from the `production-kubernetes-prerequisites` directory:
+
+    kubectl create -f jobservice-prerequisites-persistentvolumeclaim.yaml
+
+To deploy the database and RabbitMQ, issue the following command from the `production-kubernetes-prerequisites` directory:
```
dermot-hardy: I think there should be an "edit the kubernetes.env file" step, or perhaps better still the file should have sensible defaults so that it doesn't have to be edited.

deploy/src/main/resources/production-kubernetes-prerequisites/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Service Configuration
+
+### Kubernetes Template
+The `jobservice-prerequisites-deployment.yaml` template file describes the marathon deployment information required for starting the Job Service Database and RabbitMQ. The template file uses property substitution to get values for configurable properties **required** for service deployment. These properties are configured in the kubernetes environment file `kubernetes.env`.
+
+The `jobservice-prerequisites-persistentvolumeclaim.yaml` template file describes the Kubernetes persistent volume  information required by the Job Service database and RabbitMQ.
+
+### Kubernetes Environment
+The `kubernetes.env` file supports configurable property settings necessary for service deployment. These include:
```
dermot-hardy: This is not all that is in that file.

dermot-hardy (changes requested):

deploy/src/main/resources/production-kubernetes-prerequisites/README.md
michael-bryson: Copy paste mistake from marathon file, fixed now.

deploy/src/main/resources/production-kubernetes-prerequisites/README.md
michael-bryson: Added the edit step, not sure defaults works great with so called production files. In any case the edit step would still need to be there even if defaults were used.

deploy/src/main/resources/production-kubernetes-prerequisites/README.md
michael-bryson: Added rabbit variables that were missing.

deploy/src/main/resources-filtered/production-kubernetes-prerequisites/jobservice-prerequisites-deployment.yaml
```diff
@@ -0,0 +1,68 @@
...
+        - name: POSTGRES_USER
+          value: "${JOB_SERVICE_DB_USER}"
+        resources: {}
+        volumeMounts:
+          - mountPath: /var/lib/postgresql/data
+            name: job-service-db
+        ports:
+        - containerPort: 5432
+          hostPort: ${JOB_SERVICE_DB_PORT}
+      restartPolicy: Always
```
rado-straka: "Always" is default value, doesn't have to be here. 
Remove in all other occurrences, too.

deploy/src/main/resources-filtered/production-kubernetes-prerequisites/jobservice-prerequisites-deployment.yaml
```diff
@@ -0,0 +1,68 @@
...
+      volumes:
+      - name: job-service-db
+        persistentVolumeClaim:
+          claimName: job-service-db
+status: {}
```
rado-straka: empty, not mandatory
Remove in all other occurrences, too.

deploy/src/main/resources-filtered/production-kubernetes-prerequisites/jobservice-prerequisites-deployment.yaml
```diff
@@ -0,0 +1,68 @@
...
+        image: rabbitmq:3-management
+        resources: {}
+        volumeMounts:
+        - mountPath: /var/lib/rabbitmq
+          name: rabbitmq
+        ports:
+        - containerPort: 5672
+          hostPort: ${CAF_RABBITMQ_PORT}
+        - containerPort: 15672
+          hostPort: ${CAF_RABBITMQ_MANAGEMENT_PORT}
```
rado-straka: good catch, I missed that and used services to define ports forwarding, but will change to this as it is simpler and enough for this examples

deploy/src/main/resources-filtered/production-kubernetes-testing/jobservice-testing-deployment.yaml
```diff
@@ -0,0 +1,92 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: worker-globfilter
+spec:
+  replicas: 1
+  strategy: {}
```
rado-straka: Probably just mistake, but use strategy.type: Recreate like you did above.
By default the value is RollingUpdate(or something like that) and it can make service unavailable while "updating" and cause failure of dependent services.

Update in all other occurrences, too.

deploy/src/main/resources-filtered/production-kubernetes-testing/jobservice-testing-deployment.yaml
```diff
@@ -0,0 +1,92 @@
...
+  template:
+    metadata:
+      creationTimestamp: null
```
rado-straka: Remove if null and in all other occurrences, too.

deploy/src/main/resources-filtered/production-kubernetes-testing/jobservice-testing-deployment.yaml
```diff
@@ -0,0 +1,92 @@
...
+          - mountPath: /mnt/caf-datastore-root
+            name: worker-datastore
+          - mountPath: /mnt/caf-worker-input-dir
+            name: job-service-demo-input-dir
+      volumes:
+      - name: worker-datastore
+        persistentVolumeClaim:
+          claimName: worker-datastore
+      - name: job-service-demo-input-dir
+        hostPath:
```
rado-straka: Why did you used different type of volume here? It is fine if you want to leave it , but you can remove all persistent volumes and persistent volume claims and use just this hostpath. In this case it is the same, because we use only one node, so data still will be on the same node(using more nodes in cluster we would need to use pv and pvc).

deploy/src/main/resources-filtered/production-kubernetes-testing/jobservice-testing-deployment.yaml
```diff
@@ -0,0 +1,92 @@
...
+          path: ${JOB_SERVICE_DEMO_INPUT_DIR}
```
rado-straka: Do not know how is that folder created, but if you can't guarantee that it exists on the host, you should use "type: DirectoryOrCreate" to create it if missing.

deploy/src/main/resources-filtered/production-kubernetes/jobservice-deployment.yaml
```diff
@@ -0,0 +1,142 @@
...
+  template:
+    metadata:
+      creationTimestamp: null
+      labels:
+        io.kompose.service: jobservice
+    spec:
+      containers:
+      - name: jobservice
+        image: jobservice/job-service:${project.version}
+        imagePullPolicy: Never
```
rado-straka: Why "Never"? Update also in all other occurrences if you will do.

deploy/src/main/resources-filtered/production-kubernetes-prerequisites/jobservice-prerequisites-deployment.yaml
```diff
@@ -0,0 +1,68 @@
...
+    spec:
+      containers:
+      - name: jobservicedb
+        image: postgres:9.6
+        env:
+        - name: POSTGRES_PASSWORD
+          value: "${JOB_SERVICE_DB_PASSWORD}"
+        - name: POSTGRES_USER
+          value: "${JOB_SERVICE_DB_USER}"
+        resources: {}
```
rado-straka: empty, not mandatory
Remove in all other occurrences, too.

rado-straka (changes requested):
Consider as I commented at one point to use hostpath volume instead of pv and pvc volumes, although I see that it would be also documentation change, so it is up to you as it is generally correct solution just makes this more complex.

deploy/src/main/resources-filtered/production-kubernetes-prerequisites/jobservice-prerequisites-deployment.yaml
michael-bryson: Removed

deploy/src/main/resources-filtered/production-kubernetes-prerequisites/jobservice-prerequisites-deployment.yaml
michael-bryson: Removed

deploy/src/main/resources-filtered/production-kubernetes-testing/jobservice-testing-deployment.yaml
michael-bryson: Added strategy.type: Recreate to all service where it was empty.

deploy/src/main/resources-filtered/production-kubernetes-testing/jobservice-testing-deployment.yaml
michael-bryson: Added to ${maven.build.timestamp} where it was null.

deploy/src/main/resources-filtered/production-kubernetes-testing/jobservice-testing-deployment.yaml
michael-bryson: @rado-straka Not sure, the persistent volume was created by the kompose convert but the other was not added by kompose. I was trying to recreate similar to what I saw in the marathon template so not sure if this is right or not. What is the best solution here?

deploy/src/main/resources-filtered/production-kubernetes-testing/jobservice-testing-deployment.yaml
michael-bryson: Added

deploy/src/main/resources-filtered/production-kubernetes/jobservice-deployment.yaml
michael-bryson: This was a left over from my testing as I couldn't pull snapshot images so manually pulled the images and set this to never. Changed it to Always throughout.

deploy/src/main/resources-filtered/production-kubernetes-prerequisites/jobservice-prerequisites-deployment.yaml
michael-bryson: Removed

michael-bryson:
> Consider as I commented at one point to use hostpath volume instead of pv and pvc volumes, although I see that it would be also documentation change, so it is up to you as it is generally correct solution just makes this more complex.

@rado-straka I really do not know the correct way to go about configuring these volumes so any advice would be appreciated.

deploy/src/main/resources-filtered/production-kubernetes-testing/jobservice-testing-deployment.yaml
rado-straka: Instead of using pv and pvc, this can be used:
```
           volumeMounts:
        - mountPath: /srv/common/webdav
          name: webdav
      volumes:
      - name: webdav
        hostPath:
          path: /tmp/hostpath_pv/webdav
          type: DirectoryOrCreate
```
This use of volumes is usually used for development not production as it is briefer.
Also, in this case when there is one node(minikube) and one cluster, there is no advantage using pv and pvc.
There is nothing wrong with your solution, my only reason for the comment was to show that it would make less of config files, but no need to change anything.

rado-straka (approved):

deploy/src/main/resources/README.md
```diff
@@ -48,180 +46,28 @@ The Docker Compose file contains the following services:
 
     By default the output folder used is `./output-files`, but a different folder can be used by setting the `JOB_SERVICE_DEMO_OUTPUT_DIR` environment variable.
 
-## Usage
```
dermot-hardy: As discussed, we're losing too much stuff here.  For me, the main purpose of this 'deploy' branch is the quick-start.

dermot-hardy (changes requested):

michael-bryson:
@dermot-hardy Production kubernetes files removed and quick start converted to kubernetes.

deploy/src/main/resources/kubernetes.env
```diff
@@ -0,0 +1,29 @@
+# The following properties MUST be configured before the Job Service can be deployed.
```
dermot-hardy: Why do all of these have to be configured?  Can we not have sensible defaults for most of them and/or aren't a lot of them internal?

dermot-hardy (changes requested):

deploy/src/main/resources/kubernetes.env
michael-bryson: I have no idea how to set them as defaults in the templates.

deploy/src/main/resources/README.md
```diff
@@ -10,106 +10,168 @@ The Job Service is a RESTful Web Service and provides a simple API.  It tracks t
...
-      </tr>
-    </table>
+	Deploy the database and RabbitMQ by issuing the following command from the directory where you have downloaded the files to:
+
+    ```
+	source ./kubernetes.env \
+            ; cat jobservice-deployment.yaml \
+            | perl -pe 's/\$\{(\w+)\}/(exists $ENV{$1} && length $ENV{$1} > 0 ? $ENV{$1} : "NOT_SET_$1")/eg' \
+            | kubectl create -f -
+	```
```
dermot-hardy: We had to do something like this for Marathon, but doesn't the `kubectl` command have in-built facilities for this?  What does `kubectl apply` do for example?

deploy/src/main/resources/README.md
```diff
@@ -10,106 +10,168 @@ The Job Service is a RESTful Web Service and provides a simple API.  It tracks t
...
+5. Navigate to the Job Service UI  
     The Job Service is a RESTful Web Service and is primarily intended for programmatic access, however it also ships with a Swagger-generated user-interface.
 
     Using a browser, navigate to the `/job-service-ui` endpoint on the Job Service:  
 
-        http://docker-host:9411/job-service-ui
+        http://<docker-host>:<JOB_SERVICE_8080_SERVICE_PORT>/job-service-ui
 
-    Adjust 'docker-host' to be the name of your own Docker Host and adjust the port if you are not using the default.
+    Adjust `docker-host` to be the name of your own Docker Host and adjust the `JOB_SERVICE_8080_SERVICE_PORT` to match what you set it to in the `kubernetes.env` file.
```
dermot-hardy: Is this the right terminology for Kubernetes?  Do we even have access to the Docker Host?

dermot-hardy (changes requested):

deploy/src/main/resources/kubernetes.env
dermot-hardy: It feels like it is too much to ask people to fill out for a quick-start, esp. without a lot of guidance w.r.t. what the values should be.

michael-bryson: I appreciate that but I do not know how to set them as defaults.

dermot-hardy: @rado-straka Do you know how to do this?

deploy/src/main/resources/README.md
michael-bryson: The help states "Apply a configuration to a resource by filename or stdin."

At this stage I do not know enough about the kubectl to implement this. I just copied the same method as marathon as it works. I could possibly look into other nicer methods but I do not have time to do this before the end of sprint.

deploy/src/main/resources/README.md
michael-bryson: Not sure, I used the IP address of my minikube machine when testing. However what this value would be for other kubernetes systems I don't know.

deploy/src/main/resources/kubernetes.env
rado-straka: I use services in kubernetes, so for example rabbitmq can be called by its name and using port which are defaults in code:
logs from familyhashing:
```
 INFO  [2019-01-22 15:56:47,575] net.jodah.lyra.internal.ConnectionHandler: Creating connection cxn-1 to [rabbitmq:5672]
INFO  [2019-01-22 15:56:47,665] net.jodah.lyra.internal.ConnectionHandler: Created connection cxn-1 to amqp://10.101.108.140:5672/ 
```
servcie definition is:
```
apiVersion: v1
kind: Service
metadata:
  labels:
    io.kompose.service: rabbitmq
  name: rabbitmq
spec:
  ports:
  - name: "rabbitmq-ctrl-port"
    port: 15672
    #if targetPort is not set then targetPort==port 
    #targetPort: 15672
  - name: "rabbitmq-node-port"
    port: 5672
    #targetPort: 5672
  selector:
    io.kompose.service: rabbitmq
```
so this service is called by its name "rabbitmq" and pointing to the rabbitmq pod what is done by selector section which must be the same as the label in rabbitmq pod.
username and password for rabbitmq also should work with defaults in code.

For other env variables it would need to check what are defaults for some of them, and if they match with setup of other services.

michael-bryson:
@dermot-hardy I have updated the kubernetes files to make it easier to use with no property replacement so can you take a look?

deploy/src/main/resources/README.md
```diff
@@ -10,106 +10,85 @@ The Job Service is a RESTful Web Service and provides a simple API.  It tracks t
...
-2. Job Service Database  
-    Internally the Job Service uses a PostgreSQL Database to store the Job Status information.  When the stack is started a `job-service-db` volume will be created to store the database files.
-
-3. RabbitMQ  
+2. RabbitMQ  
     The Worker Framework is a pluggable infrastructure and technically it can use different messaging systems.  However it is most common for RabbitMQ to be used for messaging, and that is what is used here.
 
-4. Job Tracking Worker  
+3. Job Tracking Worker  
     For simplicity the Job Tracking Worker is not shown on the diagram above.  The diagram shows messages passing directly between the workers, but in reality the messages are passed through the Job Tracking Worker, which acts as a proxy for them.  It routes them to their intended destination but it also updates the Job Service Database with the progress.  This means that the Job Service is able to provide accurate progress reports when they are requested.
```
dermot-hardy: It's not an issue with this PR but raise a Jira for this paragraph - it is no longer accurate w.r.t. how the Job Tracking Worker works.

deploy/src/main/resources/README.md
```diff
@@ -10,106 +10,85 @@ The Job Service is a RESTful Web Service and provides a simple API.  It tracks t
...
 
-6. GlobFilter Worker  
+5. GlobFilter Worker  
     This is a simple worker developed just for this demonstration.  It is a Batch Worker which takes in a glob-pattern as the Batch Definition.  Glob-patterns are generally fairly simple.  For example, `*.txt` means "all text files in the input folder".  Even more complex patterns like `**/t*.txt`, which means "all text files which start with the letter 't' and are in the input folder or in any subfolders of the input folder", are fairly easy to understand.  The worker produces a separate task for each file which matches the glob-pattern.
+    
+    The input folder has been set to `/job-service-test/input-files` for the purposes of this demonstration.  This directory will be created on your Kubernetes Cluster when the Job Service components have been deployed.
```
dermot-hardy, line 40: Working through it, I still don't feel we've got this right.  If there is a cluster of machines I don't think we know which one the input files will be read from and which the output files will be written to.

I want to introduce a simple web-based file server into the deployment - this one looks like it might fit the bill:
https://docs.filebrowser.xyz/installation#docker

Then we can update the usage instructions to have the user upload the sample files, and download the result files.

dermot-hardy (changes requested):

deploy/src/main/resources/README.md
michael-bryson: https://autjira.microfocus.com/browse/SCMOD-5886 logged.

deploy/src/main/resources/README.md
michael-bryson: FileBrowser added to the deployment along with usage instructions updated.

michael-bryson:
@dermot-hardy Latest comments addressed by adding filebrowser to the deployment and usage instructions updated.

dermot-hardy (approved):
I'm happy with this as it is but as discussed it might be safer to split out the deployment into two to allow some of the services to come up first.

michael-bryson:
I have added a readinessProbe to the postgres container instead which appears to do the job. I will merge this once the developer build passes.

dermot-hardy:
OK

PR #86, opened 2019-03-28
SCMOD-6182: add last-update-time to job

joseph-lansdowne-opentext:
In the `JobTrackingWorkerIT` tests, I decided not to extend `JobReportingExpectation` to handle the new assertions - hopefully it's clear why from the nature of the assertions.

- ticket: https://autjira.microfocus.com/browse/SCMOD-6182
- build: http://sou-jenkins2.hpeswlab.net/job/JobService/job/JobService~job-service~SCMOD-6182~CI/


job-service-db/src/main/resources/v2.7.0/changelog.xml
```diff
@@ -105,4 +105,47 @@
...
+                         schemaName="public">
+        </createProcedure>
+    </changeSet>
+    <changeSet id="update_procedure_internal_report_task_status_last_update_time" runOnChange="true" author="Gamma Team">
+        <createProcedure path="procedures/reportTaskStatus.sql"
+                         procedureName="internal_report_task_status"
+                         schemaName="public">
+        </createProcedure>
+    </changeSet>
+    <changeSet id="update_job_table_add_column_last_update_time" author="Gamma Team">
```
dermot-hardy, line 144: Does this not need to come before the procedure updates or does it not matter?

job-service/src/main/java/com/hpe/caf/services/job/api/generated/model/Job.java
```diff
@@ -229,14 +246,15 @@ public boolean equals(Object o) {
                 Objects.equals(description, job.description) &&
                 Objects.equals(externalData, job.externalData) &&
                 Objects.equals(createTime, job.createTime) &&
-                Objects.equals(status, job.status) &&
+        Objects.equals(this.lastUpdateTime, job.lastUpdateTime) &&
+        Objects.equals(status, job.status) &&
```
dermot-hardy: Some sort of whitespace screw up here.

job-service/src/main/java/com/hpe/caf/services/job/api/generated/model/Job.java
```diff
@@ -229,14 +246,15 @@ public boolean equals(Object o) {
...
                 Objects.equals(percentageComplete, job.percentageComplete) &&
                 Objects.equals(failures, job.failures);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(id, name, description, externalData, createTime, status, percentageComplete, failures);
+    return Objects.hash(id, name, description, externalData, createTime, lastUpdateTime, status, percentageComplete, failures);
```
dermot-hardy: And here - re-indent.

dermot-hardy (approved):
Does adding all new change sets at the bottom of the changelog.xml file mean that we can upgrade from an existing 2.7.0-SNAPSHOT installation?  Or would we only support upgrading from the 2.6.0 release?

joseph-lansdowne-opentext:
I haven't tried running liquibase manually, so I'm not sure.  I assumed it didn't matter, seeing that previous versions' files also had multiple changes to them, and there's no naming convention established for multiple changelogs within a release version.

job-service-db/src/main/resources/v2.7.0/changelog.xml
joseph-lansdowne-opentext: Either way round, there'll be a period of time where the schema and procedure are mismatched, and procedure calls will fail.  As I now understand it, upgrades are expected to have downtime - stop database-using services, upgrade the database, start database-using services - so it seems to me that it shouldn't matter.

dermot-hardy:
I'm not going to worry about it.  We only need to support upgrading from release versions anyway.  Are you going to merge this or do you want me to?  We use the Assignee to indicate who should merge it.  I.e. assign to me if you want me to merge it when PR is approved or assign to yourself if you want review without merging.  No assignee is a bit of a gray area.

joseph-lansdowne-opentext:
Ah, I didn't realise, was just told the committer generally does it.  So yeah, I'll do it.

PR #88, opened 2019-04-11
SCMOD-6334: document the job ID requirements

joseph-lansdowne-opentext:
This is already validated internally, but it's hard to know how to generate valid job IDs when it's not documented.

### DIFF
--- a/job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
+++ b/job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
@@ -95,7 +95,9 @@ paths:
           across different services.
       - name: jobId
         in: path
+        type: string
         required: true
+        description: The identifier of the job
         schema:
           $ref: '#/definitions/jobId'
     get:
@@ -188,7 +190,9 @@ paths:
           across different CAF services
       - name: jobId
         in: path
+        type: string
         required: true
+        description: The identifier of the job
         schema:
           $ref: '#/definitions/jobId'
     post:
@@ -242,8 +246,8 @@ paths:
 definitions:
   jobId:
     type: string
-    format: '^[^.,:;*?!|()]{1,48}$'
     description: The job identifier
+    pattern: '^[^.,:;*?!|()]{1,48}$'
   newJob:
     type: object
     properties:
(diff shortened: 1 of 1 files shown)


andyreidz:
I didn't know the format could be a regex, have you ran a build with this change? I'm concerned it may have changed the generated code.

joseph-lansdowne-opentext:
Oh, I was sure I'd written 'pattern', not 'format'.  And also a good point on the generated code, since I was only thinking about documentation I didn't think to do that.

joseph-lansdowne-opentext:
Okay, so:
- the documentation generation doesn't really do anything for references between schemas, so I left that out
- 'description' is a listed supported JSON schema property, but seems to be ignored for documentation generation, so I moved it backup up to the parameter definition
- 'type' isn't a listed supported parameter property, but it seems to do something anyway, that the 'type' property in JSON schema seems not to do
- in the end, the 'pattern' property doesn't change the code generation at all

andyreidz (approved):

dermot-hardy (changes requested):
I pasted the updated swagger into https://editor.swagger.io to validate it and it came up with the following errors:

	Structural error at paths./jobs/{jobId}.parameters.1
	should NOT have additional properties
	additionalProperty: type, schema, name, in, required, description
	Jump to line 96
	Structural error at paths./jobs/{jobId}.parameters.1.in
	should be equal to one of the allowed values
	allowedValues: body, header, formData, query
	Jump to line 97
	Structural error at paths./jobs/{jobId}/cancel.parameters.1
	should NOT have additional properties
	additionalProperty: type, schema, name, in, required, description
	Jump to line 191
	Structural error at paths./jobs/{jobId}/cancel.parameters.1.in
	should be equal to one of the allowed values
	allowedValues: body, header, formData, query
	Jump to line 192

joseph-lansdowne-opentext:
I'm surprised that `swagger-codegen` doesn't do the same validation.  However, I've discovered the issue: I've been reading the version 3 spec, where we're using version 2.  It looks like the pattern still doesn't have any effect when I put it in the right place, but I think it's still worth having in there.

dermot-hardy (approved):

PR #93, opened 2019-05-08
SCMOD-6622: Add job partitions

joseph-lansdowne-opentext:
This is a breaking change: the base request path to the job-service API is different.  I'm not sure which parts of our own code might need changing - only `enrichment-quickstart` comes to mind right now (and, of course, `sc-gateway`).

Design notes:
- jobs created before this change will have partition `default`
- job IDs are unique within partitions
- the task ID in worker messages is `partition:jobId`

----

- build: http://sou-jenkins2.hpeswlab.net/job/JobService/job/JobService~job-service~SCMOD-6622b~CI/
- ticket: https://portal.digitalsafe.net/browse/SCMOD-6622


pom.xml
```diff
@@ -133,6 +133,11 @@
                 <artifactId>worker-jobtracking-shared</artifactId>
                 <version>3.0.0-SNAPSHOT</version>
             </dependency>
+            <dependency>
+                <groupId>com.github.jobservice</groupId>
+                <artifactId>job-service-util</artifactId>
+                <version>3.0.0-SNAPSHOT</version>
+            </dependency>
```
dermot-hardy: Dependency Management should be in alphabetical order.

pom.xml
```diff
@@ -293,6 +298,12 @@
                 <artifactId>snakeyaml</artifactId>
                 <version>1.23</version>
             </dependency>
+            <dependency>
+                <groupId>junit</groupId>
+                <artifactId>junit</artifactId>
+                <version>4.12</version>
+                <scope>test</scope>
```
dermot-hardy: Dependency management should not include scopes - only include the scope where the dependency is used (and this also should be in the correct alphabetical position).

docs/pages/en-us/Getting-Started.md
```diff
@@ -50,7 +51,7 @@ Adjust 'docker-host' to be the name of your own Docker Host and adjust the port
 	* `taskPipe`: name of the RabbitMQ queue feeding messages to the first worker. <br>
 	* `targetPipe`: name of the final worker's RabbitMQ output queue where tracking will stop. <br><br>
 
-5. Press `Try it out!`. The resulting code will show whether the addition of the job succeeds or not. 
+6. Press `Try it out!`. The resulting code will show whether the addition of the job succeeds or not. 
    - 201, if the job is successfully added
    - 204, if the job is successfully updated
```
dermot-hardy, line 57: We should probably update the images here since they are now out-of-date.

job-service-util/pom.xml
```diff
@@ -0,0 +1,40 @@
...
+        <groupId>com.github.jobservice</groupId>
+        <artifactId>job-service-aggregator</artifactId>
+        <version>3.0.0-SNAPSHOT</version>
+    </parent>
+
+    <dependencies>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+        </dependency>
```
dermot-hardy, line 38: This should be `<scope>test</scope>`.

job-service-util/src/main/java/com/hpe/caf/services/job/util/JobId.java
```diff
@@ -0,0 +1,68 @@
...
+package com.hpe.caf.services.job.util;
+
+/**
+ * Uniquely identify a job.
+ */
+public class JobId {
+    public static final String DEFAULT_PARTITION = "default";
+
+    final String partition;
+    final String jobId;
```
dermot-hardy: These should be `private`.

job-service-util/src/main/java/com/hpe/caf/services/job/util/JobId.java
```diff
@@ -0,0 +1,68 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hpe.caf.services.job.util;
+
+/**
+ * Uniquely identify a job.
+ */
+public class JobId {
```
dermot-hardy: Make classes `final` if they are not intended to be used as a base-class, which I don't think this one is.

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
```diff
@@ -38,6 +38,13 @@ produces:
   - application/json
 
 parameters:
+  partition:
+    name: partition
```
dermot-hardy: I'm inclined to think that this should be `partitionId` rather than just `partition`.

job-service/src/main/java/com/hpe/caf/services/job/api/ApiServiceUtil.java
```diff
@@ -15,6 +15,10 @@
  */
 package com.hpe.caf.services.job.api;
 
+import com.hpe.caf.services.configuration.AppConfigException;
+import com.hpe.caf.services.configuration.AppConfigProvider;
```
dermot-hardy: Unused imports

job-service-db/src/main/resources/procedures/createJobForDependencies.sql
```diff
@@ -127,14 +130,18 @@ BEGIN
         -- this job is blocked from starting. Shared locks are being used as the rows themselves will not be updated by the operation.
         -- This means that there will be no contention when simultaneously creating multiple jobs which have the same pre-requisites.
         SELECT * FROM job
-        WHERE job_id IN (SELECT job_id FROM prereqs) AND status <> 'Completed'
-        ORDER BY job_id
+        WHERE partition = in_partition
+            AND job_id IN (SELECT job_id FROM prereqs)
+            AND status <> 'Completed'
+        ORDER BY partition, job_id
```
dermot-hardy: Is there an advantage to adding `partition` to the `ORDER BY` clause (given that it is already locked down)?  Does this help it choose the correct index for ordering?

job-service-db/src/main/resources/procedures/getJob.sql
```diff
@@ -53,9 +57,12 @@ BEGIN
            job.status,
            job.percentage_complete,
            job.failure_details,
-           CAST('WORKER' AS CHAR(6)) AS actionType
+           CAST('WORKER' AS CHAR(6)) AS actionType,
+           jtd.job_id IS NULL AS can_be_progressed
     FROM job
-    WHERE job.job_id = in_job_id;
+    LEFT JOIN job_task_data AS jtd ON jtd.partition = job.partition AND jtd.job_id = job.job_id
```
dermot-hardy: You're obviously introducing this just for `can_be_progressed`.  Let's make the lookup for that a separate function, rather than including it here.

job-service-db/src/main/resources/procedures/getJobExists.sql
```diff
@@ -0,0 +1,41 @@
...
+LANGUAGE plpgsql STABLE
+AS $$
+BEGIN
+    RETURN QUERY
+    SELECT job.job_hash = in_job_hash
+    FROM job
+    WHERE job.partition = in_partition
+        AND job.job_id = in_job_id;
+END
+$$;
```
dermot-hardy, line 42: There's something a bit odd about this:
- It returns no rows if the job doesn't exist at all.
- It returns one row containing false if the job exists but the hash doesn't match.
- It returns one row containing true if the job exists but that hash does match.

Perhaps a `get_job_hash` function which just returned the hash would be cleaner?  Or use `EXISTS` so that it can be relied upon to return one row with just `true` or `false`.

dermot-hardy (changes requested):
The main issue I see is that the database changes are not sufficient to isolate partitions, as the dynamic tables don't include the partition id.  We may need to change the dynamic table names more radically because of maximum identifier length issues (i.e. we may not have room for adding the partition id directly).

job-service-db/src/main/resources/procedures/createJobForDependencies.sql
joseph-lansdowne-opentext: I'm not sure if it would fail to find the index otherwise, but writing the columns out in index-order makes it more obvious (to me, at least) that it will definitely use the index.

job-service-db/src/main/resources/procedures/getJobExists.sql
joseph-lansdowne-opentext: In an old version I wanted to differentiate between the 3 cases, but I guess I don't any more.  I'll make it always return 1 record.

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
joseph-lansdowne-opentext: I don't feel strongly either way.  Shouldn't be too much effort to change though, even though it affects all the changes.

joseph-lansdowne-opentext:
Yeah, I didn't notice the dynamic tables.  This doesn't seem straightforward - I guess it could go like this:
- disallow '_' in partition
- when creating a task table, use 'taskv2_{partition}_{taskId}'
- when accessing a task table, if partition isn't 'default', use 'taskv2_{partition}_{taskId}'
- when accessing a task table, if partition is 'default', first check for 'task_{taskId}', then use 'taskv2_{partition}_{taskId}'

And then the behaviour can be simplified in the release after this goes out.

joseph-lansdowne-opentext:
Oh, I didn't read your comment about identifier length.

What's the benefit of a table per job?  Why wasn't it a single table in the first place?

dermot-hardy:
I think the main reason for that decision was in order to avoid contention between jobs.  It was made by devs with a SQL Server background, where things like lock escalation are a concern, so I'm not really very confident that there is a contention concern with postgres.  A secondary reason may have been that the tables are quite transient, and it is easier to drop a table than to delete a set of rows from a table.  Again there may have been contention concerns there w.r.t. updating indexes and things like that.

joseph-lansdowne-opentext:
I want to make it a single table then.  As with my previous comment, we'd still need to look at task tables for jobs in the 'default' partition to handle jobs ongoing at the time of the upgrade.

dermot-hardy:
There will only be task tables in the system if there were active jobs at the point of upgrade.  I'd rather that we do something to disallow upgrading systems that have active jobs running than keep a lot of code around to deal with pre-partition task tables (or I suppose we could have an upgrade script that puts them into the new format - if that is easier - either way it should keep the current code simpler).

dermot-hardy:
W.r.t. moving from the task tables to a single table, that feels like a lot of changes for something that is working well.  I'm not saying you're wrong but it just makes me a little bit scared.  Are you sure we couldn't make a simpler change.

joseph-lansdowne-opentext:
Could name task tables 'taskv2_{hash of partition and job ID}'.  A decent-sized hash would fit in the name length limit using digest base64.

dermot-hardy:
OK.  Let's try that but if you think the volume of changes are getting out of hand then we should take a step back and again consider the new static table approach.

job-service-db/src/main/resources/v3.0.0/changelog.xml
```diff
@@ -0,0 +1,194 @@
...
+    </changeSet>
+    <changeSet id="create_procedure_get_job_can_be_progressed" runOnChange="true" author="Gamma Team">
+        <createProcedure path="procedures/getJobCanBeProgressed.sql"
+                         procedureName="get_job_can_be_progressed"
+                         schemaName="public">
+        </createProcedure>
+    </changeSet>
+    <changeSet id="create_procedure_internal_get_task_table_name" runOnChange="true" author="Gamma Team">
+        <sql>
+            CREATE EXTENSION IF NOT EXISTS "pgcrypto";
```
dermot-hardy: I'm a little worried about this.  I tried to run it on my machine and got:

    ERROR:  permission denied to create extension "pgcrypto"
    HINT:  Must be superuser to create this extension.
    SQL state: 42501

I've asked operations for their opinion on it.

dermot-hardy (approved):
Changes look good, but hold back on merging until I'm able to find out if we are OK to use the `pgcrypto` module.

job-service-db/src/main/resources/v3.0.0/changelog.xml
joseph-lansdowne-opentext: Did you work anything out yet?

joseph-lansdowne-opentext:
Fixed an issue that got missed because there are apparently no integration tests for job failure (the meaning of `RESULT_FAILURE` changed at some point to success).  Added a test for it.

job-service-db/src/main/resources/v3.0.0/changelog.xml
dermot-hardy: Hi Joseph.  I'm sorry I never managed to get back to you about this last week.  I spoke to @rob-duewer about it (he is in charge of CA operations) and his opinion is that whilst it would be possible to get the `pgcrypto` module installed that he'd prefer that we update the code so that it is not required.  I tend to agree that this is an unfortunate requirement to introduce and unfortunately I've concluded that we should make further updates to eliminate it.

There may be other possibilities but off the top of my head I think our options are (in no particular order):

1) Re-write the database code to eliminate the use of the dynamic task tables altogether.  
2) Leave partitions (largely) out of the DB-level code and update the calling code to put together a DB-level Job Id which is a hash of the partition id and the real job id.  
3) Update the functions that take in the task id (i.e. `report_complete()` and `report_failure()` I think) so that they also take in a hash of the partition id and the job id (i.e. `hash(partitionId || ':' || jobId)`), and that they use it instead of the job id as the base part of the task id.

Like I say you might come up with other solutions but here is my opinion of the options I've listed:

As you know I'd hoped to avoid option 1 because of volume of change that I imagine it will require, and because I know the existing code has been working very well, but that said there's no doubt that it's the cleanest option so if you prefer to take that approach then that is fine by me.  Option 2 is probably the least change at the DB layer (from a standing start rather than from where this branch is), but it's also pretty hacky as we'd have job id meaning something different in the db layer which would cause problems with filtering and sorting and so on.  I'm pretty sure that it would bite us in the end.  Option 3 is a decent compromise I think, although it might make the DB code more fragile, as we'd be passing around parameters that need to be in-sync with each other.  If option 3 isn't too much change then I'd probably lean towards it, but if you feel strongly for option 1 then go for it.

And once again apologies for taking so long to get back to you on this.

dermot-hardy (changes requested):

job-service-db/src/main/resources/v3.0.0/changelog.xml
joseph-lansdowne-opentext: I'll look at how much effort (3) would take.  For completeness, there are 2 other (bad) solutions:
- use MD5, which is builtin - this is insecure
- use `hashtext()`, which is an undocumented internal postgres function that would require waiting for all running jobs to finish for _all_ future postgres upgrades

dermot-hardy: I'm not against the MD5 option.  MD5 may be insecure but so long as it still produces a uniform distribution of digests then I think it can serve perfectly well for the purpose it would be being used for, which is not a security purpose.  The only reservation I have is that often in security audits blanket questions like "Is MD5 used anywhere?" are posed without reference to the context and to whether the usage actually causes a security issue.

So I think going with the MD5 option would be a good option for now as it would allow us to get this merged and we could handle any further changes in separate stories rather than as part of this initial work.

joseph-lansdowne-opentext: Someone could (theoretically) construct a job ID that causes a table-name collision with a job in a different partition.  You'd have to know the ID of the partition and job you're targeting, and it would have to be a running job.  I'm not sure exactly how you could then turn that into an exploit, but it's not obvious to me that it would be secure.

I don't think (3) would work, because `internal_is_task_completed` is a recursive function which builds the task ID from the input task ID.  The Java code would have to try to pass in hashes for parent jobs too.

dermot-hardy: - Do you think that they could construct a colliding job id whilst still staying within all the general limitations for job ids?  Perhaps I'm wrong but I think it's practically impossible.
- With (3) I was only suggesting to hash the base part of the task id rather than the whole thing, so you should still easily see the hierarchy.

joseph-lansdowne-opentext: There aren't actually many limitations on job IDs (we blacklist rather than whitelist characters), so we could say they're "almost 48 bytes".  With a little research I found some examples of collisions with 64-byte inputs from as much as 7 years ago.  Separately, Wikipedia mentions collision-finding techniques for when the 2 inputs have predetermined prefixes or suffixes.

But MD5 would certainly work, and these issues could be addressed later if we want to get this in.

(3), then, would require the Java code to strip the subtask identifier part (eg. `.123`) from the job ID and hash the partition and the remaining job ID.  The database would use the provided hash and suffix it with the subtask identifier part to obtain the table name.  Am I understanding that right?

dermot-hardy: MD5 usage:
I'd forgotten that characters are blacklisted rather than whitelisted.  If you feel that it is too dangerous then that's fair enough.

On (3):
Yes.  Although when you say "the remaining job id" you are really just referring to the job id, since the `task-id = job-id [ '.' subtask-id ]*`.

joseph-lansdowne-opentext: I did a quick change on another branch to give an idea of the scope of (3) - see https://github.com/JobService/job-service/commit/f2a7c64b06a94316e911685e1d061490256224a2

- seems small enough
- would further require some heavy commenting to make it clear what's going on
- I'd need to think a bit about string sizes and adjust the types accordingly (I don't see any imposed limits on subtask numbers and subtask nesting levels though)
- does feel a bit weird to pass the same data around in 2 forms, as you say

dermot-hardy: Doesn't seem too bad - let's go with it.
- (And you're right that subtask nesting level checks are missing - that is a preexisting issue)

joseph-lansdowne-opentext:
I ended up using the same `varchar` sizes for the new combined IDs so that I could reuse the functions that operate on job and task IDs (eg. `internal_get_parent_task_id`).

dermot-hardy (approved):

PR #89, opened 2019-04-25
SCMOD-6216: Job Service not updating dependent jobs correctly

andyreidz:
@dermot-hardy could you review the SQL function?

PR #94, opened 2019-05-28
SCMOD-6620: Fix locking in internal_process_dependent_jobs

joseph-lansdowne-opentext:
Related to #89.  I've verified that some sort of additional locking is required, and that this is sufficient to avoid the one case that Andy was trying to resolve, by just running transactions manually.  It's different enough from the previous version of the lock that I would expect the deadlock mentioned in #89 not to be possible (though it's not obvious to me where exactly that deadlock was).

----

- build: http://sou-jenkins2.hpeswlab.net/job/JobService/job/JobService~job-service~SCMOD-6620~CI/

### DIFF
--- a/job-service-db/src/main/resources/procedures/processDependentJobs.sql
+++ b/job-service-db/src/main/resources/procedures/processDependentJobs.sql
@@ -47,15 +47,18 @@ BEGIN
         WHERE jd.partition_id = in_partition_id
             AND jd.dependent_job_id = in_job_id;
 
-    -- lock rows
---     PERFORM NULL FROM public.job WHERE job.job_id IN (SELECT tmp_dependent_jobs.job_id FROM tmp_dependent_jobs)
---             ORDER BY job.job_id ASC FOR UPDATE;
-
     -- Remove corresponding dependency related rows for jobs that can be processed immediately
     DELETE
     FROM job_dependency AS jd
     WHERE jd.partition_id = in_partition_id
         AND jd.dependent_job_id = in_job_id;
+
+    -- Ensure that no other `job_dependency` deletion can run until we've committed, so we don't
+    -- miss the deletion of the last dependency
+    PERFORM NULL FROM job_dependency AS jd
+    WHERE jd.partition_id = in_partition_id
+        AND jd.job_id IN (SELECT tmp_dependent_jobs.job_id FROM tmp_dependent_jobs)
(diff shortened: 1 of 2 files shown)


dermot-hardy (changes requested):
> though it's not obvious to me where exactly that deadlock was

The code commented out in #89 could deadlock with the `create_job` code, if one of the new job's pre-requisites was the job actively being completed.

This code doesn't have that issue, but it does have a potential deadlock with itself, that could manifest if two jobs were completing simultaneously that were both pre-reqs of two other jobs.

Imagine:
- There is a J2 with pre-reqs of J0 and J1
- There is a J3 with pre-reqs of J0 and J1
- J0 and J1 complete near simultaneously 

There would be two simultaneously running `internal_process_dependent_jobs` calls trying to take out locks on all four rows of the `job_dependency` table.

The problem is that neither the `DELETE FROM job_dependency` statement (which will implicitly take out update locks) nor the `PERFORM NULL FROM job_dependency ... FOR KEY SHARE` statement specify an order that the locks should be taken out it.  In practice the primary key order would likely be used but this isn't specified so in theory they could deadlock with each other.

joseph-lansdowne-opentext:
Oh, it's actually an issue regardless of row order, isn't it?  Could deadlock if the order of those statements between the 2 transactions is delete / delete / perform / perform - since the deletes each lock only a subset of the rows.

I wasn't aware that row locks from the same statement aren't necessarily acquired atomically.  I would have expected something like that to be more explicitly spelled out - all the examples I see to do with order of locking are using multiple statements in each transaction.

dermot-hardy (approved):
> Oh, it's actually an issue regardless of row order, isn't it? Could deadlock if the order of those statements between the 2 transactions is delete / delete / perform / perform - since the deletes each lock only a subset of the rows.

Yes you're right.  The changes look good to me.

dermot-hardy (changes requested):
On second thoughts I think there is still a problem here.

dermot-hardy:
I don't think the SHARE lock is sufficient.  I think this needs to be an UPDATE lock.

joseph-lansdowne-opentext:
Right, because we want the explicit locks to block each other, not just the delete, to avoid the deadlock.

dermot-hardy:
Yep

dermot-hardy (approved):

PR #97, opened 2019-06-20
SCMOD-6955: Add job type support

joseph-lansdowne-opentext:
- ticket: https://portal.digitalsafe.net/browse/SCMOD-6955
- build: http://sou-jenkins2.hpeswlab.net/job/JobService/job/JobService~job-service~SCMOD-6955~CI/

----

Potential issues:
 - I used a singleton (`JobTypeDefinitions`).  (I would have liked to use a different mechanism, such as storing an instance in jaxrs `ServletContext`, but I don't think I have access to that without modifying the swagger generated code.  I guess another solution would be to add a `JobsPut#setJobTypeDefinitions` static method.)
 - ~~When parsing a job type definition, the code doesn't have access to the filename the definition was loaded from, so if we can't parse the ID there's no identifying information in the error message.~~
 - Job service doesn't exit if it fails loading definitions on startup.  (Not sure how to do this.)
 - There aren't any integration tests for invalid definitions.  (Couldn't think of a way to do this.)

Design notes:
 - In the end, I used JSLT.  There's an argument for using JavaScript, since it's part of the worker framework and is used by the workflow worker.  JSLT has the benefit of not needing to encode or decode JSON any additional times, and the data manipulations we'd use are generally more concise.
 - I used a single YAML file with embedded components, to be similar to the workflow worker.
 - It would be nice to have a framework for testing job type definition files, but maybe that's outside the scope of this ticket.


docs/pages/en-us/Job-Types.md
```diff
@@ -0,0 +1,104 @@
+---
```
dermot-hardy, line 1: I looking on the [Features](https://pages.github.houston.softwaregrp.net/sepg-docs-qa/JobService_CI_job-service_SCMOD-6955/pages/en-us/Features) page on the Documentation QA site and I can't see a way to navigate to this new page.  Should it be linked?  Or be listed in the nav. bar on the right-hard side?

docs/pages/en-us/Job-Types.md
```diff
@@ -0,0 +1,104 @@
...
+- type: string
+
+Notes on what the property means, and what value should be configured for it.
+
+### jobParametersSchema
+
+- required: false (default: input must be missing or null)
+- type: YAML (not a string containing JSON or YAML)
+
+An embedded schema which is used to validate parameters provided with the job before executing `taskDataScript` (below).  The schema language is JSON Schema (`draft-03` or `draft-04`); language reference can be found here:
```
dermot-hardy, line 74: Elsewhere we've used draft 7.  Why did you decide to use these older versions?

job-service-container/pom.xml
```diff
@@ -400,6 +410,12 @@
                                                 <directory>src/main/startup.d/</directory>
                                                 <outputDirectory>startup/startup.d/</outputDirectory>
                                             </fileSet>
+                                            <fileSet>
+                                                <fileMode>0644</fileMode>
+                                                <lineEnding>unix</lineEnding>
+                                                <directory>src/integration-test/resources/job-type-definitions</directory>
+                                                <outputDirectory>test-job-type-definitions</outputDirectory>
+                                            </fileSet>
```
dermot-hardy: This appears to be bundling files intended for testing into the actual image.

docs/pages/en-us/Job-Types.md
```diff
@@ -0,0 +1,104 @@
...
+## Defining a Job Type
+
+A definition is a YAML file containing an object.  The subsections here correspond to properties of that object.
+
+### id
+
+- required: true
+- type: string
+
+Unique identifier for the job type across all of the Job Service's job types.
```
dermot-hardy: Is this required or could the base filename be used as the id?

job-service-container/src/integration-test/resources/job-type-definitions/wrong-ext.notyaml
```diff
@@ -0,0 +1,4 @@
+id: wrong-ext
```
dermot-hardy: Add the license header to this file

job-service-container/src/integration-test/java/com/hpe/caf/services/job/api/JobServiceIT.java
```diff
@@ -57,14 +62,17 @@
...
+    private QueueManager testQueueManager;
 
     private static ServicePath servicePath;
     private static WorkerServices workerServices;
     private static ConfigurationSource configurationSource;
     private static RabbitWorkerQueueConfiguration rabbitConfiguration;
     private static String jobServiceOutputQueue;
     private static final long defaultTimeOutMs = 120000; // 2 minutes
-    
+    private final static ObjectMapper objectMapper = new ObjectMapper();
```
dermot-hardy: The convention is to use `static final` rather than `final static`.  Think "science fiction" rather than "fiction science".  (Also you should be consistent with the line above.)

job-service-container/src/integration-test/java/com/hpe/caf/services/job/api/JobServiceIT.java
```diff
@@ -549,6 +793,34 @@ public void testMessagesPutOnQueue(
...
 
+    /**
+     * Retrieve a single message from a queue.  Call this before triggering the message publish.
+     *
+     * @param queueManager Configured with a specific queue
+     * @return Function to call which waits for the message and returns it.  Throws
+     *         {@link AssertionError} on timeout.
+     * @throws Exception
+     */
+    private Supplier<TaskMessage> getMessageFromQueue(final QueueManager queueManager)
```
dermot-hardy: Can this method be `static`?

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
```diff
@@ -305,6 +308,17 @@ definitions:
         type: string
       targetPipe:
         type: string
+  restrictedTask:
```
dermot-hardy: Looking at the way this is defined I'm wondering if we should move it up a level (i.e. and have `type` and `parameters` directly on `newJob`).  What do you think?

job-service/src/main/java/com/hpe/caf/services/job/jobtype/JobTypes.java
```diff
@@ -0,0 +1,90 @@
...
+        }
+        return INSTANCE;
+    }
+
+    /**
+     * @param id Job type ID to look up
+     * @return Job type with the given ID
+     * @throws BadRequestException If there is no job type with the given ID available
+     */
+    public JobType getjobType(final String id) throws BadRequestException {
```
dermot-hardy: `getjobType()` --> `getJobType()`

job-service/src/test/java/com/hpe/caf/services/job/jobtype/DefaultDefinitionParserTest.java
```diff
@@ -0,0 +1,218 @@
...
+        Assert.assertEquals("should parse api version", 74, task.getTaskApiVersion().intValue());
+        Assert.assertEquals("should get task pipe from config",
+            "basic task pipe", task.getTaskPipe());
+        Assert.assertEquals("should get target pipe from config",
+            "basic target pipe", task.getTargetPipe());
+
+        final Map<String, Map<String, String>> taskData = JobTypeTestUtil.objectMapper.convertValue(
+            task.getTaskData(), new TypeReference<Map<String, Map<String, String>>>() {});
+        Assert.assertEquals("should fill in empty configuration",
+            Collections.emptyMap(), taskData.get("cfg"));
```
dermot-hardy: Either change `Collections.emptyMap()` to `Collections.EMPTY_MAP` or add `@SuppressWarnings("AssertEqualsBetweenInconvertibleTypes")` to the method.

job-service/src/test/java/com/hpe/caf/services/job/jobtype/DirectoryLoaderTest.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    @Rule
+    public final TemporaryFolder tempFiles = new TemporaryFolder();
+
+    @Test
+    public void testLoadEmptyDir() throws Exception {
+        final Path dir = tempFiles.newFolder().toPath();
+        final List<JobType> definitions = new DirectoryLoader(defaultParser, dir).load();
+        Assert.assertEquals(Collections.emptyList(), definitions);
```
dermot-hardy: Use `Collections.EMPTY_LIST` or `@SuppressWarnings("AssertEqualsBetweenInconvertibleTypes")`.

job-service/src/test/java/com/hpe/caf/services/job/jobtype/JobTypeTest.java
```diff
@@ -0,0 +1,82 @@
...
+            task.getTaskPipe(), "task pipe");
+        Assert.assertEquals("target pipe should be as provided",
+            task.getTargetPipe(), "target pipe");
+
+        final Map<String, Object> expectedTaskData = new HashMap<>();
+        expectedTaskData.put("partitionId", "partition id");
+        expectedTaskData.put("jobId", "job id");
+        expectedTaskData.put("parameters", "params");
+        Assert.assertEquals("task data should be as returned by provided builder",
+            task.getTaskData(), expectedTaskData);
```
dermot-hardy: You're putting the assert parameters in the wrong order - for JUnit, unlike TestNG, the order is `(expected, actual)`.

job-service/src/test/java/com/hpe/caf/services/job/jobtype/JsltTaskDataBuilderTest.java
```diff
@@ -0,0 +1,87 @@
...
+        expectedTaskData.put("taskPipe", "task pipe");
+        expectedTaskData.put("targetPipe", "target pipe");
+        expectedTaskData.put("partitionId", "partition id");
+        expectedTaskData.put("jobId", "job id");
+        expectedTaskData.put("parameters", "params");
+        Assert.assertEquals(JobTypeTestUtil.convertJson(expectedTaskData), actualTaskData);
+    }
+
+    @Test(expected = InvalidJobTypeDefinitionException.class)
+    public void testBuildWithIncorrectScriptSyntax() throws Exception {
```
dermot-hardy, line 64: Add `@SuppressWarnings("ResultOfObjectAllocationIgnored")` to the method.

job-service/src/main/java/com/hpe/caf/services/job/jobtype/DirectoryLoader.java
```diff
@@ -0,0 +1,58 @@
...
+        this.parser = parser;
+        this.dir = dir;
+    }
+
+    @Override
+    public List<JobType> load() throws IOException, InvalidJobTypeDefinitionException {
+        final List<JobType> defns = new ArrayList<>();
+        for (final Path file : (Iterable<Path>) Files.list(dir)::iterator) {
+            if (Files.isRegularFile(file) && file.toString().endsWith(".yaml")) {
+                try (final FileInputStream defnStream = new FileInputStream(file.toFile())) {
```
dermot-hardy: Since you're using the NIO stuff you should probably use `Files.newInputStream()` instead.

job-service/src/main/java/com/hpe/caf/services/job/jobtype/DefaultDefinitionParser.java
```diff
@@ -0,0 +1,250 @@
...
+
+    }
+
+
+    /**
+     * An item in a definition's `configurationProperties`, parsed from yaml.
+     */
+    public static final class ConfigurationProperty {
+        private String name;
+        private String description;
```
dermot-hardy, line 215: Should we not have a `getDescription()` or is this only to help with debugging?

dermot-hardy (changes requested):

docs/pages/en-us/Job-Types.md
joseph-lansdowne-opentext: This is what the validator I used claims to support.  I chose this validator because it's used elsewhere (repos `worker-elastic`, `worker-document`) and works directly with jackson's `JsonNode`.

job-service-container/pom.xml
joseph-lansdowne-opentext: Right, so the build part will become the production image but the rest (eg. env, volumes) is for tests?  So I should do this by creating another image like `keystore` and mount it over as a volume?

docs/pages/en-us/Job-Types.md
joseph-lansdowne-opentext: It could work that way; I don't prefer either way.  In the code I treat the definition source independently from its contents, so it doesn't necessarily have a name, but name could be added as a separate concept that also comes from the source.

job-service-container/pom.xml
dermot-hardy: That is one way.  The other way is to build an extra testing image layer on top of this one.  Either is fine.

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
joseph-lansdowne-opentext: Works for me.  It would save talking about the confusing `job.job`.

docs/pages/en-us/Job-Types.md
dermot-hardy: I tend to prefer that the name is used as otherwise the filename and id will get out-of-sync and cause confusion.

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
dermot-hardy: Ok

job-service/src/main/java/com/hpe/caf/services/job/jobtype/DefaultDefinitionParser.java
joseph-lansdowne-opentext: I'm not sure you would ever care what description is - the field's only there because the parser fails when it finds a YAML property and nowhere to put the value.  But there's no harm adding it for completeness.

dermot-hardy: Ok.  Either is fine.

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
joseph-lansdowne-opentext: `type` and not `typeId`?

dermot-hardy: `type` sounds fine to me.

joseph-lansdowne-opentext: Just comparing to partitions, where you preferred `partitionId`.

dermot-hardy: True - I guess I see types as closer to an enumeration but it is very subjective.

joseph-lansdowne-opentext:
That should be everything, still building though.

docs/pages/en-us/Job-Types.md
```diff
@@ -0,0 +1,97 @@
...
+
+### taskDataScript
+
+- required: true
+- type: string
+
+An embedded script which is executed to construct the `taskData` part of the `task` object.  The script language is JSLT (version 0.1.8), configured to preserve all output values.  Language reference can be found here:
+
+- https://github.com/schibsted/jslt/blob/0.1.8/tutorial.md
+- https://github.com/schibsted/jslt/blob/0.1.8/functions.md
```
dermot-hardy: Make these links

dermot-hardy (approved):

dermot-hardy:
Do you think that the headers should also be made available to the transformation script?

joseph-lansdowne-opentext:
Do you mean all the request headers?  Do you have a use case?  I can't really think of anything that couldn't be either a header handled by the job service itself (eg. auth token, instruction to disallow jobs not using type), or a parameter in the request body.

dermot-hardy:
I was thinking about whether it might be desirable to be able to read something like an "X-TENANT-ID" header.  I see you've made the partition id available, but that does not necessarily directly align to the tenant id.  Indeed we intentionally didn't name it tenant id to be clear about that.  I don't think it's important for now - we can always add the feature later if we need to.

dermot-hardy:
Let's do the 3.0 release ([SCMOD-6893](https://portal.digitalsafe.net/browse/SCMOD-6893)) before merging this and make it a 3.1 feature.  (I'd like there to be more time to test it.)

dermot-hardy:
Actually I've just tried testing this and I can't get the swagger page to come up.  Is it coming up for you OK?

dermot-hardy (changes requested):
I found this in the log which probably explains it:

     java.lang.RuntimeException: Error loading job type definitions
        at com.hpe.caf.services.job.jobtype.JobTypesContextListener.contextInitialized(JobTypesContextListener.java:47)
        at org.apache.catalina.core.StandardContext.listenerStart(StandardContext.java:4682)
        ...
        at org.apache.catalina.startup.Bootstrap.main(Bootstrap.java:492)
    Caused by: com.hpe.caf.services.job.jobtype.InvalidJobTypeDefinitionException: batch-ingestion-job: target pipe is not configured
        at com.hpe.caf.services.job.jobtype.DefaultDefinitionParser$Definition.getTargetPipe(DefaultDefinitionParser.java:148)
        at com.hpe.caf.services.job.jobtype.DefaultDefinitionParser.parse(DefaultDefinitionParser.java:66)
        at com.hpe.caf.services.job.jobtype.DirectoryLoader.load(DirectoryLoader.java:53)
        at com.hpe.caf.services.job.jobtype.JobTypes.<init>(JobTypes.java:42)
        at com.hpe.caf.services.job.jobtype.JobTypes.initialise(JobTypes.java:63)
        at com.hpe.caf.services.job.jobtype.JobTypesContextListener.contextInitialized(JobTypesContextListener.java:44)
        ... 41 more

But the reason I didn't set a target pipe is because I didn't want one - target pipe should be optional.

joseph-lansdowne-opentext:
Just target pipe?  (Task pipe is also currently required.)

In a way, I think it's nicer to have to set the environment variable for it to startup, because otherwise if you make a typo in the environment variable name it will take longer to discover and track down the issue.  The only sensible value for that is empty string though, and maybe it's weird to require you to set an environment variable to empty.

dermot-hardy:
Yes task pipe is mandatory as the message has to go somewhere.

I'm fine with it having to be present and empty so long as that translates to it not being specified.

joseph-lansdowne-opentext:
Well, the target pipe change is ready, if Jenkins ever finishes building it.

I've also made the errors change I mentioned on the pull request to the other repo.  Unfortunately I couldn't find any better way of distinguishing the error types than this.

dermot-hardy (approved):

dermot-hardy:
Did the swagger page come up OK for you?

joseph-lansdowne-opentext:
Looks like the `default: null` in the swagger definition makes the UI unhappy, although it works just fine with the client/server generator and the online validator.  It's just a label, so I'll remove it.  (I wonder how swagger UI issues could be picked up by automated testing?)

If this is waiting for the release, what state should I leave the ticket/PR in?  Just as they are?

dermot-hardy:
I don't think there's anything holding up the release.  Check with @michael-bryson.  If there isn't then we can release today and then merge this immediately after release.

michael-bryson:
@dermot-hardy There have been a couple of commits to develop today which have not been QA'd yet and there is also a test failure on the release build so can't be released until that is resolved.

dermot-hardy:
OK thanks.

joseph-lansdowne-opentext:
Merged develop to resolve release notes conflict.

dermot-hardy (approved):

PR #103, opened 2019-06-27
SCMOD-7023: Fix worker-framework and worker-document-framework upgrade issues

joseph-lansdowne-opentext:
The issue with the test was that it was calling `toString()` on the builder, not the `URI`.  In the upgrade, the jsr311 implementation changed from jersey-common to jersey-core, and the previous one just happened to implement the builder's `toString()` such that it gave the required result.

----

- ticket: https://portal.digitalsafe.net/browse/SCMOD-7023
- build: http://sou-jenkins2.hpeswlab.net/job/JobService/job/JobService~job-service~SCMOD-7023~CI/


Anthony-McGreevy:
Reassigned PR to me as Thiliga is now off for 2 weeks

Anthony-McGreevy (approved):

PR #113, opened 2019-08-05
SCMOD-7308: Allow sorting a job listing

joseph-lansdowne-opentext:
I've only included job ID (needed for classifications) and create date (the previous default) columns.

The default order is a bit weird: descending if no column is given (for backwards compatibility), ascending otherwise (I thought descending would be a strange general default).  An alternative would be to have the default order depend on the column (ascending for job ID, descending for create date).

- ticket: https://portal.digitalsafe.net/browse/SCMOD-7308
- build: http://sou-jenkins2.hpeswlab.net/job/JobService/job/JobService~job-service~SCMOD-7308~CI/

### DIFF
--- a/job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
+++ b/job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
@@ -91,6 +91,24 @@ paths:
         format: int32
         required: false
         description: The starting position from which to return results (useful for paging)
+      - name: sortField
+        in: query
+        type: string
+        required: false
+        description: The field to sort the returned results by
+        default: CREATE_DATE
+        enum:
+          - JOB_ID
+          - CREATE_DATE
+      - name: sortOrder
+        in: query
+        type: string
+        required: false
+        description: The direction in which the returned results are sorted with respect to sortField. If sortField is not specified, then the default value is DESCENDING
+        default: ASCENDING
+        enum:
+          - ASCENDING
+          - DESCENDING
     get:
       tags:
         - Jobs
--- a/job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
+++ b/job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
@@ -91,24 +91,12 @@ paths:
         format: int32
         required: false
         description: The starting position from which to return results (useful for paging)
-      - name: sortField
+      - name: sort
         in: query
         type: string
         required: false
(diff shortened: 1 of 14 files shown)


dermot-hardy:
So the usage of this would be along the lines of:

    ?sortField=JOB_ID&sortOrder=ASCENDING

I'm tempted to think that if we are adding this functionality that we should do it with a single parameter:

    ?orderBy=jobId:asc

to allow for secondary ordering in the future.

I agree with you that `:asc` should be the default when a field is specified but that `createDate:desc` should be the overall default when nothing is specified (for backwards compatibility like you said).

joseph-lansdowne-opentext:
Sure, it would be clearer to write `sort=a:asc,b:desc` than `sortField=a,b&sortOrder=asc:desc`.

release-notes-3.1.0.md
```diff
@@ -9,5 +9,7 @@ ${version-number}
        Job types can be defined, and new jobs can target a type to use a well-defined, type-specific input format.
 - [SCMOD-7309](https://portal.digitalsafe.net/browse/SCMOD-7309) NotFinished job status filter  
         The statusType parameter accepts a new value, which includes only jobs which haven't reached a final state.
+- [SCMOD-7308](https://portal.digitalsafe.net/browse/SCMOD-7308) Job list sort options  
+        The sortField and sortOrder parameters have been added to the jobsGet API.
```
dermot-hardy: Needs updated.

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
```diff
@@ -91,6 +91,12 @@ paths:
         format: int32
         required: false
         description: The starting position from which to return results (useful for paging)
+      - name: sort
+        in: query
+        type: string
+        required: false
+        description: 'How to order the returned results, in the format <field>:<direction>.  Allowed values for field: JOB_ID, CREATE_DATE.  Allowed values for direction: ASCENDING, DESCENDING.'
+        default: CREATE_DATE:DESCENDING
```
dermot-hardy: Don't you think that `createDate:desc` would be better than `CREATE_DATE:DESCENDING`?  I realize you're following an enum convention here but the fields are reported back in the payload in camelCase so I think it makes more sense for sort requests to be made in the same way (and also `:asc` and `:desc` are common abbreviations in this context).

joseph-lansdowne-opentext:
(Edit: was replying to old stuff, GitHub didn't update)

dermot-hardy (approved):

PR #115, opened 2019-08-09
SCMOD-7326: Fixed config path

rahul-kulkarni73:
https://portal.digitalsafe.net/browse/SCMOD-7326


joseph-lansdowne-opentext:
(build: http://sou-jenkins2.hpeswlab.net/job/JobService/view/Developer/job/JobService~job-service~SCMOD-7326~CI/ )

joseph-lansdowne-opentext (approved):
This changes the meaning of `JOB_SERVICE_API_CONFIG_PATH` in a couple of ways.  It doesn't seem to be a documented property, so that's fine, but it's probably worth mentioning it on the ticket, in case it's used in any deployments.

rahul-kulkarni73:
Had a quick word with @dermot-hardy . We don't want to recommend this file based config when we can set all the env variables. So we won't be documenting it, just leave in this fix.

PR #120, opened 2019-10-24
SCMOD-7944: Release notes updated

thilagasanthosh:


michael-bryson (changes requested):
What is here look fine, however SCMOD-7336 should be called out as it seems quite an important fix.

michael-bryson (approved):

PR #122, opened 2019-12-05
SCMOD-5674: Updated copyright headers

Anthony-McGreevy:
**BUILD**: http://sou-jenkins2.swinfra.net/job/JobService/job/JobService~job-service~develop~CI/ 
**JIRA**: https://portal.digitalsafe.net/browse/SCMOD-5674

### DIFF
--- a/job-service-client/pom.xml
+++ b/job-service-client/pom.xml
@@ -36,6 +36,11 @@
     
     <inceptionYear>2016</inceptionYear>
 
+    <organization>
+        <name>EntIT Software LLC, a Micro Focus company</name>
+        <url>https://www.microfocus.com</url>
+    </organization>
+
     <properties>
         <copyrightYear>2020</copyrightYear>
         <copyrightNotice>Copyright ${project.inceptionYear}-${copyrightYear} Micro Focus or one of its affiliates.</copyrightNotice>
--- a/job-service-client/pom.xml
+++ b/job-service-client/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright 2015-2018 Micro Focus or one of its affiliates.
+    Copyright 2016-2020 Micro Focus or one of its affiliates.
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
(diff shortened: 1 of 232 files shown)


job-service-db/src/main/resources/v2.7.0/changelog.xml
```diff
@@ -149,12 +149,14 @@
         </addColumn>
     </changeSet>
     <changeSet id="add_procedure_internal_get_options" author="Gamma Team">
+        <validCheckSum>7:bc95dd9476c92f7dcc586d445bf0dbf4</validCheckSum>
```
michael-bryson, line 152: @dermot-hardy Is this ok to be changing? No idea why these 2 sql files checksum has changed and the others didn't when the copyright was updated.

dermot-hardy: @Anthony-McGreevy Why are these changing but others don't need to change?

Anthony-McGreevy: These are the only two whose checksums have included the license header in the 2.7 release. Now that we have changed that file their hashes need to be updated so that we can still upgrade from that releases version of the database.

Anthony-McGreevy: @dermot-hardy @michael-bryson I tried changing the header again to 2021 as Michael suggested and again it resulted in a checksum failure to validate. So I'd say we can be confident that the copyright header is being included in the checksum for these files. Don't think this will be much of an issue in the future as it can just be updated as part of the change to the copyright headers. The build actually states "the checksum is now:" and then gives you the new checksum when it does not match what it was expecting, so updating in future will not be a big undertaking. I also tried upgrading against the 3.1 version of the database just in case there was some issue with jumping a version but the result was the same.

job-service-client2/pom.xml
```diff
@@ -32,14 +32,10 @@
         <relativePath/>
     </parent>
 
-    <organization>
-        <name>EntIT Software LLC, a Micro Focus company</name>
-        <url>https://www.microfocus.com</url>
-    </organization>
+    <inceptionYear>2016</inceptionYear>
```
dermot-hardy, line 35: Why is this changing?

dermot-hardy (changes requested):

job-service-client2/pom.xml
michael-bryson: Its not needed, is inherited from the parent.

job-service-client/pom.xml
```diff
@@ -34,13 +34,10 @@
         <relativePath/>
     </parent>
     
-    <organization>
-        <name>EntIT Software LLC, a Micro Focus company</name>
-        <url>https://www.microfocus.com</url>
-    </organization>
```
dermot-hardy, line 40: I was wrong about the other one but this one is inheriting from `swagger-restapi-client-base::1.14.0-1` whose organization is `Hewlett Packard Enterprise Development LP`.

dermot-hardy (changes requested):

dermot-hardy (approved):

PR #129, opened 2020-01-22
 Revert "SCMOD-5674: Updated copyright headers (#122)"

dgibson-microfocus:
This reverts commit 50e031df in db scripts to avoid checksum modification and
excludes changelogs from license checks.

### DIFF
--- a/job-service-db/pom.xml
+++ b/job-service-db/pom.xml
@@ -46,6 +46,7 @@
                 <configuration>
                     <excludes combine.self="append">
                         <exclude>**/changelog*.xml</exclude>
+                        <exclude>**/procedures/*.sql</exclude>
                     </excludes>
                 </configuration>
             </plugin>
--- a/job-service-db/pom.xml
+++ b/job-service-db/pom.xml
@@ -41,6 +41,15 @@
     <build>
         <plugins>
             <plugin>
+                <groupId>com.mycila</groupId>
+                <artifactId>license-maven-plugin</artifactId>
+                <configuration>
+                    <excludes combine.self="append">
+                        <exclude>**/changelog*.xml</exclude>
+                    </excludes>
+                </configuration>
+            </plugin>
+            <plugin>
                 <artifactId>maven-assembly-plugin</artifactId>
                 <version>3.1.0</version>
                 <configuration>
--- a/job-service-db/pom.xml
+++ b/job-service-db/pom.xml
@@ -44,7 +44,7 @@
                 <groupId>com.mycila</groupId>
                 <artifactId>license-maven-plugin</artifactId>
                 <configuration>
-                    <excludes combine.self="append">
+                    <excludes combine.children="append">
                         <exclude>**/changelog*.xml</exclude>
                         <exclude>**/procedures/*.sql</exclude>
                     </excludes>
--- a/job-service-db/pom.xml
+++ b/job-service-db/pom.xml
@@ -45,7 +45,7 @@
                 <artifactId>license-maven-plugin</artifactId>
                 <configuration>
                     <excludes combine.children="append">
-                        <exclude>**/changelog*.xml</exclude>
+                        <exclude>**/changelog.xml</exclude>
                     </excludes>
                 </configuration>
             </plugin>
--- a/job-service-db/pom.xml
+++ b/job-service-db/pom.xml
@@ -46,7 +46,6 @@
                 <configuration>
                     <excludes combine.children="append">
                         <exclude>**/changelog*.xml</exclude>
-                        <exclude>**/procedures/*.sql</exclude>
                     </excludes>
                 </configuration>
             </plugin>
--- a/job-service-db/src/main/resources/changelog-master.xml
+++ b/job-service-db/src/main/resources/changelog-master.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <!--
 
-    Copyright 2015-2018 Micro Focus or one of its affiliates.
+    Copyright 2016-2020 Micro Focus or one of its affiliates.
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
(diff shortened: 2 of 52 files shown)


job-service-db/pom.xml
```diff
@@ -40,6 +40,15 @@
 
     <build>
         <plugins>
+            <plugin>
+                <groupId>com.mycila</groupId>
+                <artifactId>license-maven-plugin</artifactId>
+                <configuration>
+                    <excludes combine.self="append">
```
dermot-hardy: Shouldn't this be `combine.children="append"`?

dermot-hardy (dismissed):

job-service-db/src/main/resources/v3.1.0/changelog.xml
```diff
@@ -24,7 +24,7 @@
                                        http://www.liquibase.org/xml/ns/dbchangelog
                                        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.3.xsd">
 
-    <changeSet id="update_procedures_fix_parallel_job_creation" author="Gamma Team">
+    <changeSet id="update_procedures_fix_parallel_job_creation" author="Gamma Team" runOnChange="true">
```
dgibson-microfocus, line 27: The `runOnChange=true` here is required as a previous changelog (2.7.0) was already configured with this setting, meaning that the updated stored procedure script will have already been run at this point breaking the checksum of the 3.1.0 changes.

job-service-db/src/main/resources/procedures/cancelJob.sql
```diff
@@ -1,5 +1,5 @@
 --
--- Copyright 2016-2020 Micro Focus or one of its affiliates.
+-- Copyright 2015-2018 Micro Focus or one of its affiliates.
```
dermot-hardy: I thought the procedures were safe to be changed?

job-service-db/src/main/resources/v3.1.0/changelog.xml
```diff
@@ -48,10 +48,19 @@
...
     </changeSet>
 
-    <changeSet id="add_job_sorting" author="Gamma Team">
+    <changeSet id="add_job_sorting" author="Gamma Team" runOnChange="true">
         <createProcedure schemaName="public"
                          procedureName="get_jobs"
                          path="procedures/getJobs.sql" />
+    </changeSet>
+
+    <changeSet id="add_job_sorting_index" author="">
```
dgibson-microfocus: Had to separate out this index creation to it's own `changeSet` so the `add_job_sorting` can be run on change to avoid checksum issues when the stored proc is modified in the future.

The `preCondition` is to guard against attempting to create the index on a 3.1.0 db where the old version of the script has already been applied.

job-service-db/src/main/resources/procedures/cancelJob.sql
dgibson-microfocus: I thought so too, the documentation was vague. After a look at their source code they're explicitly including the sql files into the checksum.

dermot-hardy: This seems worrying to me.  I don't understand how it's never been a problem before or what it means in future.  Should these procedures never be touched again?  Let's talk about this sometime tomorrow.

dgibson-microfocus: Now that I understand the problem more, I can probably back out this part of the change and not exclude the procedures from the license check. This was all caused by the the 3.1.0 changelog which didn't have `runOnChange=true` set on the `changeSets` which modified stored procedures and so it's checksum validation would fail whenever a procedure was changed further down the line (e.g. the label changes).

I think the issue is our approach. Either we should:

-  leave the stored procedures as static and create separate, versioned sql files and apply them in `changeSets` sequentially
Or
-  Modify the stored procedures as we've done, and have a single `changeSet` tag for each procedure that has `runOnChange=true` so the latest version of the procedure is apply every time.

Currently we have a mix where we've kept modifying the stored procedure files and adding a `changeSet` every time we do. That only works if they're all flagged to run on change which wasn't the case in the 3.1.0 change log.

dermot-hardy: Ok if those are the options then it sounds like the second one is the simpler from an ongoing maintenance perspective.

dermot-hardy (changes requested):

job-service-db/src/main/resources/procedures/cancelJob.sql
dgibson-microfocus: I assume you mean this going forward and not saying to remove the duplicate `changeSets` from the current change logs? That would cause a whole load more upgrade issues where the removed `changeSets` have already been applied etc.

job-service-db/pom.xml
```diff
@@ -40,6 +40,15 @@
 
     <build>
         <plugins>
+            <plugin>
+                <groupId>com.mycila</groupId>
+                <artifactId>license-maven-plugin</artifactId>
+                <configuration>
+                    <excludes combine.children="append">
+                        <exclude>**/changelog*.xml</exclude>
```
dermot-hardy: `**/changelog.xml`

job-service-db/src/main/resources/changelog-master.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <!--
 
-    Copyright 2016-2020 Micro Focus or one of its affiliates.
+    Copyright 2015-2018 Micro Focus or one of its affiliates.
```
dermot-hardy: This can go to 2020.

job-service-db/src/main/resources/changelog-master.xml
```diff
@@ -24,6 +24,7 @@
                                        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.3.xsd">
 
    <include file="changelog.xml" relativeToChangelogFile="true"/>
+   <include file="procedures/changelog.xml" relativeToChangelogFile="true"/>
```
dermot-hardy: Does this need to move to the end?

job-service-db/src/main/resources/v3.1.0/changelog.xml
dermot-hardy: Set the author.

dermot-hardy (changes requested):

dermot-hardy (approved):

PR #131, opened 2020-02-03
SCMOD-8513: Fix filtering by label to return all labels of matching jobs

dgibson-microfocus:


job-service-db/src/main/resources/procedures/getJobs.sql
```diff
@@ -90,8 +90,11 @@ BEGIN
 
 
     IF in_labels IS NOT NULL AND ARRAY_LENGTH(in_labels, 1) > 0 THEN
-        sql := sql || whereOrAnd || ' lbl.label = ANY(' || quote_literal(in_labels) || ') ';
-        whereOrAnd := andConst;
+        sql := sql || ' INNER JOIN public.label lbl_filter ON lbl_filter.partition_id = job.partition_id '
+                   || ' AND lbl_filter.job_id = job.job_id '
+                   || ' AND lbl_filter.label = ANY(' || quote_literal(in_labels) || ') ';
+--         sql := sql || whereOrAnd || ' lbl.label = ANY(' || quote_literal(in_labels) || ') ';
+--         whereOrAnd := andConst;
```
Anthony-McGreevy: Why are these lines commented out instead of removed?

Anthony-McGreevy (changes requested):

job-service-db/src/main/resources/procedures/getJobs.sql
dgibson-microfocus: Because I forgot to remove them

Anthony-McGreevy (approved):

PR #132, opened 2020-02-03
SCMOD-7554: Backward compatibility fix

thilagasanthosh:
Daid_output_withOldVariables

Deployed darwin-job-service with older variables on larry and deployment was successful. Please find the output for the same.


dermot-hardy (changes requested):
Why are you giving the `CAF_` variables priority?

thilagasanthosh:
> Why are you giving the `CAF_` variables priority?

@dermot-hardy ,

If JOB_SERVICE_ variables given first priority, it will pick a value from config.properties.
https://github.com/JobService/job-service/blob/develop/job-service-config/src/main/resources/config.properties#L18
The value is "jdbc:postgresql://localhost:5432/jobservice" . So it will fail to connect DB without actual value from Environment variables. The correct application connection values are passed from Environment variables. 

That was the issue darwin-job-service deployment broken with older variables. 
It tried to connect db by URL from config.properties then failed.

dermot-hardy (approved):

PR #133, opened 2020-02-04
SCMOD-8513: Fix paging with labels

dgibson-microfocus:
https://portal.digitalsafe.net/browse/SCMOD-8513

### DIFF
--- a/job-service-db/src/main/resources/procedures/getJobs.sql
+++ b/job-service-db/src/main/resources/procedures/getJobs.sql
@@ -84,16 +84,23 @@ BEGIN
                CAST('WORKER' AS CHAR(6)) AS actionType,
                lbl.label,
                lbl.value
+        FROM
+        (SELECT
+               job.partition_id,
+               job.job_id,
+               job.name,
+               job.description,
+               job.data,
+               job.create_date,
+               job.last_update_date,
+               job.status,
+               job.percentage_complete,
+               job.failure_details,
+               CAST('WORKER' AS CHAR(6)) AS actionType
         FROM job
-        LEFT JOIN public.label lbl ON lbl.partition_id = job.partition_id AND lbl.job_id = job.job_id
+
         $q$;
 
-
-    IF in_labels IS NOT NULL AND ARRAY_LENGTH(in_labels, 1) > 0 THEN
-        sql := sql || ' INNER JOIN public.label lbl_filter ON lbl_filter.partition_id = job.partition_id '
-                   || ' AND lbl_filter.job_id = job.job_id '
-                   || ' AND lbl_filter.label = ANY(' || quote_literal(in_labels) || ') ';
-    END IF;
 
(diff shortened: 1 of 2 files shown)


job-service-db/src/main/resources/procedures/getJobs.sql
```diff
@@ -122,7 +129,7 @@ BEGIN
     END IF;
 
     sql := sql || ' ORDER BY ' || quote_ident(in_sort_field) ||
-        ' ' || CASE WHEN in_sort_ascending THEN 'ASC' ELSE 'DESC' END;
+           ' ' || CASE WHEN in_sort_ascending THEN 'ASC' ELSE 'DESC' END;
```
dermot-hardy: Remove pointless change.

job-service-db/src/main/resources/procedures/getJobs.sql
```diff
@@ -133,7 +140,14 @@ BEGIN
...
     END IF;
-
+    -- Join onto the labels after paging to avoid them bloating the row count
+    sql := sql || ' ) as job LEFT JOIN public.label lbl ON lbl.partition_id = job.partition_id '
+               || 'AND lbl.job_id = job.job_id';
+    IF in_labels IS NOT NULL AND ARRAY_LENGTH(in_labels, 1) > 0 THEN
+        sql := sql || ' INNER JOIN public.label lbl_filter ON lbl_filter.partition_id = job.partition_id '
+                   || ' AND lbl_filter.job_id = job.job_id '
+                   || ' AND lbl_filter.label = ANY(' || quote_literal(in_labels) || ') ';
+    END IF;
```
dermot-hardy: This doesn't look right to me.  You're applying the label restriction after the limit/offset.

dermot-hardy (changes requested):

job-service-db/src/main/resources/procedures/getJobs.sql
```diff
@@ -76,23 +76,36 @@ BEGIN
...
+               job.failure_details,
+               CAST('WORKER' AS CHAR(6)) AS actionType
         FROM job
-        LEFT JOIN public.label lbl ON lbl.partition_id = job.partition_id AND lbl.job_id = job.job_id
-        $q$;
 
+        $q$;
 
     IF in_labels IS NOT NULL AND ARRAY_LENGTH(in_labels, 1) > 0 THEN
         sql := sql || ' INNER JOIN public.label lbl_filter ON lbl_filter.partition_id = job.partition_id '
```
dermot-hardy: I notice you've changed this but not resubmitted it for review yet.  I'm not sure if that was just a mistake or if you were planning to have another look at it but I'll just plough on in anyway and tell you this still isn't right.  (Hint: You should be thinking about using a Semi-Join here rather than an Inner-Join.)

dermot-hardy (changes requested):

dermot-hardy (approved):

PR #144, opened 2020-04-01
SCMOD-9193: Added catch for pgsql exception code

Anthony-McGreevy:
**Jira**: https://portal.digitalsafe.net/browse/SCMOD-9193
**Build**: http://sou-jenkins2.swinfra.net/job/JobService/view/Developer/job/JobService~job-service~SCMOD-9193~CI/

For future reference here is the list of PostgreSQL error codes and their meanings: https://www.postgresql.org/docs/10/errcodes-appendix.html

### DIFF
--- a/worker-jobtracking/src/main/java/com/hpe/caf/worker/jobtracking/JobTrackingWorkerReporter.java
+++ b/worker-jobtracking/src/main/java/com/hpe/caf/worker/jobtracking/JobTrackingWorkerReporter.java
@@ -28,6 +28,7 @@ import java.text.DateFormat;
 import java.text.MessageFormat;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Locale;
 import java.util.Objects;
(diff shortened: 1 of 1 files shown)


worker-jobtracking/src/main/java/com/hpe/caf/worker/jobtracking/JobTrackingWorkerReporter.java
```diff
@@ -277,7 +282,7 @@ private static String getFailureDetailsString(final JobTrackingWorkerFailure rej
      * The SQL Error Code should be a 5 letter code - the first 2 characters denote the class of the error and the final 3 indicate the
      * specific condition.
      */
-    private static boolean isSqlStateIn(final SQLException ex, final String... errorClasses)
+    private static boolean isSqlStateIn(final SQLException ex, final List<String> errorCodes, final String... errorClasses)
```
dermot-hardy: This usage of this function appears awkward now.  How about you put the signature back the way it was and determine whether it is an error class or code by checking the length?

worker-jobtracking/src/main/java/com/hpe/caf/worker/jobtracking/JobTrackingWorkerReporter.java
```diff
@@ -250,7 +254,8 @@ private Connection getConnection() throws JobReportingException
             final String CONNECTION_EXCEPTION = "08";
             final String INSUFFICIENT_RESOURCES = "53";
 
-            if (isSqlStateIn(ex, CONNECTION_EXCEPTION, INSUFFICIENT_RESOURCES, POSTGRES_OPERATOR_FAILURE_CODE_PREFIX)) {
+            if (isSqlStateIn(ex, Arrays.asList(POSTGRES_UNABLE_TO_EXECUTE_READ_ONLY_TRANSACTION_FAILURE_CODE),
+                             CONNECTION_EXCEPTION, INSUFFICIENT_RESOURCES, POSTGRES_OPERATOR_FAILURE_CODE_PREFIX)) {
```
dermot-hardy: I don't think you need to be checking this in `getConnection()`.  It doesn't happen until the connection is being used.

dermot-hardy (changes requested):

dermot-hardy (approved):

PR #146, opened 2020-04-20
SCMOD-8484: Add job failure roll up behaviour

Anthony-McGreevy:
https://portal.digitalsafe.net/browse/SCMOD-8484

### DIFF
--- a/job-service-db/src/main/resources/procedures/changelog.xml
+++ b/job-service-db/src/main/resources/procedures/changelog.xml
@@ -199,6 +199,12 @@
                          schemaName="public">
         </createProcedure>
     </changeSet>
+    <changeSet id="create_process_failed_dependent_jobs_proc" runOnChange="true" author="Gamma Team">
+        <createProcedure path="procedures/processFailedDependentJobs.sql"
+                         procedureName="internal_process_failed_dependent_jobs"
+                         schemaName="public">
+        </createProcedure>
+    </changeSet>
     <changeSet id="create_report_complete_proc" runOnChange="true" author="Gamma Team">
         <createProcedure path="procedures/reportComplete.sql"
                          procedureName="report_complete"
--- a/job-service-db/src/main/resources/procedures/processFailedDependentJobs.sql
+++ b/job-service-db/src/main/resources/procedures/processFailedDependentJobs.sql
@@ -1,0 +1,57 @@
+--
+-- Copyright 2016-2020 Micro Focus or one of its affiliates.
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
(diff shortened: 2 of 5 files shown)


job-service-db/src/main/resources/procedures/changelog.xml
```diff
@@ -199,6 +199,12 @@
                          schemaName="public">
         </createProcedure>
     </changeSet>
+    <changeSet id="create_process_failed_dependent_jobs_proc" runOnChange="true" author="Gamma Team">
+        <createProcedure path="procedures/processFailedDependentJobs.sql"
+                         procedureName="internal_process_failed_dependent_jobs"
+                         schemaName="public">
+        </createProcedure>
+    </changeSet>
```
dermot-hardy, line 207: @dgibson-microfocus I know you looked at the rules around this recently.  Remind me whether this is allowed or if it should be in a v3.3 folder.

worker-jobtracking/src/main/java/com/hpe/caf/worker/jobtracking/JobDatabaseProperties.java
```diff
@@ -58,6 +58,11 @@ public static String getApplicationName() {
             : getPropertyOrEnvVar("JOB_DATABASE_APPNAME");
     }
     
+    public static boolean propagateFailures() {
```
dermot-hardy:
```suggestion
    public static boolean getPropagateFailures() {
```

```suggestion
    public static boolean getShouldPropagateFailures() {
```

job-service-db/src/main/resources/procedures/processFailedDependentJobs.sql
```diff
@@ -0,0 +1,68 @@
...
+    -- Ensure that no other `job` updates can run on these jobs until we've committed
+    -- Lock ALL possibly conflicting rows up-front to avoid deadlocks.
+    PERFORM NULL FROM job AS j
+    WHERE j.partition_id = in_partition_id
+        AND j.job_id IN (SELECT tmp_dependent_jobs.job_id FROM tmp_dependent_jobs)
+    ORDER BY j.partition_id, j.job_id
+    FOR UPDATE;
+
+    UPDATE job AS j
+    SET status = 'Failed', percentage_complete = 0.00
```
dermot-hardy: The failure details and last update date both need to be updated.

dermot-hardy (changes requested):

job-service-db/src/main/resources/procedures/changelog.xml
dgibson-microfocus: It should be in this changelog, not in it's own v3.3 folder. Having it in here means the latest version of the stored procedure will always be applied on upgrade and avoids any checksum and duplication problems if changed later on.

dermot-hardy (approved):

PR #147, opened 2020-04-22
SCMOD-8484: Updated release notes and Docs

Anthony-McGreevy:


release-notes-3.3.0.md
```diff
@@ -1,8 +1,8 @@
-!not-ready-for-release!
-
 #### Version Number
 ${version-number}
 
-#### New Features
+#### New Features  
+- [SCMOD-8484](https://portal.digitalsafe.net/browse/SCMOD-8484): Propagate failures through subtasks
```
michael-bryson: Remove the link to JIRA as its an open source so only some people can access. Leave the JIRA number in place though.

release-notes-3.3.0.md
```diff
@@ -1,8 +1,8 @@
-!not-ready-for-release!
-
 #### Version Number
 ${version-number}
 
-#### New Features
+#### New Features  
+- [SCMOD-8484](https://portal.digitalsafe.net/browse/SCMOD-8484): Propagate failures through subtasks  
+Functionality has been added that allows for the failure of a task to propagate up through any tasks that are no longer able to run because the task has failed.
```
michael-bryson, line 6: Is it worth mentioning that these are prerequisite to the task?

worker-jobtracking/README.md
```diff
@@ -9,3 +9,18 @@ checks for task cancellation, reports the progress of the task to the Job Databa
 queue, e.g. the input queue of another worker.
 
 The Job Service has the ability to execute jobs in an order defined by a job having a dependency on another job.  The Job Service will not execute a job until all the jobs it is dependent upon have completed.  The job tracking worker monitors the progress of jobs, once all dependent jobs have been completed the job tracking worker will automatically forward on the jobs for execution which were previously waiting upon dependent jobs.
+
+## Configuration
+This section describes environment variables that can be used to change the behaviour of the worker-jobtracking.  
+
+- `CAF_JOB_TRACKING_PROPAGATE_FAILURES`  
+`description`: This environment variable will indicate if the worker should propagate job failures through subtasks. This means that if a task fails and as a results other tasks are not able to be run due to prerequisite commitments, those other tasks will be marked as failures also.
```
michael-bryson: "results" should be singular

michael-bryson (changes requested):

release-notes-3.3.0.md
Anthony-McGreevy: these aren't the prerequisite to the task, the task that failed is the prerequisite to the ones involved in the roll up, but since it goes back up multiple levels it being described as affecting the tasks that will no longer ever be eligible to run due to the failure I think is better.

michael-bryson (approved):

PR #153, opened 2020-05-14
SCMOD-9347: Added label filtering functionality

Anthony-McGreevy:
https://portal.digitalsafe.net/browse/SCMOD-9347


Anthony-McGreevy:
I have tested this with local instances of job service using manual testing. I am currently writing integration tests to cover all cases.

dermot-hardy:
This isn't building for me.  I'm going to create a build on the build machine.

dermot-hardy (changes requested):
This doesn't compile on the build machine either:
http://sou-jenkins2.swinfra.net/job/JobService/job/JobService~job-service~SCMOD-9347~CI/

    [ERROR] CustomRSQLVisitorTest.java:[21,30] package org.testng.annotations does not exist
    [ERROR] CustomRSQLVisitorTest.java:[22,25] package org.testng does not exist
    [ERROR] CustomRSQLVisitorTest.java:[26,6] cannot find symbol

I'm moving the Jira back to In Progress.

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
```diff
@@ -102,6 +102,11 @@ paths:
         type: string
         required: false
         description: 'Filter to jobs with the specified labels, in the format label=<labelName>,<labelName>'
+      - name: filter
+        in: query
+        type: string
+        required: false
+        description: 'Filter to jobs with the specified labels, in the format `label == labelOne and value == valueOne`'
```
dermot-hardy: I hope this is a documentation mistake because that's not at all what I was expecting.

job-service/src/main/java/com/hpe/caf/services/job/api/DatabaseHelper.java
```diff
@@ -61,18 +66,20 @@ public DatabaseHelper(AppConfig appConfig)
...
     public Job[] getJobs(final String partitionId, String jobIdStartsWith, String statusType, Integer limit,
                          Integer offset, final JobSortField sortField, final SortDirection sortDirection,
-                         final List<String> labels) throws Exception {
+                         final List<String> labels, final String filter) throws Exception {
 
         final Map<String, Job> jobs = new LinkedHashMap<>(); //Linked rather than hash to preserve order of results.
-
+        String filterQuery = null;
+        if (filter != null) {
+            final Node rootNode = rsqlParser.parse(filter);
```
dermot-hardy: It seems odd to be doing the RSQL parsing in the `DatabaseHelper` module.  Whether this string is valid or not has nothing to do with the database.  If it is invalid you're now throwing non-db related issues out of this module which is clearly intended for database access.

job-service/src/main/java/com/hpe/caf/services/job/api/DatabaseHelper.java
```diff
@@ -61,18 +66,20 @@ public DatabaseHelper(AppConfig appConfig)
...
-     * Returns a list of job definitions in the system.
-     */
     public Job[] getJobs(final String partitionId, String jobIdStartsWith, String statusType, Integer limit,
                          Integer offset, final JobSortField sortField, final SortDirection sortDirection,
-                         final List<String> labels) throws Exception {
+                         final List<String> labels, final String filter) throws Exception {
 
         final Map<String, Job> jobs = new LinkedHashMap<>(); //Linked rather than hash to preserve order of results.
-
+        String filterQuery = null;
```
dermot-hardy: I've said a million times now: if you're variable is not `final` then you need to be thinking about why that is the case and whether you can change the code so that it can be made `final`.

If you'd done that you'd know that this is crying out for functional decomposition:

    final String filterQuery = convertToSqlSyntax(filter);

This is why I bang on about `final`.  The rule is not about keeping the code tidy; it is about driving quality of code improvements.

job-service/src/main/java/com/hpe/caf/services/job/api/DatabaseHelper.java
```diff
@@ -100,6 +107,8 @@ public DatabaseHelper(AppConfig appConfig)
                 array = conn.createArrayOf("VARCHAR", new String[0]);
             }
             stmt.setArray(8, array);
+            LOG.info("Filter = " + filterQuery);
```
dermot-hardy: Again I've told you before that you should be looking at [this](https://github.houston.softwaregrp.net/caf/caf-development/blob/master/mandates/Logging.md#logs) table when deciding on log levels.

Is this log statement really required?  If it is then is info really the right level for it?

job-service/src/main/java/com/hpe/caf/services/job/api/DatabaseHelper.java
```diff
@@ -144,6 +153,16 @@ public DatabaseHelper(AppConfig appConfig)
         return jobArr;
     }
 
+    /**
+     * Returns a list of job definitions in the system.
+     */
+    public Job[] getJobs(final String partitionId, String jobIdStartsWith, String statusType, Integer limit,
```
dermot-hardy: What uses this new method?  I see nothing using it.

job-service/src/main/java/com/hpe/caf/services/job/api/generated/JobsApi.java
```diff
@@ -54,9 +54,10 @@ public Response getJobs(
             @ApiParam(value = "An identifier that can be used to correlate events that occurred\nacross different CAF services" )@HeaderParam("CAF-Correlation-Id") String cAFCorrelationId,
             @ApiParam(value = "How to order the returned results, in the format <field>:<direction>.  Allowed values for field: jobId, createTime.  Allowed values for direction: asc, desc.") @QueryParam("sort") String sort,
             @ApiParam(value = "Filter jobs with any of the specified labels, in the format label=<labelName>,<labelName>") @QueryParam("labelExist") String label,
+            @ApiParam(value = "Filter jobs with any of the specified filter, in the format label=<labelName>,<labelName>") @QueryParam("filter") String filter,
```
dermot-hardy: This parameter description is misleading.

job-service/src/main/java/com/hpe/caf/services/job/api/JobsGet.java
```diff
@@ -60,6 +60,23 @@
      */
     public static Job[] getJobs(final String partitionId, final String jobId, final String statusType, Integer limit,
                                 final Integer offset, final String sort, final String labelExists) throws Exception {
+        return getJobs(partitionId, jobId, statusType, limit, offset, sort, labelExists, null);
```
dermot-hardy: What is the purpose of keeping this method around?  I would have assumed to save you from updating the tests but I can see test modules in this PR where lots of `null`s have been added?

job-service/src/main/java/com/hpe/caf/services/job/exceptions/FilterException.java
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hpe.caf.services.job.exceptions;
+
+public final class FilterException extends RuntimeException
+{
+    private static final String MESSAGE
+        = "Invalid filter provided. Please ensure filter has the appropriate format (examples here: "
+        + "https://github.com/jirutka/rsql-parser#examples)";
```
dermot-hardy: I'd rather direct to our own documentation I think, which reminds me, I don't see any .md file changes in this PR.

job-service/src/main/java/com/hpe/caf/services/job/api/filter/JobFilterQuery.java
```diff
@@ -0,0 +1,163 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hpe.caf.services.job.api.filter;
+
+import com.hpe.caf.services.job.exceptions.FilterException;
+import com.healthmarketscience.sqlbuilder.BinaryCondition;
+import com.healthmarketscience.sqlbuilder.ComboCondition;
```
dermot-hardy: Unused import

job-service/src/main/java/com/hpe/caf/services/job/api/filter/JobFilterQuery.java
```diff
@@ -0,0 +1,163 @@
...
+        }
+        return new UnaryCondition(UnaryCondition.Op.EXISTS, query);
+    }
+}
+
+
+
+
+
+
```
dermot-hardy: Large amount of pointless whitespace at the bottom of the file.

job-service/src/main/java/com/hpe/caf/services/job/api/filter/CustomRsqlVisitor.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+package com.hpe.caf.services.job.api.filter;
+
+import com.healthmarketscience.sqlbuilder.Condition;
+import cz.jirutka.rsql.parser.ast.AndNode;
+import cz.jirutka.rsql.parser.ast.ComparisonNode;
+import cz.jirutka.rsql.parser.ast.OrNode;
+import cz.jirutka.rsql.parser.ast.RSQLVisitor;
+
+public final class CustomRsqlVisitor<T> implements RSQLVisitor<Condition, Void>
```
dermot-hardy: Perhaps I'm missing something but I can't see the point of the split between the three classes in this package.  Personally I think I would have just had one and called it something like `RsqlToSqlConverter`.  And why is it generic - the type parameter isn't being used!

```suggestion
public final class RsqlToSqlConverter implements RSQLVisitor<Condition, Void>
```

job-service/src/main/java/com/hpe/caf/services/job/api/filter/CustomRsqlVisitor.java
```diff
@@ -0,0 +1,55 @@
...
+{
+    private final JdbcQueryBuilder builder;
+
+    public CustomRsqlVisitor()
+    {
+        this.builder = new JdbcQueryBuilder();
```
dermot-hardy: This object doesn't contain any state, so why are we creating an instance of it?

job-service/src/main/java/com/hpe/caf/services/job/api/filter/JdbcQueryBuilder.java
```diff
@@ -0,0 +1,72 @@
...
+import cz.jirutka.rsql.parser.ast.LogicalNode;
+import cz.jirutka.rsql.parser.ast.Node;
+import cz.jirutka.rsql.parser.ast.OrNode;
+import com.healthmarketscience.sqlbuilder.Condition;
+import java.util.Objects;
+import java.util.stream.Stream;
+
+public class JdbcQueryBuilder
+{
+    public Condition createQuery(final ComparisonNode comparisonNode)
```
dermot-hardy: Like I said before I don't understand why this class exists but since it has no state its methods should be `static`.

job-service/src/main/java/com/hpe/caf/services/job/api/filter/JdbcQueryBuilder.java
```diff
@@ -0,0 +1,72 @@
...
+import cz.jirutka.rsql.parser.ast.AndNode;
+import cz.jirutka.rsql.parser.ast.ComparisonNode;
+import cz.jirutka.rsql.parser.ast.LogicalNode;
+import cz.jirutka.rsql.parser.ast.Node;
+import cz.jirutka.rsql.parser.ast.OrNode;
+import com.healthmarketscience.sqlbuilder.Condition;
+import java.util.Objects;
+import java.util.stream.Stream;
+
+public class JdbcQueryBuilder
```
dermot-hardy: Whatever it is this class is not a builder.  In the builder pattern you are building something up in the state of the builder.

Here are some decent examples of builders to look at from worker-elastic to help you to understand what they are:
- [JsonBufferBuilder.java](https://github.houston.softwaregrp.net/caf/worker-elastic/blob/v1.1.0/worker-elastic-jsonbuilderimpl/src/main/java/com/microfocus/caf/worker/elastic/jsonbuilder/impl/JsonBufferBuilder.java#L41)  
It has methods like `writeStartObject()` and `writeStartString()` that are used for gradually building up Json.  This is actually a very good example of the builder pattern because note that rather than having a `String getJson()` method that can be called when the Json has been built up it instead abstracts itself away from the nature of what has been built up (i.e. it can be building up a normal string in memory, or streaming the json across the network - it all depends on the nature of the implementation of `StringBuilderInterface` that was passed to the constuctor).

- [HoleyStringBuilder.java](https://github.houston.softwaregrp.net/caf/worker-elastic/blob/v1.1.0/worker-elastic-textutils/src/main/java/com/microfocus/caf/worker/elastic/textutils/HoleyStringBuilder.java#L28)
This has the normal `append()` methods that you see on string builders but also has a method for appending `Reader`s.  Ultimately it builds up a reader itself that can be used to read back the string interspersed with the readers in the correct locations.

job-service/src/main/java/com/hpe/caf/services/job/api/filter/JdbcQueryBuilder.java
```diff
@@ -0,0 +1,72 @@
...
+    {
+        final Condition result = JobFilterQuery.getQueryStatement(comparisonNode.getSelector(),
+                                                                  comparisonNode.getOperator(),
+                                                                  comparisonNode.getArguments());
+        return result;
+    }
+
+    public Condition createQuery(final AndNode logicalNode)
+    {
+        return getStream(logicalNode).reduce((r, s) -> ComboCondition.and(r, s)).get();
```
dermot-hardy: Pointless lambda just passing the parameters it was given.

```suggestion
        return getStream(logicalNode).reduce(ComboCondition::and).get();
```

job-service/src/main/java/com/hpe/caf/services/job/api/filter/JdbcQueryBuilder.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    public Condition createQuery(final OrNode logicalNode)
+    {
+        return getStream(logicalNode).reduce((r, s) -> ComboCondition.or(r, s)).get();
```
dermot-hardy:
```suggestion
        return getStream(logicalNode).reduce(ComboCondition::or).get();

```

job-service/src/main/java/com/hpe/caf/services/job/api/filter/JdbcQueryBuilder.java
dermot-hardy: Looking at the JavaDoc for `ComboCondition::and` I can see that it takes multiple subconditions, so why wouldn't we take advantage of this rather than using reduction?

job-service/src/main/java/com/hpe/caf/services/job/api/filter/JdbcQueryBuilder.java
dermot-hardy: Looking at the JavaDoc for `ComboCondition::or` I can see that it takes multiple subconditions, so why wouldn't we take advantage of this rather than using reduction?

job-service/src/main/java/com/hpe/caf/services/job/api/filter/JdbcQueryBuilder.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    private Stream<Condition> getStream(final LogicalNode logicalNode)
+    {
+        return logicalNode.getChildren()
+            .stream()
+            .map(node -> createQuery(node))
```
dermot-hardy: Pointless lambda

job-service/src/main/java/com/hpe/caf/services/job/api/filter/JdbcQueryBuilder.java
```diff
@@ -0,0 +1,72 @@
...
+            .filter(Objects::nonNull);
```
dermot-hardy: I see the `createQuery()` method below is returning `null` for unrecognised nodes and you are throwing them away here.  Do you really think it's a good idea to throw away unrecognised nodes?  Shouldn't they cause an exception to be thrown rather than actually producing some condition?

job-service/src/main/java/com/hpe/caf/services/job/api/filter/JobFilterQuery.java
```diff
@@ -0,0 +1,163 @@
...
+    private static final DbColumn JOB_PERCENTAGE_COMPLETE;
+
+    private static final DbColumn LABEL_JOB_ID;
+    private static final DbColumn LABEL_LABEL;
+    private static final DbColumn LABEL_PARTITION_ID;
+    private static final DbColumn LABEL_VALUE;
+
+    static {
+        JOB_TABLE = new DbTable(new DbSchema(new DbSpec(), ""), "job", "job");
+        LABEL_TABLE = new DbTable(new DbSchema(new DbSpec(), ""), "label", "label");
```
dermot-hardy, line 62: Is it safe to create two different `DbSpec` instances for this?

I note that the JavaDoc states:
> Note, you generally do not want to mix objects from different specs because they may have conflicting aliases.

job-service/src/main/java/com/hpe/caf/services/job/api/filter/JobFilterQuery.java
```diff
@@ -0,0 +1,163 @@
...
+        JOB_NAME = JOB_TABLE.addColumn("name");
+        JOB_PARTITION_ID = JOB_TABLE.addColumn("partition_id");
+        JOB_CREATE_TIME = JOB_TABLE.addColumn("create_date");
+        JOB_LAST_MODIFIED = JOB_TABLE.addColumn("last_update_date");
+        JOB_STATUS = JOB_TABLE.addColumn("job_status");
+        JOB_PERCENTAGE_COMPLETE = JOB_TABLE.addColumn("percentage_complete");
+
+        LABEL_PARTITION_ID = LABEL_TABLE.addColumn("partition_id");
+        LABEL_LABEL = LABEL_TABLE.addColumn("label");
+        LABEL_JOB_ID = LABEL_TABLE.addColumn("job_id");
```
dermot-hardy, line 73: A minor point but shouldn't job id come before label.

job-service/src/main/java/com/hpe/caf/services/job/api/filter/JobFilterQuery.java
```diff
@@ -0,0 +1,163 @@
...
+        LABEL_VALUE = LABEL_TABLE.addColumn("value");
+
+        COLUMN_MAPPINGS = new ConcurrentHashMap<>();
```
dermot-hardy: You can just use an ordinary `HashMap` here.  You only need a `ConcurrentHashMap` if there was a possibility of multiple threads modifying the map concurrently, but in this case the map is never modified after it is initialized.

job-service/src/main/java/com/hpe/caf/services/job/api/filter/JobFilterQuery.java
```diff
@@ -0,0 +1,163 @@
...
+        COLUMN_MAPPINGS.put("percentageComplete", JOB_PERCENTAGE_COMPLETE);
+    }
+
+    public static Condition getQueryStatement(final String key, final ComparisonOperator comparisonOperator, final List<String> args)
+    {
+        if (key.startsWith("labels.")) {
+            final String[] keys = key.split("\\.");
+            final Condition con1 = BinaryCondition.equalTo(JOB_PARTITION_ID, LABEL_PARTITION_ID);
+            final Condition con2 = BinaryCondition.equalTo(JOB_ID, LABEL_JOB_ID);
+            final Condition con3 = convertConditionString(LABEL_LABEL,
```
dermot-hardy: Calling into `convertConditionString()` for this instead of using `BinaryCondition.equalTo()` doesn't make sense.

job-service/src/main/java/com/hpe/caf/services/job/api/filter/JobFilterQuery.java
```diff
@@ -0,0 +1,163 @@
...
+                                                          new ComparisonOperator("=="),
+                                                          Arrays.asList(keys[1]));
+            final Condition con4 = convertConditionString(LABEL_VALUE,
+                                                          comparisonOperator,
+                                                          args);
+            return labelExistsCon(con1, con2, con3, con4);
+
+        }
+        return convertConditionString(COLUMN_MAPPINGS.get(key), comparisonOperator, args);
```
dermot-hardy: I'd stick this in an `else` for clarity.

job-service/src/main/java/com/hpe/caf/services/job/api/filter/JobFilterQuery.java
```diff
@@ -0,0 +1,163 @@
...
+        final Condition condition;
+        switch (comparisonOperator.getSymbol()) {
+            case "==":
+                condition = BinaryCondition.equalTo(key, args.get(0));
+                break;
+            case "!=":
+                condition = BinaryCondition.notEqualTo(key, args.get(0));
+                break;
+            case "=gt=":
+            case ">":
```
dermot-hardy: Surely you only need to be switching on the main symbols here?  I wouldn't have expected you to have to list `>` as well.  Same comment for the others that you have doubled up.

job-service/src/main/java/com/hpe/caf/services/job/api/filter/JobFilterQuery.java
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    private static Condition convertConditionString(final DbColumn key, final ComparisonOperator comparisonOperator,
+                                                    final List<String> args)
+    {
+
+        final Condition condition;
+        switch (comparisonOperator.getSymbol()) {
+            case "==":
+                condition = BinaryCondition.equalTo(key, args.get(0));
```
dermot-hardy: What about [wildcards](https://github.houston.softwaregrp.net/tlc-apimanagement/rest-standard#wildcards)?  I thought there was something in the FIQL/RSQL docs about asterisks at the beginning or end of the string being treated as wildcards (which is why I wrote that in the REST document).  Wouldn't they need to be converted to a `LIKE` statement in SQL?

job-service-db/src/main/resources/procedures/getJobs.sql
```diff
@@ -134,6 +141,10 @@ BEGIN
         END IF;
     END IF;
 
+    IF in_filter IS NOT NULL THEN
+        sql := sql || whereOrAnd || in_filter;
```
dermot-hardy, line 145: Don't you need brackets to ensure that OR expressions are handled correctly?  (Perhaps the condition object is already handling that in the Java?)

```suggestion
        sql := sql || whereOrAnd || '(' || in_filter || ')';
```

dermot-hardy (changes requested):

docs/pages/en-us/Filtering.md
```diff
@@ -0,0 +1,54 @@
...
+Below is a list of the RSQL operators and what they can be used for.
+
+| Operators  | Meaning  |  
+|----------|----------|  
+|    ==    | Equal to |  
+|    !=    | Not equal to |  
+| > / =gt= | Greater than |  
+| >= / =ge= | Greater than or equal to |  
+| > / =lt= | Less than |  
+| >= / =le= | Less than or equal to |
```
dermot-hardy: The way you've written the comparison operators with the slash between them looks confusing.

image

docs/pages/en-us/Filtering.md
```diff
@@ -0,0 +1,54 @@
...
+|    ==    | Equal to |  
+|    !=    | Not equal to |  
+| > / =gt= | Greater than |  
+| >= / =ge= | Greater than or equal to |  
+| > / =lt= | Less than |  
+| >= / =le= | Less than or equal to |  
+| =in= | Value present in array ["job1", "job2"] |  
+| =out= | Value not present in array ["job1", "job2"] |  
+| and | Compounding operator that combines two statements together |  
+| or | Compounding operator that combines two statements together |
```
dermot-hardy: I notice you didn't bother with the FIQL variant here, sticking with just the RSQL syntax.  I think this is the correct decision and what you should do throughout this document, including in the examples below.

docs/pages/en-us/Filtering.md
```diff
@@ -0,0 +1,54 @@
...
+Job Status is not Completed or Failed.  
+- `percentageComplete=gt=50`  
+Percentage complete is more than 50% complete.  
+- `percentageComplete=lt=50`  
+Percentage complete is less than 50% complete.  
+- `percentageComplete=ge=50`  
+Percentage complete is more than or equal to 50% complete.  
+- `percentageComplete=le=50`  
+Percentage complete is less than or equal to 50% complete.  
+- `percentageComplete=gt=50 and status!='Failed`
```
dermot-hardy: Like I said above, just use RSQL operators like `>` in the example, rather than `=gt=`.

job-service/src/main/java/com/hpe/caf/services/job/exceptions/FilterException.java
dermot-hardy: This is still not directing to our own documentation.

job-service/src/main/java/com/hpe/caf/services/job/api/filter/RsqlToSqlConverter.java
```diff
@@ -0,0 +1,91 @@
...
+import cz.jirutka.rsql.parser.ast.Node;
+import cz.jirutka.rsql.parser.ast.OrNode;
+import cz.jirutka.rsql.parser.ast.RSQLVisitor;
+import java.util.stream.Stream;
+
+public final class RsqlToSqlConverter implements RSQLVisitor<Condition, Void>
+{
+    public RsqlToSqlConverter()
+    {
+    }
```
dermot-hardy: Since you don't have any state, I would just make this a singleton:

```suggestion
public enum RsqlToSqlConverter implements RSQLVisitor<Condition, Void>
{
    INSTANCE;
```

Then change your calls to the constructor to be just `RsqlToSqlConverter.INSTANCE`.

job-service/src/main/java/com/hpe/caf/services/job/api/filter/RsqlToSqlConverter.java
```diff
@@ -0,0 +1,91 @@
...
+        if (node instanceof AndNode) {
+            return createQuery((AndNode) node);
+        }
+        if (node instanceof OrNode) {
+            return createQuery((OrNode) node);
+        }
+        if (node instanceof ComparisonNode) {
+            return createQuery((ComparisonNode) node);
+        }
+        throw new FilterException("Unkown node type, node did not match comparision, And or Or node.");
```
dermot-hardy: Why not just do this the same way you do in the `convertToSqlSyntax()` method, rather than manually working through all the possibilities?

```suggestion
        return node.accept(new RsqlToSqlConverter());
```

(or `return node.accept(INSTANCE);` if you change it to a singleton like I said to in the previous comment).

job-service/src/main/java/com/hpe/caf/services/job/api/filter/RsqlToSqlConverter.java
```diff
@@ -0,0 +1,91 @@
...
+    {
+        final Condition result = JobFilterQuery.getQueryStatement(comparisonNode.getSelector(),
+                                                                  comparisonNode.getOperator(),
+                                                                  comparisonNode.getArguments());
+        return result;
+    }
+
+    private static Condition createQuery(final AndNode logicalNode)
+    {
+        return getStream(logicalNode).reduce(ComboCondition::and).get();
```
dermot-hardy: From previous review:
> Looking at the JavaDoc for `ComboCondition::and` I can see that it takes multiple subconditions, so why wouldn't we take advantage of this rather than using reduction?

I assume you didn't understand the comment.  Without reduction it should be something more like this:
```suggestion
        return ComboCondition.and(
            StreamSupport.stream(logicalNode.spliterator(), false)
                .map(RsqlToSqlConverter::createQuery)
                .toArray()
        );
```

job-service/src/main/java/com/hpe/caf/services/job/api/filter/RsqlToSqlConverter.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    private static Condition createQuery(final OrNode logicalNode)
+    {
+        return getStream(logicalNode).reduce(ComboCondition::or).get();
```
dermot-hardy: From previous review:
> Looking at the JavaDoc for `ComboCondition::or` I can see that it takes multiple subconditions, so why wouldn't we take advantage of this rather than using reduction?

job-service/src/main/java/com/hpe/caf/services/job/api/filter/RsqlToSqlConverter.java
```diff
@@ -0,0 +1,91 @@
...
+        return createQuery(node);
+    }
+
+    @Override
+    public Condition visit(final ComparisonNode node, final Void params)
+    {
+        return createQuery(node);
+    }
+
+    private static Condition createQuery(final ComparisonNode comparisonNode)
```
dermot-hardy: When you've applied the other comments, re-consider whether any of these `private static Condition createQuery()` methods can be scrapped (which their implementations moving directly into the `visit()` methods).  I suspect most of them should be scrapped.

job-service/src/main/java/com/hpe/caf/services/job/api/filter/JobFilterQuery.java
```diff
@@ -0,0 +1,147 @@
...
+import com.healthmarketscience.sqlbuilder.dbspec.basic.DbColumn;
+import com.healthmarketscience.sqlbuilder.dbspec.basic.DbSchema;
+import com.healthmarketscience.sqlbuilder.dbspec.basic.DbSpec;
+import com.healthmarketscience.sqlbuilder.dbspec.basic.DbTable;
+import cz.jirutka.rsql.parser.ast.ComparisonOperator;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public final class JobFilterQuery
```
dermot-hardy: Utility classes should have a private empty default constructor.

job-service/src/main/java/com/hpe/caf/services/job/api/filter/JobFilterQuery.java
```diff
@@ -0,0 +1,147 @@
...
+            return labelExistsCon(con1, con2, con3, con4);
+        } else {
+            return convertConditionString(COLUMN_MAPPINGS.get(key), comparisonOperator, args);
+        }
+    }
+
+    private static Condition convertConditionString(final DbColumn key, final ComparisonOperator comparisonOperator,
+                                                    final List<String> args)
+    {
+
```
dermot-hardy: Remove this blank line.

job-service/src/main/java/com/hpe/caf/services/job/api/filter/JobFilterQuery.java
```diff
@@ -0,0 +1,147 @@
...
+                break;
+            case "=gt=":
+                condition = BinaryCondition.greaterThan(key, value);
+                break;
+            case "=ge=":
+                condition = BinaryCondition.greaterThanOrEq(key, value);
+                break;
+            case "=lt=":
+                condition = BinaryCondition.lessThan(key, value);
+                break;
```
dermot-hardy: Setting `condition` is OK but I don't really see why you don't just return directly from the switch statement.

```suggestion
                return BinaryCondition.lessThan(key, value);
```

job-service/src/main/java/com/hpe/caf/services/job/api/filter/JobFilterQuery.java
```diff
@@ -0,0 +1,147 @@
...
+            return convertConditionString(COLUMN_MAPPINGS.get(key), comparisonOperator, args);
+        }
+    }
+
+    private static Condition convertConditionString(final DbColumn key, final ComparisonOperator comparisonOperator,
+                                                    final List<String> args)
+    {
+
+        final Condition condition;
+        final String value = args.get(0);
```
dermot-hardy: This seems a little risky to me.  Are you definitely guaranteed that there is at least one argument?  Could "=in=" not be passed an empty list for example?

job-service/src/main/java/com/hpe/caf/services/job/api/filter/JobFilterQuery.java
dermot-hardy: I'd be tempted to make this package scope as I don't see it as being useful except from `RsqlToSqlConverter`.

```suggestion
final class JobFilterQuery
```

job-service/src/main/java/com/hpe/caf/services/job/api/filter/JobFilterQuery.java
```diff
@@ -0,0 +1,147 @@
...
+        switch (comparisonOperator.getSymbol()) {
+            case "==":
+                condition = value.contains("*")
+                    ? BinaryCondition.like(key, value.replace("*", "%"))
```
dermot-hardy: Do you also need to escape the percent (%) and underscore (_) characters to prevent them being interpreted as wildcards?

job-service/src/main/java/com/hpe/caf/services/job/api/DatabaseHelper.java
```diff
@@ -61,18 +61,14 @@ public DatabaseHelper(AppConfig appConfig)
...
-    /**
-     * Returns a list of job definitions in the system.
-     */
     public Job[] getJobs(final String partitionId, String jobIdStartsWith, String statusType, Integer limit,
                          Integer offset, final JobSortField sortField, final SortDirection sortDirection,
-                         final List<String> labels) throws Exception {
+                         final List<String> labels, final String filter) throws Exception {
 
         final Map<String, Job> jobs = new LinkedHashMap<>(); //Linked rather than hash to preserve order of results.
-
```
dermot-hardy, line 72: Pointless whitespace change

dermot-hardy (changes requested):

job-service/src/main/java/com/hpe/caf/services/job/api/filter/JobFilterQuery.java
```diff
@@ -0,0 +1,147 @@
...
+        COLUMN_MAPPINGS.put("createTime", JOB_CREATE_TIME);
+        COLUMN_MAPPINGS.put("lastUpdateTime", JOB_LAST_MODIFIED);
+        COLUMN_MAPPINGS.put("status", JOB_STATUS);
+        COLUMN_MAPPINGS.put("percentageComplete", JOB_PERCENTAGE_COMPLETE);
+    }
+
+    public static Condition getQueryStatement(final String key, final ComparisonOperator comparisonOperator, final List<String> args)
+    {
+        if (key.startsWith("labels.")) {
+            final String[] keys = key.split("\\.");
```
dermot-hardy: Should a limit not be specified?
```suggestion
            final String[] keys = key.split("\\.", 2);
```

dermot-hardy (changes requested):

dermot-hardy (approved):

PR #157, opened 2020-06-04
SCMOD-9780: Updated images to use Java 11

michael-bryson:
There is a test failure in the job service caller when I update this. I updated it to use python3 base image as the python2 one contained Java 8. @dermot-hardy Could you take a look please?


michael-bryson:
2 things to note here is that I converted job-service-caller to a dockerfile because I was struggling to get pip installed and then run the pip install command as the python3 base image does not include pip. The python2 image that was used previously did include pip as well as Java 8, that is why I moved to python3 which does not include Java at all.

dermot-hardy (approved):

michael-bryson:
@dermot-hardy Should I just log a JIRA to deal with the test failure or let the build machine do it from the release build?

dermot-hardy (changes requested):
Oh sorry I missed that.  Don't merge it.  I'll take a look at it later on.

michael-bryson:
@dermot-hardy Can you take a look at this test failure when you get a chance please?

dermot-hardy (approved):

PR #160, opened 2020-06-23
SCMOD-9913: Change JSLT filter to default implmentation

rorytorneymf:
The [taskDataScript ](https://github.com/JobService/job-service/blob/develop/docs/pages/en-us/Job-Types.md#taskdatascript) JSLT parsing code was previously using the `TrueJsonFilter`, which means that all JSON values, including null, empty arrays, and empty objects were included in the JSLT output.

I've changed this to use the `DefaultJSONFilter` instead, which will exclude any JSON properties from the output if their value is null, an empty array, or an empty object.


dermot-hardy:
I'm not sure about this because it's potentially a breaking change.  Ask @joseph-lansdowne-micro-focus why `TrueJsonFilter` was used in the first place.

rorytorneymf:
> I'm not sure about this because it's potentially a breaking change. Ask @joseph-lansdowne-micro-focus why `TrueJsonFilter` was used in the first place.

Yes, I've sent an email to Joseph, will wait to hear back.

dermot-hardy:
We could make this change with an added setting to control which filter you want to use, defaulting to the current behavior so that it is not a breaking change, but I don't think we need to do it at this stage.

PR #164, opened 2020-07-02
SCMOD-10028: Enhanced getJobCount api with filter functionality

Anthony-McGreevy:
https://portal.digitalsafe.net/browse/SCMOD-10028


dermot-hardy:
Why have you not created a developer build?

Anthony-McGreevy:
> Why have you not created a developer build?

Build created and kicked off, should take about 25 mins or so.

dermot-hardy (approved):

PR #167, opened 2020-07-10
SCMOD-10185: Remove early createProcedure changesets

dermot-hardy:
https://portal.digitalsafe.net/browse/SCMOD-10185

- These are done in the procedures/changeset.xml


dermot-hardy:
@dgibson-microfocus I know you looked at this before.  Is there any reason this can't be done?

dgibson-microfocus:
It won't cause any issues with the creation of the stored procedures, but won't this completely break the upgrade? You've modified the checksum of several steps and a 3.2 db will have entries for steps that no longer exist.

dermot-hardy:
What checksums have I modified?  Aren't the checksums calculated at the changeset level?  I've only removed changesets.

dgibson-microfocus:
> 
> 
> What checksums have I modified? Aren't the checksums calculated at the changeset level? I've only removed changesets.

Of course, sorry. I was thinking of flyway which builds a checksum of the whole file.

dgibson-microfocus (approved):
Assuming upgrade still works

PR #174, opened 2020-08-04
SCMOD-9455: Updated Job Service Contract

Anthony-McGreevy:


Anthony-McGreevy:
close as it will be easier to take a new branch and just use the learnings from the experimenting that was done in this branch.

PR #177, opened 2020-08-12
SCMOD-10460: Updated contract for enum issues

Anthony-McGreevy:
https://portal.digitalsafe.net/browse/SCMOD-10460

### DIFF
--- a/job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
+++ b/job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
@@ -372,7 +372,6 @@ definitions:
         description: The time the job status or progress last changed
       status:
         type: string
-        string:
         description: String value of the job status enum, possible values are Completed, Active, Waiting, Failed, Paused and Cancelled.
       percentageComplete:
         type: number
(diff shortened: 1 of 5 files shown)


job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
```diff
@@ -373,13 +373,8 @@ definitions:
...
         description: The status of the job
-        enum:
-          - Active
-          - Cancelled
-          - Completed
-          - Failed
-          - Paused
-          - Waiting
+        string:
+        description: String value of the status enum, possible values are Completed, Active, Waiting, Failed, Paused and Cancelled.
```
dermot-hardy: Surely it isn't valid to have two descriptions?  And what is `string:`?

dermot-hardy (changes requested):

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
```diff
@@ -372,14 +372,8 @@ definitions:
...
         type: string
-        description: The status of the job
-        enum:
-          - Active
-          - Cancelled
-          - Completed
-          - Failed
-          - Paused
-          - Waiting
+        string:
```
dermot-hardy: What is `string:`?  How is it different to `type: string` which seems to make much more sense?

dermot-hardy (changes requested):

dermot-hardy (approved):

PR #182, opened 2020-08-27
SCMOD-10455: Add subtask report table

ghost:

Adding sub-task details into a specific report table for analysis purpose.


job-service-db/src/main/resources/procedures/reportComplete.sql
```diff
@@ -53,6 +53,9 @@ BEGIN
     -- Get the job id
     v_job_id = internal_get_job_id(in_task_id);
 
+    -- Adding sub-task details into a specific report table for analysis purpose ( SCMOD-10455 )
+    INSERT INTO subtask_report (partition_id, task_id, status, create_date) values (in_partition_id, in_task_id, 'Completed', now());
```
dermot-hardy: - Include the job id in the table as well.
- Rename `create_date` to `report_date`.
- Instead of `now()` use `now() AT TIME ZONE 'UTC'`.

worker-jobtracking/src/main/java/com/hpe/caf/worker/jobtracking/JobTrackingWorkerReporter.java
```diff
@@ -105,7 +105,7 @@ public JobTrackingWorkerReporter() throws JobReportingException {
     @Override
     public void reportJobTaskProgress(final String jobTaskId, final int estimatedPercentageCompleted) throws JobReportingException
     {
-        LOG.trace("Recieved progress update message for task {}; taking no action", jobTaskId);
+        LOG.trace("Received progress update message for task {}; taking no action", jobTaskId);
```
dermot-hardy: A good change but it shouldn't be in this PR as it has nothing to do with [SCMOD-10455](https://portal.digitalsafe.net/browse/SCMOD-10455).  Changes for different purposes should be in separate PRs (if a PR is required for the change at all).  For an obvious spelling mistake a PR isn't required at all - i.e. this change should go straight into the main branch.

Please review this:
https://github.com/salemove/github-review-helper/blob/master/doc/rules.md

and in particular what it says about keeping commits focused.

job-service-db/src/main/resources/v3.5.0/changelog.xml
```diff
@@ -0,0 +1,45 @@
...
+            </column>
+            <column name="create_date" type="TIMESTAMP" defaultValueComputed="now()">
+                <constraints nullable="false"/>
+            </column>
+            <column name="failure_details" type="TEXT">
+                <constraints nullable="true"/>
+            </column>
+        </createTable>
+    </changeSet>
+</databaseChangeLog>
```
dermot-hardy: Red no-entry sign = missing newline at the end of the file

job-service-db/src/main/resources/v3.5.0/changelog.xml
```diff
@@ -0,0 +1,45 @@
...
+-->
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd
+                   http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.3.xsd">
+
+    <changeSet id="create_subtask_report" author="Gamma Team">
+        <createTable tableName="subtask_report">
+            <column name="partition_id" type="VARCHAR(48)">
```
dermot-hardy: Isn't this 40 in the other tables?

job-service-db/src/main/resources/v3.5.0/changelog.xml
```diff
@@ -0,0 +1,45 @@
...
+                <constraints nullable="false"/>
+            </column>
+            <column name="task_id" type="VARCHAR(58)">
+                <constraints primaryKey="false" />
```
dermot-hardy: Isn't primaryKey=false the default?

job-service-db/src/main/resources/v3.5.0/changelog.xml
```diff
@@ -0,0 +1,45 @@
...
+            </column>
+            <!--true if completed, false if failed-->
+            <column name="status" type="VARCHAR(12)">
```
dermot-hardy: If this is a boolean, why is it varchar(12)?  Would an enumeration be more efficient?

job-service-db/src/main/resources/v3.5.0/changelog.xml
```diff
@@ -0,0 +1,45 @@
...
+                <constraints nullable="false"/>
+            </column>
+            <column name="create_date" type="TIMESTAMP" defaultValueComputed="now()">
```
dermot-hardy: We don't really need a default value since you're always supplying this.

job-service-db/src/main/resources/v3.5.0/changelog.xml
```diff
@@ -0,0 +1,45 @@
...
+                <constraints nullable="false"/>
+            </column>
+            <column name="failure_details" type="TEXT">
+                <constraints nullable="true"/>
```
dermot-hardy: Is nullable=true not the default?

dermot-hardy (changes requested):

job-service-db/src/main/resources/procedures/reportComplete.sql
```diff
@@ -53,6 +53,9 @@ BEGIN
     -- Get the job id
     v_job_id = internal_get_job_id(in_task_id);
 
+    -- Adding sub-task details into a specific report table for analysis purpose ( SCMOD-10455 )
+    INSERT INTO subtask_report (partition_id, job_id, task_id, status, report_date) values (in_partition_id, v_job_id, in_task_id, 'Completed', now() AT TIME ZONE 'UTC');
```
dermot-hardy: Keep the keywords in capitals:
```suggestion
    INSERT INTO subtask_report (partition_id, job_id, task_id, status, report_date) VALUES (in_partition_id, v_job_id, in_task_id, 'Completed', now() AT TIME ZONE 'UTC');
```

job-service-db/src/main/resources/procedures/reportFailure.sql
```diff
@@ -60,6 +60,10 @@ BEGIN
     -- Get the job id
     v_job_id = internal_get_job_id(in_task_id);
 
+    -- Adding sub-task details into a specific report table for analysis purpose ( SCMOD-10455 )
+    INSERT INTO subtask_report (partition_id, job_id, task_id, status, report_date, failure_details) values (in_partition_id, v_job_id, in_task_id, 'Failed', now() AT TIME ZONE 'UTC', in_failure_details);
```
dermot-hardy:
```suggestion
    INSERT INTO subtask_report (partition_id, job_id, task_id, status, report_date, failure_details) VALUES (in_partition_id, v_job_id, in_task_id, 'Failed', now() AT TIME ZONE 'UTC', in_failure_details);
```

job-service-db/src/main/resources/v3.5.0/changelog.xml
```diff
@@ -0,0 +1,47 @@
...
+-->
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd
+                   http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.3.xsd">
+
+    <changeSet id="create_subtask_report" author="Gamma Team">
+        <sql>CREATE TYPE task_status AS ENUM ('Completed','Failed')</sql>
+        <createTable tableName="subtask_report">
```
dermot-hardy: Check this but I think it is good Liquibase practice to have only one change per changeset (i.e. I think this should really be two changesets rather than two changes in a single changeset).

dermot-hardy (changes requested):

dermot-hardy (approved):

xbreizh:
Added the upgradeJobProgress function and moved the functionality from reportComplete onto getJob

job-service-db/src/main/resources/v3.5.0/changelog.xml
```diff
@@ -0,0 +1,50 @@
...
+            <column name="partition_id" type="VARCHAR(40)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="job_id" type="VARCHAR(48)">
+                <constraints nullable="false" />
+            </column>
+            <column name="task_id" type="VARCHAR(58)">
+                <constraints nullable="false" />
+            </column>
+            <column name="status" type="task_status">
```
dermot-hardy: Now that we've decided only to insert 'Completed' reports into the table let's just drop the status column.

job-service-db/src/main/resources/v3.5.0/changelog.xml
```diff
@@ -0,0 +1,50 @@
...
+                <constraints nullable="false"/>
+            </column>
+            <column name="report_date" type="TIMESTAMP" >
+                <constraints nullable="false"/>
+            </column>
+            <column name="failure_details" type="TEXT">
```
dermot-hardy: And no longer any need for the failure_details column either.

job-service-db/src/main/resources/procedures/reportFailure.sql
```diff
@@ -60,6 +60,10 @@ BEGIN
     -- Get the job id
     v_job_id = internal_get_job_id(in_task_id);
 
+    -- Adding sub-task details into a specific report table for analysis purpose ( SCMOD-10455 )
+    INSERT INTO subtask_report (partition_id, job_id, task_id, status, report_date, failure_details) VALUES (in_partition_id, v_job_id, in_task_id, 'Failed', now() AT TIME ZONE 'UTC', in_failure_details);
```
dermot-hardy: As discussed the other day let's use the table only for successfully completed reports.

job-service-db/src/main/resources/procedures/getJob.sql
```diff
@@ -22,51 +22,65 @@
...
+               job.status,
+               job.percentage_complete,
+               job.failure_details,
+               CAST('WORKER' AS CHAR(6)) AS actionType,
+               lbl.label,
+               lbl.value
+        FROM job
+                 LEFT JOIN public.label lbl ON lbl.partition_id = job.partition_id AND lbl.job_id = job.job_id
+        WHERE job.partition_id = in_partition_id
+          AND job.job_id = in_job_id;
```
dermot-hardy: Don't make unnecessary whitespace changes in reviews.  If whitespace tidy-ups are required then they can be done in separate commits and shouldn't be mixed in with functional changes.

Please review [Four rules for effective collaboration](https://github.com/salemove/github-review-helper/blob/master/doc/rules.md) and specifically the rule about keeping commits focused:
> Formatting changes should be separate from bug fixes.

job-service-db/src/main/resources/procedures/getJob.sql
```diff
@@ -22,51 +22,65 @@
...
+                last_update_date    TEXT,
+                status              job_status,
+                percentage_complete DOUBLE PRECISION,
+                failure_details     TEXT,
+                actionType          CHAR(6),
+                label               VARCHAR(255),
+                label_value         VARCHAR(255)
+            )
+    LANGUAGE plpgsql
+    VOLATILE
```
dermot-hardy: Again I believe the only change here is the change from `STABLE` to `VOLATILE` but lumping in all these formatting changes makes it difficult to see if that is definitely all that is changing here.

job-service-db/src/main/resources/procedures/getJob.sql
```diff
@@ -22,51 +22,65 @@
...
 
+    -- Take out an exclusive update lock on the job row
+    PERFORM NULL
+    FROM job j
+    WHERE j.partition_id = in_partition_id
+      AND j.job_id = in_job_id
+        FOR UPDATE;
+
+
+    PERFORM internal_update_job_progress(in_partition_id, in_job_id);
```
dermot-hardy, line 59: Add a comment above this line.  It looks like it's still in the `Take out an exclusive update lock on the job row` section.

job-service-db/src/main/resources/procedures/updateJobProgress.sql
```diff
@@ -0,0 +1,55 @@
...
+/*
+ *  Name: internal_update_job_progress
+ *
+ *  Description:
+ *  Updates the percentage_complete in the job table with data from subtask_report table.
+ */
+CREATE FUNCTION internal_update_job_progress(in_partition_id VARCHAR(40),
+                                             in_job_id VARCHAR(48))
+    RETURNS VOID
+    LANGUAGE plpgsql
```
dermot-hardy: Function volatility category missing

job-service-db/src/main/resources/procedures/updateJobProgress.sql
```diff
@@ -0,0 +1,55 @@
...
+
+    if array_length(subtask_array, 1) > 0 then
+        FOREACH t_id IN ARRAY subtask_array
+            loop
+                RAISE NOTICE 'in_partition_id: %, task_id: %', in_partition_id, t_id;
+                PERFORM internal_report_task_status(in_partition_id, t_id , 'Completed', 100.00, NULL);
+                END LOOP;
+    end if;
+END
+$$;
```
dermot-hardy: A red no-entry sign means that the newline that should be at the end of the file is missing.

job-service-db/src/main/resources/procedures/updateJobProgress.sql
```diff
@@ -0,0 +1,55 @@
...
+DECLARE
+    t_id          varchar(58);
+    subtask_array varchar[];
+BEGIN
+
+    -- deleting the completed subtasks for that job from subtask_report table
+    -- adding result to an array (subtask_array)
+
+    WITH subtask_completed AS (
+        delete from subtask_report sr where sr.job_id = in_job_id and sr.status = 'Completed'
```
dermot-hardy: - Never, ever, refer to the job_id without the partition_id.  The Job Service is used in a multi-tenant product, where each tenant has its own partition.  Two tenants could be running jobs with the same job id so this code could cause one tenant's job progress to effect another tenant's job.  Any sort of cross-tenant pollution is a very serious issue in any multi-tenant product.
- Also the `status = 'Completed'` isn't required since we decided to only include completed subtasks in the table.

job-service-db/src/main/resources/procedures/updateJobProgress.sql
```diff
@@ -0,0 +1,55 @@
...
+ *  Updates the percentage_complete in the job table with data from subtask_report table.
+ */
+CREATE FUNCTION internal_update_job_progress(in_partition_id VARCHAR(40),
+                                             in_job_id VARCHAR(48))
+    RETURNS VOID
+    LANGUAGE plpgsql
+AS
+$$
+DECLARE
+    t_id          varchar(58);
```
dermot-hardy: This seems like a poor choice of variable name.  It's not immediately clear what it is.

job-service-db/src/main/resources/procedures/reportComplete.sql
```diff
@@ -53,27 +52,7 @@ BEGIN
...
     v_job_id = internal_get_job_id(in_task_id);
 
-    -- Get the job status
-    -- And take out an exclusive update lock on the job row
-    SELECT status INTO v_job_status
-    FROM job j
-    WHERE j.partition_id = in_partition_id
-        AND j.job_id = v_job_id
-    FOR UPDATE;
+    INSERT INTO subtask_report (partition_id, job_id, task_id, status, report_date) VALUES (in_partition_id, v_job_id, in_task_id, 'Completed', now() AT TIME ZONE 'UTC');
```
dermot-hardy: Add a single-line comment here to section this off from the `Get the job id` section.

job-service-db/src/main/resources/procedures/reportComplete.sql
```diff
@@ -53,27 +52,7 @@ BEGIN
...
 
-    -- Check that the job hasn't been deleted, cancelled or completed
-    IF NOT FOUND OR v_job_status IN ('Cancelled', 'Completed') THEN
```
dermot-hardy: Unless we come up with a cleanup strategy for the `subtask_report` table we may need to keep this code (we'll need to think about that).

job-service-db/src/main/resources/procedures/updateJobProgress.sql
```diff
@@ -0,0 +1,55 @@
...
+    SELECT array_agg(task_id)
+    FROM subtask_completed
+    INTO subtask_array;
+
+    -- looping into subtask_array to update the job percentage_complete
+
+    if array_length(subtask_array, 1) > 0 then
+        FOREACH t_id IN ARRAY subtask_array
+            loop
+                RAISE NOTICE 'in_partition_id: %, task_id: %', in_partition_id, t_id;
```
dermot-hardy: I don't see another `RAISE NOTICE` statements in the code.  Is this just for debugging?  Do we need it?  Where are the notices raised to?

job-service-db/src/main/resources/procedures/updateJobProgress.sql
```diff
@@ -0,0 +1,55 @@
...
+        delete from subtask_report sr where sr.job_id = in_job_id and sr.status = 'Completed'
+            RETURNING task_id
+    )
+    SELECT array_agg(task_id)
+    FROM subtask_completed
+    INTO subtask_array;
+
+    -- looping into subtask_array to update the job percentage_complete
+
+    if array_length(subtask_array, 1) > 0 then
```
dermot-hardy: Do you have to check the array length?  If it's an empty array wouldn't the `FOREACH` loop just return?

dermot-hardy (changes requested):
- Next we need to deal with checking if there are dependent jobs, in which case we won't want to take this approach.
- And then we need to deal with other effected functions, like `delete_job`, `get_jobs`, etc.

xbreizh:
- removed completed_subtask_report from reportFailure.sql
- renamed subtask_report into completed_subtask_report
- removed columns from completed_subtask_report
- restored verification on job status

xbreizh:
Made corrections for update_job_progress
created cleanup function for completed_subtask_report table
added the function to delete and cancel functions

xbreizh:
Added cleanup on dependent job if present in completed_subtask_report table

job-service-db/src/main/resources/v3.5.0/changelog.xml
```diff
@@ -0,0 +1,45 @@
...
+-->
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd
+                   http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.3.xsd">
+
+    <changeSet id="create_task_status_enum_type" author="Gamma Team">
+        <sql>CREATE TYPE task_status AS ENUM ('Completed','Failed')</sql>
+    </changeSet>
```
dermot-hardy: Scrap this

job-service-db/src/main/resources/procedures/hasDependentJobs.sql
```diff
@@ -0,0 +1,30 @@
...
+ *  Name: internal_has_dependent_jobs
+ *
+ *  Description:
+ *  Checks if the specified job has dependencies
+ */
+
+CREATE OR REPLACE FUNCTION internal_has_dependent_jobs(
+    in_partition_id VARCHAR(40), in_job_id VARCHAR(58))
+RETURNS BOOLEAN
+LANGUAGE SQL IMMUTABLE
```
dermot-hardy: As discussed, this is not immutable.

job-service-db/src/main/resources/procedures/createJobForDependencies.sql
```diff
@@ -151,6 +151,12 @@ BEGIN
     SELECT in_partition_id, in_job_id, prerequisite_job_id
     FROM all_incomplete_prereqs;
 
+    -- Clears all rows from completed_subtask_report for depending jobs passed in in_prerequisite_job_ids
+    -- and matching partition_id
+    DELETE from completed_subtask_report csr where csr.partition_id = in_partition_id AND csr.job_id = ANY (in_prerequisite_job_ids);
```
dermot-hardy: You haven't processed these reports that you are deleting!

dermot-hardy (changes requested):

job-service-db/src/main/resources/procedures/createJobForDependencies.sql
```diff
@@ -151,6 +151,11 @@ BEGIN
     SELECT in_partition_id, in_job_id, prerequisite_job_id
     FROM all_incomplete_prereqs;
 
+    -- Process outstanding job updates
+    PERFORM internal_update_job_progress_from_array(in_partition_id, in_prerequisite_job_ids);
```
dermot-hardy: I'd be worried that this may not be in the correct place.
1. You've put it between an `INSERT INTO` and an `IF FOUND`.  Are you sure your function doesn't corrupt the state of the `FOUND` variable?  It would be very unusual to see the `FOUND` variable not being checked immediately after the statement it's referring to.
2. It looks like the code above in the `prereqs_created_but_not_complete` section already needs this code to have run in order to get accurate results.

job-service-db/src/main/resources/procedures/updateJobProgressFromArray.sql
```diff
@@ -0,0 +1,51 @@
...
+-- limitations under the License.
+--
+
+/*
+ *  Name: internal_update_job_progress_from_array
+ *
+ *  Description:
+ *  Updates the percentage_complete in the job table with data from subtask_report table from array.
+ */
+CREATE FUNCTION internal_update_job_progress_from_array(in_partition_id VARCHAR(40),
```
dermot-hardy: You're now adding both `internal_update_job_progress` and `internal_update_job_progress_from_array` and they both look almost identical.  Is there any way that one function can be used for both?

job-service-db/src/main/resources/procedures/getJobs.sql
```diff
@@ -173,6 +174,17 @@ BEGIN
...
+    EXECUTE 'CREATE TEMPORARY TABLE getJobTemp ON COMMIT DROP AS ' || sql ;
+
+    FOR jobId IN SELECT * FROM getJobTemp LOOP
+            -- Process outstanding job updates
+            PERFORM internal_update_job_progress(in_partition_id, jobId);
+    END LOOP;
+
+
+    -- Execute and return the table created
     RETURN QUERY EXECUTE sql;
```
dermot-hardy: You're now executing the `sql` twice.  What if things have changed by the second execution and it gives different jobs to the ones you just updated?  Shouldn't you just be returning the contents of your temporary table, with statuses updated of course?

dermot-hardy (changes requested):

job-service-db/src/main/resources/procedures/createJobForDependencies.sql
dermot-hardy: Point 1 done but not 2.

job-service-db/src/main/resources/procedures/getJobs.sql
dermot-hardy: Statuses returned have not been updated.

job-service-db/src/main/resources/procedures/createJobForDependencies.sql
```diff
@@ -102,6 +102,11 @@ BEGIN
         RETURN;
     END IF;
 
+
+    -- Process outstanding job updates
+    PERFORM internal_update_job_progress(in_partition_id, in_prerequisite_job_ids);
```
dermot-hardy: - This isn't taking account of the options flags being present.
- Locks not being taken out upfront leading to some edge cases not being resolved.

job-service-db/src/main/resources/procedures/getJobs.sql
```diff
@@ -183,6 +184,36 @@ BEGIN
...
+                    status = j.status,
+                    percentage_complete = j.percentage_complete,
+                    failure_details = j.failure_details
+            FROM job j WHERE nt.job_id = j.job_id;
+
+    END LOOP;
+
+
+    -- Execute and return the table created
+    RETURN QUERY EXECUTE 'select * from new_table';
```
dermot-hardy: - Does dynamic SQL really need to used here?  It's not obvious to me why it would need to be used.
- PostgreSQL is technically free to return these rows in any order since one has not been specified.

job-service-db/src/main/resources/procedures/getJobs.sql
```diff
@@ -183,6 +184,36 @@ BEGIN
...
+
+    -- Create a duplicate of that temporary table to store the results after the update
+
+    CREATE TEMPORARY TABLE new_table ON COMMIT DROP AS TABLE get_job_temp;
+
+
+
+    FOR jobId IN SELECT * FROM get_job_temp LOOP
+
+            -- Take out an exclusive update lock on the job row
```
dermot-hardy, line 202: Minor issue but all these statements will be run multiple times for each job that has > 1 label.

dermot-hardy (changes requested):

job-service-db/src/main/resources/procedures/createJobForDependencies.sql
dermot-hardy: Point 1 complete.  Point 2 still to be taken care of.

job-service-db/src/main/resources/procedures/getJobs.sql
dermot-hardy: Point 1 complete. Point 2 still to be taken care of.

job-service-db/src/main/resources/procedures/getJobs.sql
```diff
@@ -183,6 +185,54 @@ BEGIN
...
         ' ' || CASE WHEN in_sort_ascending THEN 'ASC' ELSE 'DESC' END;
-    RETURN QUERY EXECUTE sql;
+
+    -- Create temporary table as a base to update the job progress
+    EXECUTE 'CREATE TEMPORARY TABLE get_job_temp ON COMMIT DROP AS ' || sql;
+
+    ALTER TABLE get_job_temp ADD COLUMN id SERIAL PRIMARY KEY;
+
+
+    FOR jobId IN SELECT * FROM get_job_temp
```
dermot-hardy, line 195: I see two `FOR` statements but only one `END LOOP`.  Surely this can't be working?

dermot-hardy: I see two `FOR` statements but only one `END LOOP`.  Surely this can't be working?

dermot-hardy (changes requested):

job-service-db/src/main/resources/v3.5.0/changelog.xml
```diff
@@ -0,0 +1,48 @@
...
+            <column name="task_id" type="VARCHAR(58)">
+                <constraints nullable="false" />
+            </column>
+            <column name="report_date" type="TIMESTAMP" >
+                <constraints nullable="false"/>
+            </column>
+        </createTable>
+    </changeSet>
+    <changeSet id="add_dependency_column" runOnChange="true" author="Gamma Team">
+        <addColumn schemaName="public" tableName="job">
```
dermot-hardy: What is this column?  Why is it needed?

job-service-acceptance-tests/src/test/java/com/hpe/caf/jobservice/acceptance/JobServiceEndToEndIT.java
```diff
@@ -248,6 +248,11 @@ public void testTargetPipeForJobWithNoAndWithCompletedPrerequisiteJobs(final Str
 
         Thread.sleep(1000); // Add short delay to allow previous job to complete
 
+        // Call getJob to trigger the subtask completion
+        jobsApi.getJob(defaultPartitionId, job3Id, jobCorrelationId);
+
+        Thread.sleep(3000); // Add short delay to allow previous jobs to complete
```
dermot-hardy: Why are these sleeps required?

job-service-db/src/main/resources/procedures/updateJobProgress.sql
```diff
@@ -0,0 +1,76 @@
...
+    IF subtask_array IS NOT NULL THEN
+        FOREACH taskId IN ARRAY subtask_array
+        LOOP
+            PERFORM internal_report_task_status(in_partition_id, taskId , 'Completed', 100.00, NULL);
+        END LOOP;
+    END IF;
+
+    RETURN QUERY
+    SELECT j.partition_id, j.job_id, j.status
+    FROM job j;
```
dermot-hardy, line 75: What is the point of the alias here?

```suggestion
    SELECT partition_id, job_id, status
    FROM job;
```

job-service-db/src/main/resources/procedures/getJobs.sql
```diff
@@ -183,6 +184,46 @@ BEGIN
...
+    -- Create temporary table as a base to update the job progress
+    EXECUTE 'CREATE TEMPORARY TABLE get_job_temp ON COMMIT DROP AS ' || sql;
+
+    ALTER TABLE get_job_temp ADD COLUMN id SERIAL PRIMARY KEY;
+
+    -- Create a duplicate of that temporary table to store the results after the update
+    CREATE TEMPORARY TABLE new_table ON COMMIT DROP AS SELECT * FROM get_job_temp ORDER BY id;
+
+    FOR jobId IN SELECT * FROM get_job_temp LOOP
+        -- Take out an exclusive update lock on the job row
```
dermot-hardy: Minor issue but all these statements will be run multiple times for each job that has > 1 label.

job-service-db/src/main/resources/procedures/getJobs.sql
```diff
@@ -183,6 +184,46 @@ BEGIN
          ELSE quote_ident(in_sort_field)
        END ||
         ' ' || CASE WHEN in_sort_ascending THEN 'ASC' ELSE 'DESC' END;
-    RETURN QUERY EXECUTE sql;
+
+    -- Create temporary table as a base to update the job progress
+    EXECUTE 'CREATE TEMPORARY TABLE get_job_temp ON COMMIT DROP AS ' || sql;
```
dermot-hardy, line 191: I've just realized there could be a problem here with the status and percentage complete ordering planned for [SCMOD-10977](https://portal.digitalsafe.net/browse/SCMOD-10977).  I think we can address it in another PR rather than pursuing it now.

job-service-db/src/main/resources/procedures/getJobs.sql
```diff
@@ -183,6 +184,46 @@ BEGIN
...
+        UPDATE new_table nt SET
+               status = j.status,
+               percentage_complete = j.percentage_complete
+        FROM job j
+        WHERE nt.job_id = j.job_id;
+    END LOOP;
+
+    -- Return the new table created
+    RETURN QUERY
+    SELECT at.job_id,
```
dermot-hardy: What is the point of the `at.` alias here?  And if there is a point of it, why doesn't `value` have it?

job-service-db/src/main/resources/procedures/updateJobProgress.sql
xbreizh: It prevents from the following:

{
  "message": "ERROR: column reference \"partition_id\" is ambiguous\n  Detail: It could refer to either a PL/pgSQL variable or a table column.\n  Where: PL/pgSQL function internal_update_job_progress(character varying,anyelement) line 41 at RETURN QUERY\nSQL statement \"SELECT internal_update_job_progress(in_partition_id, jobId)\"\nPL/pgSQL function get_jobs(character varying,character varying,character varying,integer,integer,character varying,character varying,boolean,character varying[],character varying) line 131 at PERFORM"
}

I checked the other scripts that we have returning a table and it seems that we always use an alias

job-service-db/src/main/resources/procedures/getJobs.sql
xbreizh: it prevents the following error:
{
  "message": "ERROR: column reference \"job_id\" is ambiguous\n  Detail: It could refer to either a PL/pgSQL variable or a table column.\n  Where: PL/pgSQL function get_jobs(character varying,character varying,character varying,integer,integer,character varying,character varying,boolean,character varying[],character varying) line 140 at RETURN QUERY"
}
It seems to be conflicting with the job_id field from the table we return.
I think it could be used only on the

job-service-db/src/main/resources/procedures/getJobs.sql
```diff
@@ -56,27 +56,28 @@ CREATE OR REPLACE FUNCTION get_jobs(
...
-    status job_status,
-    percentage_complete DOUBLE PRECISION,
-    failure_details TEXT,
-    actionType CHAR(6),
-    label VARCHAR(255),
-    label_value VARCHAR(255)
-)
-LANGUAGE plpgsql STABLE
+    RETURNS TABLE(
+                     job_id VARCHAR(48),
```
dermot-hardy: The reintroduction of all these whitespace changes is just senseless.

dermot-hardy (changes requested):

job-service-db/src/main/resources/procedures/getJobs.sql
dermot-hardy: I've fixed them in 0958938.

job-service-acceptance-tests/src/test/java/com/hpe/caf/jobservice/acceptance/JobServiceEndToEndIT.java
```diff
@@ -606,7 +616,10 @@ public void testJobWithPrerequisiteJobsAndDelays() throws Exception
         JobServiceDatabaseUtil.assertJobStatus(job3Id, "waiting");
         Assert.assertTrue(JobServiceDatabaseUtil.getJobTaskDataEligibleRunDate(job3Id) != null);
 
-        Thread.sleep(15000); // Add short delay to allow J3 to complete
+        Thread.sleep(90000); // Add delay to allow jobs to complete
```
dermot-hardy: Are these sleep time increases required?  This one in particular is massive, going up from 15 seconds to 90 seconds.

job-service-db/src/main/resources/procedures/updateJobProgress.sql
dermot-hardy: Ok

job-service-db/src/main/resources/procedures/getJobs.sql
```diff
@@ -183,6 +184,52 @@ BEGIN
...
+    ALTER TABLE get_job_temp ADD COLUMN id SERIAL PRIMARY KEY;
+
+    -- Create a duplicate of that temporary table to store the results after the update
+    CREATE TEMPORARY TABLE new_table ON COMMIT DROP AS SELECT * FROM get_job_temp ORDER BY id;
+
+    -- Remove duplicates based on the label
+    DELETE FROM get_job_temp a
+    USING get_job_temp b
+    WHERE a.id < b.id
+        AND a.label = b.label;
```
dermot-hardy: I get the concept of removing the duplicates but does this statement really do that?  It's a difficult one to understand.  Why don't you just change the statement below to `FOR jobId IN SELECT DISTINCT(jobId) FROM get_job_temp LOOP`?  I think it would be easier to understand.

job-service-db/src/main/resources/procedures/getJobs.sql
dermot-hardy: Ok

dermot-hardy (approved):

job-service-db/src/main/resources/procedures/getJobs.sql
xbreizh: replaced the second temp table with a simplified array

job-service-acceptance-tests/src/test/java/com/hpe/caf/jobservice/acceptance/JobServiceEndToEndIT.java
xbreizh: reduced to 30000

job-service-db/src/main/resources/procedures/getJobs.sql
```diff
@@ -183,6 +186,51 @@ BEGIN
...
+    -- Create temporary table as a base to update the job progress
+    EXECUTE 'CREATE TEMPORARY TABLE get_job_temp ON COMMIT DROP AS ' || sql;
+
+    ALTER TABLE get_job_temp ADD COLUMN id SERIAL PRIMARY KEY;
+
+    -- Create an array of job_id(s) based on the get_job_temp table
+    jobArray := ARRAY(SELECT DISTINCT (jt.job_id) FROM get_job_temp jt);
+
+    -- Check that the array is not empty
+    IF array_length(jobArray, 1) > 0 THEN
```
dermot-hardy, line 199: Do you actually need to check if the array is empty?  Wouldn't it just go around the loop zero times?

dermot-hardy (approved):

job-service-db/src/main/resources/procedures/getJobs.sql
xbreizh: It seems so, I ran the tests again without it and getting
image

dermot-hardy: Ok.

dermot-hardy (approved):

PR #183, opened 2020-09-02
SCMOD-9697: Allow sorting on any column

gaana-vishwnath:
JIRA: https://portal.digitalsafe.net/browse/SCMOD-9697
JENKINS: http://sou-jenkins2.swinfra.net/job/JobService/job/JobService~job-service~SCMOD-9697~CI/


job-service/src/main/java/com/hpe/caf/services/job/api/generated/model/JobSortField.java
```diff
@@ -23,7 +23,13 @@
  */
 public enum JobSortField {
     JOB_ID("jobId", "job_id"),
-    CREATE_DATE("createTime", "create_date");
+    CREATE_DATE("createTime", "create_date"),
+    NAME("name", "name"),
+    LAST_UPDATE_DATE("lastUpdateTime", "last_update_date"),
+    STATUS("status", "status"),
+    LABEL("label", "label"),
```
dermot-hardy: `label` makes no sense here.

job-service-db/src/main/resources/procedures/getJobs.sql
```diff
@@ -145,8 +145,7 @@ BEGIN
         sql := sql || whereOrAnd || in_filter;
     END IF;
 
-    sql := sql || ' ORDER BY ' || quote_ident(in_sort_field) ||
-        ' ' || CASE WHEN in_sort_ascending THEN 'ASC' ELSE 'DESC' END;
+    sql := sql || ' ORDER BY create_date ' || CASE WHEN in_sort_ascending THEN 'ASC' ELSE 'DESC' END;
```
dermot-hardy: This change isn't correct because when you combine it with the limit it will eliminate rows that shouldn't be eliminated.

dermot-hardy (changes requested):

gaana-vishwnath:
made all the changes after Monday's (7-sept) discussion.
Kindly review again

dermot-hardy (approved):

job-service/src/main/java/com/hpe/caf/services/job/api/JobsGet.java
```diff
@@ -57,19 +57,30 @@
 
             final JobSortField sortField;
             final SortDirection sortDirection;
+            final String sortByLabelValue;
```
dermot-hardy: I've thought about it further and think it would be worthwhile to introduce that inheritance mechanism that we talked about, to correctly eliminate the either/or situation with `sortField` and `sortByLabelValue`.  It would only be one additional interface and one additional class after all, so go ahead and update the code to take that approach instead.

dermot-hardy (changes requested):

job-service/src/main/java/com/hpe/caf/services/job/api/JobsGet.java
gaana-vishwnath: implemented

dermot-hardy (approved):

PR #186, opened 2020-09-25
SCMOD-11149: Check if queue exists instead of creating it

dgibson-microfocus:
https://portal.digitalsafe.net/browse/SCMOD-11149


dermot-hardy (approved):
Release notes

PR #188, opened 2020-09-29
SCMOD-10455: Correct getJobs function

xbreizh:

Job Tracking Worker consuming all PSQL CPU resources


dermot-hardy (approved):

dermot-hardy (changes requested):
Does line 205 also need to be changed?

job-service-db/src/main/resources/procedures/getJobs.sql
```diff
@@ -206,7 +206,7 @@ BEGIN
                 FOR UPDATE;
 
             -- Process outstanding job updates
-            PERFORM internal_update_job_progress(in_partition_id, jobId);
+            PERFORM internal_update_job_progress(in_partition_id, job);
```
dermot-hardy: Actually I don't get this.  Isn't `jobId` correct, as `job` is an array?

dermot-hardy (changes requested):

job-service-db/src/main/resources/procedures/getJobs.sql
xbreizh: the jobId wasn't being fed. job as an array was unnecessary. updateJobProgress wasn't throwing an error for a null value as it's only checking the type and we were passing a varchar[] which is considered as valid

dermot-hardy (approved):

PR #192, opened 2020-10-20
SCMOD-10825 Upgrade job cancellation process to accept large queries

xbreizh:
https://portal.digitalsafe.net/browse/SCMOD-10825


job-service/src/main/java/com/hpe/caf/services/job/api/DatabaseHelper.java
```diff
@@ -377,7 +377,7 @@ public void deleteJob(final String partitionId, String jobId) throws Exception {
 
         try (
                 Connection conn = DatabaseConnectionProvider.getConnection(appConfig);
-                CallableStatement stmt = conn.prepareCall("{call delete_job(?,?)}")
+                PreparedStatement stmt = conn.prepareCall("call delete_job(?,?)")
```
dermot-hardy: Do we actually need to make these changes to the Java code?

According to https://jdbc.postgresql.org/documentation/head/callproc.html:
> PostgreSQL™ supports two types of stored objects, functions that can return a result value and - starting from v11 - procedures that can perform transaction control. Both types of stored objects are invoked using CallableStatement and the standard JDBC escape call syntax {call storedobject(?)}.

job-service-db/src/main/resources/procedures/dropTaskTables2.sql
```diff
@@ -0,0 +1,62 @@
...
+                END IF;
+                -- we increment the counter
+                counter :=counter+1;
+            END LOOP;
+
+        -- Drop the table itself
+        EXECUTE 'DROP TABLE ' || task_table_ident;
+    END IF;
+END
+$$;
```
dermot-hardy: Take a newline.

job-service-db/src/main/resources/procedures/changelog.xml
```diff
@@ -85,6 +85,12 @@
                          schemaName="public">
         </createProcedure>
     </changeSet>
+    <changeSet id="create_drop_task_tables2_proc" runOnChange="true" author="Gamma Team">
+        <createProcedure path="procedures/dropTaskTables2.sql"
```
dermot-hardy: How about `dropTaskTablesSP.sql` instead of `dropTaskTables2.sql`?  To prevent it sounding like just another version.

job-service-db/src/main/resources/procedures/cancelJob.sql
```diff
@@ -25,11 +25,14 @@ DROP FUNCTION IF EXISTS cancel_job(
     in_job_id VARCHAR(48),
     in_short_job_id VARCHAR(48)
 );
-CREATE OR REPLACE FUNCTION cancel_job(
+DROP FUNCTION IF EXISTS cancel_job(
     in_partition_id VARCHAR(40),
     in_job_id VARCHAR(48)
+);
+CREATE OR REPLACE PROCEDURE cancel_job(
```
dermot-hardy: Have you considered whether the change in semantics going from function to procedure has any other impact?  Is there any difference w.r.t what happens when there is an error at different stages?

job-service-db/src/main/resources/procedures/deleteJob.sql
```diff
@@ -25,11 +25,14 @@ DROP FUNCTION IF EXISTS delete_job(
     in_job_id VARCHAR(48),
     in_short_job_id VARCHAR(48)
 );
-CREATE OR REPLACE FUNCTION delete_job(
+DROP FUNCTION IF EXISTS delete_job(
+    in_partition_id VARCHAR(40),
+    in_job_id VARCHAR(48)
+);
+CREATE OR REPLACE PROCEDURE delete_job(
```
dermot-hardy: Have you considered whether the change in semantics going from function to procedure has any other impact? Is there any difference w.r.t what happens when there is an error at different stages?

job-service-db/src/main/resources/procedures/dropTaskTables2.sql
```diff
@@ -0,0 +1,62 @@
...
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+--
+
+/*
+ *  Name: internal_drop_task_tables2
+ *
+ *  Description:
+ *  Drops all task tables belonging to the specified task and all its subtasks
+ *  This is a procedure that commits on bulks whose size is defined by the maxBeforeCommit variable.
```
dermot-hardy: Is the `maxBeforeCommit` variable meant to be configurable?  Can it's size be automatically chosen based on `max_locks_per_transaction`?

job-service-db/src/main/resources/procedures/dropTaskTables2.sql
```diff
@@ -0,0 +1,62 @@
...
+        -- Drop the referenced subtask tables
+        FOR subtask_suffix IN
+            EXECUTE $ESC$SELECT '.' || subtask_id || CASE WHEN is_final THEN '*' ELSE '' END AS subtask_suffix FROM $ESC$ || task_table_ident
+            LOOP
+                CALL internal_drop_task_tables2(in_partition_id, in_task_id || subtask_suffix);
+
+                -- we commit on each maxBeforeCommit
+                IF (MOD(counter,maxBeforeCommit)=0)
+                THEN
+                    COMMIT;
```
dermot-hardy: I'm not aware of PostgreSQL syntax in this area but it seems surprising to have a `COMMIT` without a `BEGIN`.  Does this automatically start a new transaction?  Does it also commit the changes made outside of the procedure before it was called?  Are locks taken out by the called retained after the commit?

job-service/src/main/java/com/hpe/caf/services/job/api/DatabaseHelper.java
xbreizh: image
Here, I tried with a CallableStatement and got that error on those failing tests. 

In the code while testing, I have `CallableStatement stmt = conn.prepareCall("{call delete_job(?,?)}")`

simranjeetc:
@dermot-hardy 
Couple of open questions/infos -
1. How often do we want to run the deletion procedure? Right now, I have kept the period to be of 60 sec.
2. I have increased the thread pool size from 1 to 2. Let me know if it needs to be something else.
3. In the drop_tables() procedure, I am committing in a batch of 10. I have also not exposed this to be configurable. Since this is a periodic task which is going to execute in background, I did not find any good reason to make it configurable. I also think we do not need to depend on any other postgres parameter for the same. Let me know if you differ.

dermot-hardy:
1. It should be configurable.  I think I'm fine with 60 seconds as a default, though I'd like to understand how much load this much the PostgreSQL under when it runs and how long it takes before settling on that.
2. Why does it need to be increased?  If you don't have a good reason I'd probably be against that.
3. It feels a little low but I have nothing to base that on.  I'm happy to go with it because testing for the optimum value would probably take quite a while.  I'm also happy that it doesn't need to be configurable.

simranjeetc:
>     1. It should be configurable.  I think I'm fine with 60 seconds as a default, though I'd like to understand how much load this much the PostgreSQL under when it runs and how long it takes before settling on that.
Made the period configurable
>     2. Why does it need to be increased?  If you don't have a good reason I'd probably be against that.
I think one should be ok for now unless we see a delay in the scheduled tasks being triggered and if we are not ok with the delay. Reverted to 1.
>     3. It feels a little low but I have nothing to base that on.  I'm happy to go with it because testing for the optimum value would probably take quite a while.  I'm also happy that it doesn't need to be configurable.
@michael-bryson I have added [timer](https://github.com/JobService/job-service/blob/20b4f87d2d9349f7420965e47741bed21ebfe039/job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/DropTablesTask.java#L48) which will be enabled in debug

job-service-container/src/test/java/com/hpe/caf/services/job/api/JobServiceIT.java
```diff
@@ -1129,6 +1142,57 @@ public void testInvalidLabelFormat() {
...
     }
+    
+    @Test
+    public void testDeleteLog() throws SQLException
+    {
+        //prepare
+        List<String> droppedTables = new ArrayList();
+        try (final java.sql.Connection dbConnection = getDbConnection();
+        ) {
+            int totalCount = 100;
```
simranjeetc: @dermot-hardy Although I have tested with a count of 30,000 tables with this test, I am checking in a smaller number. I can add a bigger number here if we are interested in this test to run with a bigger number at every check-in.

job-service-acceptance-tests/src/test/java/com/hpe/caf/jobservice/acceptance/JobServiceDatabaseUtil.java
```diff
@@ -280,4 +280,15 @@ private static String getPropertyOrEnvVar(final String key)
...
+    {
+        try(final Connection dbConnection = getDbConnection();
+            final PreparedStatement st = dbConnection.prepareStatement("SELECT count(*) as result FROM public.delete_log ");
+            final ResultSet rs = st.executeQuery())
+        {
+            rs.next();
+            Assert.assertNotEquals(rs.getInt("result"), 0, "Soft deleted table names not present. ");
+        }
+    }
+}
```
dermot-hardy: Red no-entry sign = Take a newline

job-service-container/pom.xml
```diff
@@ -41,6 +41,7 @@
         <caf.codec.impl>codec-json</caf.codec.impl>
         <caf.worker-queue.impl>worker-queue-rabbit</caf.worker-queue.impl>
         <dummyUrl>http://self:1234/job-service/v1</dummyUrl>
+        <postgres.db.port>54328</postgres.db.port>
```
dermot-hardy: Why would we have a hard-coded port?  Surely this will mean that multiple builds can't run simultaneously.  And what's this change got to do with job cancellation?

job-service-container/pom.xml
```diff
@@ -161,7 +162,7 @@
                 <webservice.https.adminport>25083</webservice.https.adminport>
                 <rabbitmq.node.port>5672</rabbitmq.node.port>
                 <rabbitmq.ctrl.port>15672</rabbitmq.ctrl.port>
-                <postgres.db.port>5432</postgres.db.port>
+                <postgres.db.port>${postgres.db.port}</postgres.db.port>
```
dermot-hardy: Why is this changing?

dermot-hardy (changes requested):

job-service-container/src/test/java/com/hpe/caf/services/job/api/JobServiceIT.java
dermot-hardy: I would have thought that it should be a larger number than was possible before but I don't know what the cost is.  If it takes significantly longer then 100 is fine.

job-service-db/src/main/resources/v3.5.0/changelog.xml
```diff
@@ -45,4 +45,14 @@
             <column name="dependent_job_id" />
         </createIndex>
     </changeSet>
+    <changeSet id="create_table_delete_log" author="Gamma Team">
```
dermot-hardy: The worker is on `4.0.0-SNAPSHOT` so why would we be modifying the `v3.5.0` changelog?  You're not meant to change these files after the release.

job-service-db/src/main/resources/v3.5.0/changelog.xml
```diff
@@ -45,4 +45,14 @@
...
+        <createTable tableName="delete_log">
+            <column name="table_name" type="VARCHAR(255)">
+                <constraints nullable="false" />
+            </column>
+        </createTable>
+        <createIndex schemaName="public" tableName="delete_log" indexName="idx_delete_log_table_name">
```
dermot-hardy: Is this index required?  Do we lookup by table name?  (I'll probably see the code in a later file.)  It seems strange that we have this but don't have a primary key (but not necessarily incorrect).

job-service-db/src/main/resources/v3.5.0/changelog.xml
```diff
@@ -45,4 +45,14 @@
             <column name="dependent_job_id" />
         </createIndex>
     </changeSet>
+    <changeSet id="create_table_delete_log" author="Gamma Team">
+        <createTable tableName="delete_log">
+            <column name="table_name" type="VARCHAR(255)">
```
dermot-hardy: According to https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS the maximum table name length is 63.  Do we need 255?

job-service-db/src/main/resources/procedures/dropTables.sql
```diff
@@ -0,0 +1,46 @@
...
+    WHILE EXISTS (SELECT 1 FROM delete_log) LOOP
+            FOR rec IN EXECUTE selected_table_names
+                LOOP
+                    EXECUTE 'DROP TABLE IF EXISTS ' ||  quote_ident(rec.table_name);
+                    DELETE FROM DELETE_LOG WHERE TABLE_NAME = rec.table_name;
+                END LOOP;
+            COMMIT;
+        END LOOP;
+END
+$$;
```
dermot-hardy: Red no-entry sign = missing newline

job-service-db/src/main/resources/procedures/changelog.xml
```diff
@@ -295,4 +295,16 @@
                          schemaName="public">
         </createProcedure>
     </changeSet>
+    <changeSet id="drop_tables_proc" runOnChange="true" author="Gamma Team">
```
dermot-hardy: Follow the same indentation policy as the rest of the file (i.e. procedureName under path etc.)

job-service-db/src/main/resources/procedures/dropTaskTables.sql
```diff
@@ -23,6 +23,10 @@
 DROP FUNCTION IF EXISTS internal_drop_task_tables(
     in_short_task_id VARCHAR(58)
 );
+DROP FUNCTION IF EXISTS internal_drop_task_tables(
+    in_partition_id VARCHAR(40),
+    in_task_id VARCHAR(58)
+);
```
dermot-hardy: Is this required with `CREATE OR REPLACE FUNCTION`?

job-service-db/src/main/resources/procedures/dropTaskTables.sql
```diff
@@ -47,8 +52,9 @@ BEGIN
             PERFORM internal_drop_task_tables(in_partition_id, in_task_id || subtask_suffix);
         END LOOP;
 
-        -- Drop the table itself
-        EXECUTE 'DROP TABLE ' || task_table_ident;
+        -- Insert table name to be dropped later
+        task_table_name := (REPLACE(task_table_ident, '"', ''))::varchar;
```
dermot-hardy: This seems odd.  Couldn't you have used the result from the `internal_get_task_table_name()` call above?

job-service-db/src/main/resources/procedures/dropTaskTables.sql
```diff
@@ -47,8 +52,9 @@ BEGIN
...
+        PERFORM insert_delete_log(task_table_name);
```
dermot-hardy: So this is an internal call.  I believe the convention is that it should start with `internal_` (unless it is also called publicly somewhere else).

job-service-db/src/main/resources/procedures/insertDeleteLog.sql
```diff
@@ -0,0 +1,37 @@
...
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+--
+
+/*
+ *  Name: internal_drop_task_tables
+ *
+ *  Description:
+ *  Drops all task tables belonging to the specified task and all its subtasks
```
dermot-hardy: Completely wrong

job-service-db/src/main/resources/procedures/insertDeleteLog.sql
```diff
@@ -0,0 +1,37 @@
...
+
+CREATE OR REPLACE FUNCTION insert_delete_log(
+    task_table_name VARCHAR
+)
+RETURNS VOID
+LANGUAGE plpgsql
+AS $$
+DECLARE
+BEGIN
+        INSERT INTO public.delete_log VALUES ( task_table_name );
```
dermot-hardy: Is it even worth having this function?

job-service-db/src/main/resources/procedures/dropTables.sql
```diff
@@ -0,0 +1,46 @@
...
+CREATE OR REPLACE PROCEDURE drop_tables()
+    LANGUAGE plpgsql
+AS $$
+DECLARE
+    selected_table_names VARCHAR;
+    commit_limit integer:=10;
+    rec record;
+
+BEGIN
+    selected_table_names := $q$SELECT TABLE_NAME FROM DELETE_LOG LIMIT $q$;
```
dermot-hardy: I believe the convention in the project is that table and column names are lower-case.

job-service-db/src/main/resources/procedures/dropTables.sql
```diff
@@ -0,0 +1,46 @@
...
+    selected_table_names := selected_table_names || commit_limit || $q$ FOR UPDATE SKIP LOCKED$q$;
```
dermot-hardy: It's quite confusing the way that has been written.

```
x=6
x=x+5
```

Why not just write `x=6+5`?

job-service-db/src/main/resources/procedures/dropTables.sql
```diff
@@ -0,0 +1,46 @@
...
+
+    WHILE EXISTS (SELECT 1 FROM delete_log) LOOP
+            FOR rec IN EXECUTE selected_table_names
+                LOOP
+                    EXECUTE 'DROP TABLE IF EXISTS ' ||  quote_ident(rec.table_name);
+                    DELETE FROM DELETE_LOG WHERE TABLE_NAME = rec.table_name;
+                END LOOP;
```
dermot-hardy: Pull `LOOP` and `END LOOP` back a level - they are the braces of the `FOR` statement.

job-service-db/src/main/resources/procedures/dropTables.sql
```diff
@@ -0,0 +1,46 @@
...
+
+/*
+ *  Name: drop_tables
+ *
+ *  Description:
+ *  Drops all tables present in delete_log table
+ *  This procedure does batch commits. The batch is defined by commit_limit variable. Default batch size being 10.
+ */
+
+CREATE OR REPLACE PROCEDURE drop_tables()
```
dermot-hardy: Can we call this `drop_deleted_task_tables`?  `drop_tables` seems very generic.

dermot-hardy (changes requested):
Non-Java code reviewed.

job-service-acceptance-tests/src/test/java/com/hpe/caf/jobservice/acceptance/JobServiceDatabaseUtil.java
simranjeetc: Resolved

job-service-db/src/main/resources/v3.5.0/changelog.xml
simranjeetc: Resolved

job-service-db/src/main/resources/v3.5.0/changelog.xml
simranjeetc: Look up is being done [here](https://github.com/JobService/job-service/blob/f3aa0051b5a21e7c028e025e603d179efd564e54/job-service-db/src/main/resources/procedures/dropTables.sql#L40). I did think about adding a primary key column but did not find a compelling reason to do so.

job-service-db/src/main/resources/v3.5.0/changelog.xml
simranjeetc: Changed it to 63.

job-service-db/src/main/resources/procedures/dropTables.sql
simranjeetc: Added

job-service-db/src/main/resources/procedures/changelog.xml
simranjeetc: Corrected indentation

job-service-db/src/main/resources/procedures/insertDeleteLog.sql
simranjeetc: I extracted it out to use it for testing [here](https://github.com/JobService/job-service/blob/f3aa0051b5a21e7c028e025e603d179efd564e54/job-service-container/src/test/java/com/hpe/caf/services/job/api/JobServiceIT.java#L1159).

job-service-db/src/main/resources/procedures/dropTables.sql
simranjeetc: Changed it to [this](https://github.com/JobService/job-service/blob/f3aa0051b5a21e7c028e025e603d179efd564e54/job-service-db/src/main/resources/procedures/dropTables.sql#L34)
`selected_table_names := $q$SELECT TABLE_NAME FROM DELETE_LOG LIMIT $q$ || commit_limit || $q$ FOR UPDATE SKIP LOCKED$q$;
`

job-service-db/src/main/resources/procedures/dropTables.sql
simranjeetc: I kept it generic so that the same piece of code can be reused in the future. But I am fine changing it if you think this needs to be specifically named.

job-service-db/src/main/resources/procedures/dropTables.sql
simranjeetc: Resolved

job-service-db/src/main/resources/procedures/dropTables.sql
simranjeetc: Resolved

job-service-db/src/main/resources/procedures/dropTables.sql
dermot-hardy: Yes rename it.

job-service-db/src/main/resources/v3.5.0/changelog.xml
dermot-hardy: Ok.

job-service-container/src/test/java/com/hpe/caf/services/job/api/JobServiceIT.java
simranjeetc: image

Above screenshot is from jenkins [job](http://sou-jenkins2.swinfra.net/job/JobService/view/Developer/job/JobService~job-service~SCMOD-10825~CI/51). The creation of 30k tables takes approx 5 minutes. Dropping those tables is taking 75 seconds. I checked these numbers for two runs. So, keeping a count of 30k would increase the build time by ~ 6 minutes.

job-service-container/pom.xml
```diff
@@ -151,6 +152,11 @@
             <artifactId>job-service-util</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-simple</artifactId>
+            <scope>test</scope>
+        </dependency>
```
simranjeetc, line 159: Added this dependency to make sure tests are logging.

job-service-container/src/test/java/com/hpe/caf/services/job/api/JobServiceIT.java
dermot-hardy: Ok do the full check for the release build and the quick check for the CI build.

job-service-container/pom.xml
simranjeetc: I was using this change to connect to a fixed port in my local while development. This was pushed by mistake. Reverted.

job-service-container/pom.xml
simranjeetc: I was using this change to connect to a fixed port in my local while development. This was pushed by mistake. Reverted.

job-service-db/src/main/resources/procedures/dropTables.sql
simranjeetc: Renamed.

job-service-container/src/test/java/com/hpe/caf/services/job/api/JobServiceIT.java
simranjeetc: Made changes to exclude the long running IT from CI builds and to include it in release builds only.

job-service-container/pom.xml
```diff
@@ -168,6 +173,34 @@
...
         </profile>
+        <profile>
+            <id>nightly-test</id>
+            <activation>
+                <property>
+                    <name>RE_BUILD_TYPE</name>
+                    <value>release</value>
+                </property>
+            </activation>
+        </profile>
```
dermot-hardy: What is this profile meant to do?  There's nothing in it.

job-service-container/pom.xml
```diff
@@ -168,6 +173,34 @@
...
+        <profile>
+            <id>continuous</id>
+            <activation>
+                <activeByDefault>true</activeByDefault>
+            </activation>
```
dermot-hardy: What is the idea here?  If this is activated by default then why not jut put it in the standard failsafe configuration?

job-service-container/pom.xml
```diff
@@ -232,6 +265,10 @@
                         <CAF_STATUS_CHECK_INTERVAL_SECONDS>5</CAF_STATUS_CHECK_INTERVAL_SECONDS>
                         <CAF_INTEGRATIONTESTS_FS_PATH>\\${docker.host.address}@${webdav.apache.port}\webdav</CAF_INTEGRATIONTESTS_FS_PATH>
                         <CAF_WEBSERVICE_URL>${dummyUrl}</CAF_WEBSERVICE_URL>
+                        <JOB_SERVICE_DATABASE_APPNAME>Job Service</JOB_SERVICE_DATABASE_APPNAME>
+                        <JOB_SERVICE_DATABASE_URL>jdbc:postgresql://${docker.host.address}:${postgres.db.port}/${db.name}</JOB_SERVICE_DATABASE_URL>
+                        <JOB_SERVICE_DATABASE_USERNAME>postgres</JOB_SERVICE_DATABASE_USERNAME>
+                        <JOB_SERVICE_DATABASE_PASSWORD>root</JOB_SERVICE_DATABASE_PASSWORD>
```
dermot-hardy, line 257: Are we adding new tests that require direct access to the database?

job-service-db/src/main/resources/procedures/changelog.xml
```diff
@@ -295,4 +295,16 @@
                          schemaName="public">
         </createProcedure>
     </changeSet>
+    <changeSet id="drop_deleted_task_tables_proc" runOnChange="true" author="Gamma Team">
+        <createProcedure path="procedures/dropTables.sql"
```
dermot-hardy: The procedure filename should align with the actual name.

job-service-db/src/main/resources/procedures/dropTaskTables.sql
```diff
@@ -47,8 +48,9 @@ BEGIN
             PERFORM internal_drop_task_tables(in_partition_id, in_task_id || subtask_suffix);
         END LOOP;
 
-        -- Drop the table itself
-        EXECUTE 'DROP TABLE ' || task_table_ident;
+        -- Insert table name to be dropped later
+        task_table_name := internal_get_task_table_name(in_partition_id, in_task_id);
```
dermot-hardy: We shouldn't be calling `internal_get_task_table_name(in_partition_id, in_task_id)` twice in the same function.  You should just set the `task_table_name` and then use it when setting `task_table_ident`.

job-service-db/src/main/resources/procedures/insertDeleteLog.sql
```diff
@@ -0,0 +1,37 @@
...
+/*
+ *  Name: internal_insert_delete_log
+ *
+ *  Description:
+ *  This is used to insert a record into delete log table.
+ */
+DROP FUNCTION IF EXISTS internal_insert_delete_log(
+    task_table_name VARCHAR
+);
+
```
dermot-hardy: Is this required with `CREATE OR REPLACE FUNCTION`?

job-service-db/src/main/resources/procedures/insertDeleteLog.sql
```diff
@@ -0,0 +1,37 @@
...
+CREATE OR REPLACE FUNCTION internal_insert_delete_log(
+    task_table_name VARCHAR
+)
+RETURNS VOID
+LANGUAGE plpgsql
```
dermot-hardy: I know `VOLATILE` is the default but I'd prefer if we were always explicit w.r.t. the volatility category (to show that we've thought about it).
```suggestion
LANGUAGE plpgsql VOLATILE
```

job-service-db/src/main/resources/procedures/insertDeleteLog.sql
```diff
@@ -0,0 +1,37 @@
...
+AS $$
+DECLARE
```
dermot-hardy: There are no variables being declared.
```suggestion
```

dermot-hardy (changes requested):

job-service-db/src/main/resources/procedures/insertDeleteLog.sql
dermot-hardy: Removed in commit 6dcba54.

job-service-db/src/main/resources/procedures/dropTables.sql
```diff
@@ -0,0 +1,45 @@
...
+
+/*
+ *  Name: drop_deleted_task_tables
+ *
+ *  Description:
+ *  Drops all tables present in delete_log table
+ *  This procedure does batch commits. The batch is defined by commit_limit variable. Default batch size being 10.
+ */
+CREATE OR REPLACE PROCEDURE drop_deleted_task_tables()
+LANGUAGE plpgsql
```
dermot-hardy: Should volatility categories be specified for procedures?  Or does the concept only exist for functions?

dermot-hardy: Don't do it in this PR but now that we have both procedures and functions I'd be tempted to have a follow on separate them out into different folders.  I can easily see the distinction being lost with them all just sitting in the same folder.

job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/DatabasePoller.java
```diff
@@ -36,7 +32,8 @@
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
-import java.util.Properties;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
```
dermot-hardy: Moving these imports should not be in this PR.

job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/DBConnection.java
```diff
@@ -0,0 +1,93 @@
...
+import java.sql.SQLException;
+import java.text.MessageFormat;
+import java.util.Properties;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DBConnection
+{
+    private static final Logger LOG = LoggerFactory.getLogger(DatabasePoller.class);
```
dermot-hardy:
```suggestion
    private static final Logger LOG = LoggerFactory.getLogger(DBConnection.class);
```

job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/DBConnection.java
```diff
@@ -0,0 +1,93 @@
...
+            SCMOD-6525 - FALSE POSITIVE on FORTIFY SCAN for Log forging. The values of databaseUrl, dbUser, dbPass are all set using
+            properties or env variables.
+            */
+            LOG.error(errorMessage);
+            throw new ScheduledExecutorException(errorMessage);
+        }
+        
+        return conn;
+    }
+}
```
dermot-hardy: Re no-entry sign = missing newline

job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/ScheduledExecutor.java
```diff
@@ -15,21 +15,20 @@
...
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import java.util.Arrays;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
```
dermot-hardy: Imports should not be moved except in a commit expressly for the purpose of organizing them.

job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/DropTablesTask.java
```diff
@@ -0,0 +1,68 @@
...
+import java.sql.SQLException;
+import java.text.MessageFormat;
+import java.time.Duration;
+import java.time.Instant;
+import java.util.Arrays;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DropTablesTask implements Runnable
```
dermot-hardy: Unless there is a requirement for a class to be overridden then make it `final`.

```suggestion
public final class DropTablesTask implements Runnable
```

job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/DropTablesTask.java
```diff
@@ -0,0 +1,68 @@
...
+{
+    private static final Logger LOG = LoggerFactory.getLogger(DropTablesTask.class);
+    
+    @Override
+    public void run()
+    {
+        {
```
dermot-hardy, line 35: Pointless level of braces

job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/DropTablesTask.java
```diff
@@ -0,0 +1,68 @@
...
+                    LOG.debug("Total time taken to drop tables in ms. "+ Duration.between(start, end).toMillis());
+                    
+                }
+                catch(final SQLException e)
+                {
+                    final String errorMessage = MessageFormat
+                            .format("Failed in call to drop_deleted_task_tables() database procedure.{0}",
+                                    e.getMessage());
+                    LOG.error(errorMessage);
+                    throw new ScheduledExecutorException(errorMessage);
```
dermot-hardy: What is the purpose of throwing this here for it to be caught and re-logged?  This is just causing double-logging.

job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/DropTablesTask.java
```diff
@@ -0,0 +1,68 @@
...
+                }
```
dermot-hardy: I can't see any purpose to this catch block.  It isn't logging anything more than the general catch block you have.

job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/DropTablesTask.java
```diff
@@ -0,0 +1,68 @@
...
+    
+    @Override
+    public void run()
+    {
+        {
+            try
+            {
+                try(
+                        Connection connection = DBConnection.get();
+                        PreparedStatement stmt = connection.prepareStatement("CALL  drop_deleted_task_tables()")
```
dermot-hardy: Variables should almost always be `final`.  (Same for `start` and `end` below.)

```suggestion
                        final Connection connection = DBConnection.get();
                        final PreparedStatement stmt = connection.prepareStatement("CALL  drop_deleted_task_tables()")
```

job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/DropTablesTask.java
```diff
@@ -0,0 +1,68 @@
...
+                )
+                {
+                    LOG.debug("Calling drop_deleted_task_tables() database procedure ...");
+                    Instant start = Instant.now();
+                    stmt.execute();
+                    Instant end = Instant.now();
+                    LOG.debug("Total time taken to drop tables in ms. "+ Duration.between(start, end).toMillis());
```
dermot-hardy: I'm not strongly opinionated on it given the amount of extra work you're doing for the debug logging it might be worth doing this:

```suggestion
                    if (LOG.isDebugEnabled()) {
                        LOG.debug("Calling drop_deleted_task_tables() database procedure ...");
                        final Instant start = Instant.now();
                        stmt.execute();
                        final Instant end = Instant.now();
                        LOG.debug("Total time taken to drop tables in ms. " + Duration.between(start, end).toMillis());
                    } else {
                        stmt.execute();
                    }
```

job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/DropTablesTask.java
```diff
@@ -0,0 +1,68 @@
...
+            }
+            catch(final Throwable t)
+            {
+                LOG.error("Caught exception while dropping soft deleted tables. Message:\n" + t
+                        .getMessage()
+                        + "StackTrace:\n" + Arrays.toString(t.getStackTrace()));
+            }
+        }
+    }
+}
```
dermot-hardy: Red no-entry sign = missing newline.

dermot-hardy (changes requested):

job-service-container/pom.xml
simranjeetc: Yes. https://github.com/JobService/job-service/pull/192/files#diff-3800e074160deaca30cf6b85c3d8e2852de68b0a0c2837f667c9ed73975ed388R43

job-service-db/src/main/resources/procedures/dropTaskTables.sql
simranjeetc: Addressed

job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/DropTablesTask.java
simranjeetc: Removed the inner catch block

job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/DropTablesTask.java
simranjeetc: Removed the inner catch block

job-service-db/src/main/resources/procedures/dropTables.sql
simranjeetc: I did not find any mention of volatility categories w.r.t procedures in PG documentation. Got below error if I create a procedure with volatile keyword.
image
The screenshot does not show the cursor but it pointed at the volatile keyword at the time of error.

job-service-db/src/main/resources/procedures/changelog.xml
simranjeetc: Renamed

job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/DatabasePoller.java
simranjeetc: Reverted

job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/ScheduledExecutor.java
simranjeetc: Reverted

job-service-container/pom.xml
simranjeetc: This profile was not meant to do anything. The would make pom execute all the tests including the performance IT. This was possible because of a default activation profile(continuous) being present which would exclude the performance IT. So, the effective-pom for release profile had no exclusion whereas the default profile had an exclusion.

But I liked the part where we can have the exclusion as part of plugin's configuration(as suggested by you). So, now the release profile will just override the "excluded" tag's configuration to an empty exclude tag. You can refer the effective-poms generated by help plugin in both the cases below - 

> mvn help:effective-pom 

> i.e. default configuration. This should effectively **exclude** the performance  IT during execution.

```xml
            <configuration>
              <systemPropertyVariables>
                <webservice.adminport>${webservice.adminport}</webservice.adminport>
                <docker.host.address>${docker.host.address}</docker.host.address>
                <rabbitmq.node.port>${rabbitmq.node.port}</rabbitmq.node.port>
                <rabbitmq.ctrl.port>${rabbitmq.ctrl.port}</rabbitmq.ctrl.port>
              </systemPropertyVariables>
              <environmentVariables>
                <CAF_APPNAME>test/jobservice</CAF_APPNAME>
                <CAF_CONFIG_PATH>/home/simran/codebase/mf/JobService/job-service/job-service-container/test-config</CAF_CONFIG_PATH>
                <CAF_TRACKING_PIPE>tracking-workflow-in</CAF_TRACKING_PIPE>
                <CAF_STATUS_CHECK_INTERVAL_SECONDS>5</CAF_STATUS_CHECK_INTERVAL_SECONDS>
                <CAF_INTEGRATIONTESTS_FS_PATH>\\${docker.host.address}@${webdav.apache.port}\webdav</CAF_INTEGRATIONTESTS_FS_PATH>
                <CAF_WEBSERVICE_URL>http://self:1234/job-service/v1</CAF_WEBSERVICE_URL>
                <JOB_SERVICE_DATABASE_APPNAME>Job Service</JOB_SERVICE_DATABASE_APPNAME>
                <JOB_SERVICE_DATABASE_URL>jdbc:postgresql://${docker.host.address}:${postgres.db.port}/jobservice</JOB_SERVICE_DATABASE_URL>
                <JOB_SERVICE_DATABASE_USERNAME>postgres</JOB_SERVICE_DATABASE_USERNAME>
                <JOB_SERVICE_DATABASE_PASSWORD>root</JOB_SERVICE_DATABASE_PASSWORD>
                <webserviceurl>http://${docker.host.address}:${webservice.adminport}/job-service/v1</webserviceurl>
                <healthcheckurl>http://${docker.host.address}:${webservice.healthcheckport}/healthcheck</healthcheckurl>
                <webserviceurlhttps>https://${docker.host.address}:${webservice.https.adminport}/job-service/v1</webserviceurlhttps>
              </environmentVariables>
              <excludes>
                <exclude>com.hpe.caf.services.job.api.performance.JobServicePerformanceIT</exclude>
              </excludes>
            </configuration>
```



> mvn help:effective-pom -P nightly-test

> This should effectively **include** the performance IT during execution.

```xml
           <configuration>
              <excludes combine.self="override">
                <exclude />
              </excludes>
              <systemPropertyVariables>
                <webservice.adminport>${webservice.adminport}</webservice.adminport>
                <docker.host.address>${docker.host.address}</docker.host.address>
                <rabbitmq.node.port>${rabbitmq.node.port}</rabbitmq.node.port>
                <rabbitmq.ctrl.port>${rabbitmq.ctrl.port}</rabbitmq.ctrl.port>
              </systemPropertyVariables>
              <environmentVariables>
                <CAF_APPNAME>test/jobservice</CAF_APPNAME>
                <CAF_CONFIG_PATH>/home/simran/codebase/mf/JobService/job-service/job-service-container/test-config</CAF_CONFIG_PATH>
                <CAF_TRACKING_PIPE>tracking-workflow-in</CAF_TRACKING_PIPE>
                <CAF_STATUS_CHECK_INTERVAL_SECONDS>5</CAF_STATUS_CHECK_INTERVAL_SECONDS>
                <CAF_INTEGRATIONTESTS_FS_PATH>\\${docker.host.address}@${webdav.apache.port}\webdav</CAF_INTEGRATIONTESTS_FS_PATH>
                <CAF_WEBSERVICE_URL>http://self:1234/job-service/v1</CAF_WEBSERVICE_URL>
                <JOB_SERVICE_DATABASE_APPNAME>Job Service</JOB_SERVICE_DATABASE_APPNAME>
                <JOB_SERVICE_DATABASE_URL>jdbc:postgresql://${docker.host.address}:${postgres.db.port}/jobservice</JOB_SERVICE_DATABASE_URL>
                <JOB_SERVICE_DATABASE_USERNAME>postgres</JOB_SERVICE_DATABASE_USERNAME>
                <JOB_SERVICE_DATABASE_PASSWORD>root</JOB_SERVICE_DATABASE_PASSWORD>
                <webserviceurl>http://${docker.host.address}:${webservice.adminport}/job-service/v1</webserviceurl>
                <healthcheckurl>http://${docker.host.address}:${webservice.healthcheckport}/healthcheck</healthcheckurl>
                <webserviceurlhttps>https://${docker.host.address}:${webservice.https.adminport}/job-service/v1</webserviceurlhttps>
              </environmentVariables>
            </configuration>
```

job-service-container/pom.xml
simranjeetc: Resolving this comment as the discussion is captured in [this](https://github.com/JobService/job-service/pull/192#discussion_r603910575) comment.

job-service-container/pom.xml
dermot-hardy: I don't think we should be completely excluding the test by default.  What I was expecting was the full 30,000 iteration check to be done for the release build and a quick 100 iteration check for the CI build.  The way I'd expect this to be implemented is to have a property in the pom file that is 100 by default (in the normal properties section) and 30,000 if the nightly-test profile is activated (i.e. by having a properties section in nightly-test profile).  The value of the property should be passed to the integration tests in much the same way that the `webservice.adminport` etc. are (i.e. as a system property or an environment variable).  Within the test you should then use [`System.getProperty(key)`](https://docs.oracle.com/javase/7/docs/api/java/lang/System.html#getProperty(java.lang.String)) or [`System.getenv(name)`](https://docs.oracle.com/javase/7/docs/api/java/lang/System.html#getenv(java.lang.String)) to retrieve the number of iterations to perform.

dermot-hardy: P.S. I do like your exclusion/inclusion override as you have it now.  Even though I don't want to use it in this case I could imagine it being applicable in future cases across the product.

job-service-container/src/test/java/com/hpe/caf/services/job/api/performance/JobServicePerformanceIT.java
```diff
@@ -0,0 +1,121 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(JobServicePerformanceIT.class);
+    
+    @Test
+    public void testDeleteLog() throws SQLException
+    {
+        //prepare
+        List<String> droppedTables = new ArrayList();
+        try(final java.sql.Connection dbConnection = JobServiceConnectionUtil.getDbConnection())
+        {
+            int totalCount = 30000;
```
dermot-hardy: As discussed above.

dermot-hardy (changes requested):

job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/DBConnection.java
```diff
@@ -0,0 +1,94 @@
...
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.SQLException;
+import java.text.MessageFormat;
+import java.util.Properties;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DBConnection
```
dermot-hardy: Unless there is a requirement for a class to be overridden then make it `final`.

```suggestion
public final class DBConnection
```

This should be considered for all new classes that you are introducing.

job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/DBConnection.java
```diff
@@ -0,0 +1,94 @@
...
+{
+    private static final Logger LOG = LoggerFactory.getLogger(DBConnection.class);
+    
+    private static final String JDBC_POSTGRESQL_PREFIX = "jdbc:postgresql:";
+    private static final String JDBC_DRIVER = "org.postgresql.Driver";
+
```
dermot-hardy, line 37: This is actually a utility class so it should have an empty private default constructor (to prevent accidental instantiation).

dermot-hardy (changes requested):

job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/DBConnection.java
simranjeetc: Resolved

job-service-container/pom.xml
simranjeetc: Added a property task.table.deletion.count with default value 100 in properties section. This value gets overridden to 30000 if nightly-test profile is activated. Also moved the test back into JobServiceIT(deleted JobServicePerformanceIT) class since this test is no longer a performance only test and will be getting executed on CI builds with smaller count.

dermot-hardy (approved):

PR #193, opened 2020-10-27
SCMOD-10455 / 3 collapsing items from the completed_subtask_report table

xbreizh:


job-service-db/src/main/resources/procedures/updateJobProgress.sql
```diff
@@ -58,10 +58,14 @@ BEGIN
         RETURNING csr.task_id
     )
 
+    -- converting table into array
     SELECT array_agg(task_id)
     FROM completed_subtask
     INTO subtask_array;
 
+    -- collapsing whenever possible
+    select * from task_collapse(subtask_array) into subtask_array;
```
dermot-hardy: Capitalize sql keywords

job-service-db/src/main/resources/procedures/taskCollapse.sql
```diff
@@ -0,0 +1,127 @@
+CREATE OR REPLACE FUNCTION task_collapse(tasks VARCHAR[])
```
dermot-hardy, line 23: - Missing copyright header
- This function cannot go in without a good set of unit tests.
- Also we'll need some performance tests done on this before checking it in.

job-service-db/src/main/resources/procedures/taskCollapse.sql
```diff
@@ -0,0 +1,127 @@
...
+                END LOOP;
+
+            -- if no modification done, then we break the loop
+            IF ( final_array = backup ) THEN
+
+                modified='f';
+
+            ELSE
+                -- sorting tasks
+                select array(SELECT id FROM unnest( final_array ) as id order BY substring(id,regex_body), substring(id, regex_last_nb)::INTEGER desc ) into tasks;
```
dermot-hardy: Goes over the 138-character guideline

job-service-db/src/main/resources/procedures/taskCollapse.sql
```diff
@@ -0,0 +1,127 @@
...
+    -- tasks := array_agg(x ORDER BY x DESC) FROM unnest(tasks) x ;
+    SELECT array(
+        SELECT id FROM unnest( tasks ) AS id
+        ORDER BY substring(id,regex_body),
+                 substring(id, regex_last_nb)::INTEGER DESC
+        )
+    INTO tasks;
+
+    -- we loop until the array stop being modified
+    WHILE (modified='t') LOOP
```
dermot-hardy: Can this be simplified?  `== true` is unusual in a while (in general).

job-service-db/src/main/resources/procedures/taskCollapse.sql
```diff
@@ -0,0 +1,127 @@
...
+
+
+            --we loop until the tasks array is empty
+            WHILE (cardinality(tasks)!=0) LOOP
+
+                    -- we store the first element of the array
+                    task = tasks[1];
+
+                    -- if there's no more task, then we return the final_array
+                    IF task IS NULL THEN RETURN final_array; END IF;
```
dermot-hardy: Looks like this could never be hit given the loop condition.

(Or if NULLs are being put into the array then is this intentional?)

dermot-hardy (changes requested):

job-service-db/src/main/resources/procedures/updateJobProgress.sql
xbreizh: done

job-service-db/src/main/resources/procedures/taskCollapse.sql
xbreizh: done

job-service-db/src/main/resources/procedures/taskCollapse.sql
xbreizh: Done for true

job-service-db/src/main/resources/procedures/taskCollapse.sql
xbreizh: removed null checking and maxi

job-service-db/src/main/resources/procedures/taskCollapse.sql
xbreizh: - [ X ] copyright ok
- [ X ] unit tests added 
- [ X ] perf test: 

   - Inserted some lines in the completed_subtask-report table
   - Tested without the function, then with it

```
For 65318 lines Before  2541.163 ms / After 61631.399 ms
```

It slows down the process

`

xbreizh: Actually, I am unsure if the task list passed in the test is really relevant. I will try to do it again with the list generated from one of the last tests. Also, the algorithm can be improved for sure.

For example, instead of checking all elements one by one when a star appears, we could try to "predict". 
If we have 4* for tasks[1] then we check if tasks[4] has the same body and ends with 1. 

If so, we collapse. Otherwise, we skip to the next "*"

PR #194, opened 2020-11-12
SCMOD-10455 fix database shutdown issue

xbreizh:
- closing postgres connection 
- increased postgres waiting time and shutdown


dermot-hardy (changes requested):
As discussed, this is not correct.

PR #199, opened 2020-11-28
SCMOD-11843: Add local lock_timeout

xbreizh:
https://portal.digitalsafe.net/browse/SCMOD-11843


xbreizh:
Add local lock_timeout would delay the triggering of the  deadlock. only for the function.
We could also do it on the db directly more permanently. For example, I saw that on larry, we use the default 1s which is supposed to be the very minimum to use  (show deadlock_timeout;). I don't know about Beta

The other idea would be to use the task_collapse function that should reduce the locking time

PR #200, opened 2020-11-30
SCMOD-11843: Fix deadlock

xbreizh:
https://portal.digitalsafe.net/browse/SCMOD-11843

### DIFF
--- a/job-service-db/src/main/resources/procedures/getJobs.sql
+++ b/job-service-db/src/main/resources/procedures/getJobs.sql
@@ -192,7 +192,7 @@ BEGIN
     ALTER TABLE get_job_temp ADD COLUMN id SERIAL PRIMARY KEY;
 
     -- Create an array of job_id(s) based on the get_job_temp table
-    jobIdArray := ARRAY(SELECT DISTINCT (jt.job_id) FROM get_job_temp jt order by jt.job_id );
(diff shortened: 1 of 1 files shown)


dermot-hardy (changes requested):
This is part of what we talked about earlier.  I'm glad to see the tests failed because the way you have it at the minute will have broken sorting from the client's perspective.

job-service-db/src/main/resources/procedures/getJobs.sql
```diff
@@ -192,7 +192,7 @@ BEGIN
     ALTER TABLE get_job_temp ADD COLUMN id SERIAL PRIMARY KEY;
 
     -- Create an array of job_id(s) based on the get_job_temp table
-    jobIdArray := ARRAY(SELECT DISTINCT (jt.job_id) FROM get_job_temp jt);
+    jobIdArray := ARRAY(SELECT DISTINCT (jt.job_id) FROM get_job_temp jt order by jt.job_id );
```
dermot-hardy:
```suggestion
    jobIdArray := ARRAY(SELECT DISTINCT (jt.job_id) FROM get_job_temp jt ORDER BY jt.job_id);
```

dermot-hardy (approved):

PR #202, opened 2020-12-03
SCMOD-11919: Implement report_progress

xbreizh:
https://portal.digitalsafe.net/browse/SCMOD-11919

### DIFF
--- a/job-service-db/src/main/resources/procedures/reportProgress.sql
+++ b/job-service-db/src/main/resources/procedures/reportProgress.sql
@@ -30,7 +30,7 @@ CREATE OR REPLACE FUNCTION report_progress(
     in_percentage_complete DOUBLE PRECISION
 )
 RETURNS VOID
-LANGUAGE plsql
+LANGUAGE plpgsql
 AS $$
 DECLARE
     v_job_id VARCHAR(48);
--- a/job-service-db/src/main/resources/procedures/reportProgress.sql
+++ b/job-service-db/src/main/resources/procedures/reportProgress.sql
@@ -34,6 +34,7 @@ LANGUAGE plpgsql
 AS $$
 DECLARE
     v_job_id VARCHAR(48);
+    v_job_status job_status;
 
 BEGIN
     -- Raise exception if task identifier has not been specified
@@ -52,9 +53,9 @@ BEGIN
     -- Get the job status
     -- And take out an exclusive update lock on the job row
     SELECT status INTO v_job_status
-    FROM job j
-    WHERE j.partition_id = in_partition_id
-        AND j.job_id = v_job_id
+    FROM job
+    WHERE partition_id = in_partition_id
+      AND job_id = v_job_id
     FOR UPDATE;
 
     -- Check that the job hasn't been deleted, cancelled or completed
--- a/job-service-db/src/main/resources/procedures/reportProgress.sql
+++ b/job-service-db/src/main/resources/procedures/reportProgress.sql
@@ -30,7 +30,7 @@ CREATE OR REPLACE FUNCTION report_progress(
     in_percentage_complete DOUBLE PRECISION
 )
 RETURNS VOID
-LANGUAGE plpgsql
+LANGUAGE plsql
 AS $$
 DECLARE
     v_job_id VARCHAR(48);
--- a/job-service-db/src/main/resources/procedures/reportProgress.sql
+++ b/job-service-db/src/main/resources/procedures/reportProgress.sql
@@ -24,44 +24,46 @@ DROP FUNCTION IF EXISTS report_progress(
     in_task_id VARCHAR(58),
     in_status job_status
 );
-CREATE OR REPLACE PROCEDURE report_progress(
+CREATE OR REPLACE FUNCTION report_progress(
     in_partition_id VARCHAR(40),
-    in_job_id VARCHAR(58),
     in_task_id VARCHAR(58),
-    in_task_completion DOUBLE PRECISION
+    in_percentage_complete DOUBLE PRECISION
 )
-    LANGUAGE plpgsql
+RETURNS VOID
+LANGUAGE plpgsql
 AS $$
 DECLARE
-    in_status job_status;
+    v_job_id VARCHAR(48);
+    v_status job_status:='Active';
 BEGIN
 
-    -- Raise exception if the job identifier has not been specified
-    IF in_job_id IS NULL OR in_job_id = '' THEN
-        RAISE EXCEPTION 'Job identifier has not been specified' USING ERRCODE = '02000'; -- sqlstate no data
+    -- Raise exception if task progress is < 0 or > 100
+    IF in_percentage_complete < 0 OR in_percentage_complete > 100 THEN
+        RAISE EXCEPTION 'Invalid in_percentage_complete %', in_percentage_complete USING ERRCODE = '22023'; -- invalid_parameter_value
     END IF;
 
-    -- Exit the procedure if task progress is 0
-    IF in_task_completion = 0
-        THEN RETURN;
+    -- Round down if 100 value passed in
+    IF in_percentage_complete = 100 THEN
+       in_percentage_complete=99.9;
+    END IF;
+
+    -- Get the job id
+    v_job_id = internal_get_job_id(in_task_id);
+
+    -- Raise exception if the job identifier has not been specified
+    IF v_job_id IS NULL OR v_job_id = '' THEN
+        RAISE EXCEPTION 'Invalid Job identifier' USING ERRCODE = '02000'; -- sqlstate no data
     END IF;
 
     -- Take out an exclusive update lock on the job row
     PERFORM NULL
     FROM job j
     WHERE j.partition_id = in_partition_id
-      AND j.job_id = in_job_id
+      AND j.job_id = v_job_id
         FOR UPDATE;
 
-    IF in_task_completion = 100
-    THEN
-        in_status = 'Completed';
-    ELSE
-        in_status = 'Active';
-    END IF;
-
     -- Process outstanding job updates
-    PERFORM internal_report_task_status(in_partition_id, in_task_id , in_status, in_task_completion, NULL);
+    PERFORM internal_report_task_status(in_partition_id, in_task_id , v_status, in_percentage_complete, NULL);
 
 END
 $$;
--- a/job-service-db/src/main/resources/procedures/reportProgress.sql
+++ b/job-service-db/src/main/resources/procedures/reportProgress.sql
@@ -34,36 +34,35 @@ LANGUAGE plpgsql
 AS $$
 DECLARE
     v_job_id VARCHAR(48);
-    v_status job_status:='Active';
+
 BEGIN
+    -- Raise exception if task identifier has not been specified
+    IF in_task_id IS NULL OR in_task_id = '' THEN
+        RAISE EXCEPTION 'Task identifier has not been specified';
+    END IF;
 
     -- Raise exception if task progress is < 0 or > 100
     IF in_percentage_complete < 0 OR in_percentage_complete > 100 THEN
         RAISE EXCEPTION 'Invalid in_percentage_complete %', in_percentage_complete USING ERRCODE = '22023'; -- invalid_parameter_value
     END IF;
 
-    -- Round down if 100 value passed in
-    IF in_percentage_complete = 100 THEN
-       in_percentage_complete=99.9;
-    END IF;
-
     -- Get the job id
     v_job_id = internal_get_job_id(in_task_id);
 
-    -- Raise exception if the job identifier has not been specified
-    IF v_job_id IS NULL OR v_job_id = '' THEN
-        RAISE EXCEPTION 'Invalid Job identifier' USING ERRCODE = '02000'; -- sqlstate no data
-    END IF;
-
-    -- Take out an exclusive update lock on the job row
-    PERFORM NULL
+    -- Get the job status
+    -- And take out an exclusive update lock on the job row
+    SELECT status INTO v_job_status
     FROM job j
     WHERE j.partition_id = in_partition_id
-      AND j.job_id = v_job_id
-        FOR UPDATE;
+        AND j.job_id = v_job_id
+    FOR UPDATE;
 
-    -- Process outstanding job updates
-    PERFORM internal_report_task_status(in_partition_id, in_task_id , v_status, in_percentage_complete, NULL);
+    -- Check that the job hasn't been deleted, cancelled or completed
+    IF NOT FOUND OR v_job_status IN ('Cancelled', 'Completed') THEN
+        RETURN;
+    END IF;
 
+    -- Update the task tables
+    PERFORM internal_report_task_status(in_partition_id, in_task_id, 'Active', LEAST(in_percentage_complete, 99.9), NULL);
 END
 $$;
--- a/job-service-db/src/main/resources/procedures/reportProgress.sql
+++ b/job-service-db/src/main/resources/procedures/reportProgress.sql
@@ -17,9 +17,51 @@
 /*
  *  Name: report_progress
  *
- *  REMOVED
+ *  Description:
+ *  Reports a task progress to the database
  */
 DROP FUNCTION IF EXISTS report_progress(
     in_task_id VARCHAR(58),
     in_status job_status
 );
+CREATE OR REPLACE PROCEDURE report_progress(
+    in_partition_id VARCHAR(40),
+    in_job_id VARCHAR(58),
+    in_task_id VARCHAR(58),
+    in_task_completion DOUBLE PRECISION
+)
+    LANGUAGE plpgsql
+AS $$
+DECLARE
+    in_status job_status;
+BEGIN
+
+    -- Raise exception if the job identifier has not been specified
+    IF in_job_id IS NULL OR in_job_id = '' THEN
+        RAISE EXCEPTION 'Job identifier has not been specified' USING ERRCODE = '02000'; -- sqlstate no data
+    END IF;
+
+    -- Exit the procedure if task progress is 0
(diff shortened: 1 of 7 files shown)


job-service-db/src/main/resources/procedures/reportProgress.sql
```diff
@@ -17,9 +17,51 @@
...
+ *  Description:
+ *  Reports a task progress to the database
  */
 DROP FUNCTION IF EXISTS report_progress(
     in_task_id VARCHAR(58),
     in_status job_status
 );
+CREATE OR REPLACE PROCEDURE report_progress(
+    in_partition_id VARCHAR(40),
+    in_job_id VARCHAR(58),
```
dermot-hardy: Not needed - work it out from the task id.

job-service-db/src/main/resources/procedures/reportProgress.sql
```diff
@@ -17,9 +17,51 @@
...
+    in_task_id VARCHAR(58),
+    in_task_completion DOUBLE PRECISION
+)
+    LANGUAGE plpgsql
+AS $$
+DECLARE
+    in_status job_status;
```
dermot-hardy: The convention appears to be that `in_` is for parameters.

job-service-db/src/main/resources/procedures/reportProgress.sql
```diff
@@ -17,9 +17,51 @@
...
  */
 DROP FUNCTION IF EXISTS report_progress(
     in_task_id VARCHAR(58),
     in_status job_status
 );
+CREATE OR REPLACE PROCEDURE report_progress(
+    in_partition_id VARCHAR(40),
+    in_job_id VARCHAR(58),
+    in_task_id VARCHAR(58),
+    in_task_completion DOUBLE PRECISION
```
dermot-hardy: Rename to `in_percentage_complete`

job-service-db/src/main/resources/procedures/reportProgress.sql
```diff
@@ -17,9 +17,51 @@
...
+    -- Take out an exclusive update lock on the job row
+    PERFORM NULL
+    FROM job j
+    WHERE j.partition_id = in_partition_id
+      AND j.job_id = in_job_id
+        FOR UPDATE;
+
+    IF in_task_completion = 100
+    THEN
+        in_status = 'Completed';
```
dermot-hardy: I don't want this to be able to be used as an alternate to `reportComplete`.  If the caller provides 100% then round down to 99.9%.  If they provide <0% or >100% then raise an exception.

worker-jobtracking/src/main/java/com/hpe/caf/worker/jobtracking/JobTrackingWorkerReporter.java
```diff
@@ -99,14 +100,42 @@ public JobTrackingWorkerReporter() throws JobReportingException {
...
      *
-     * @param jobTaskId identifies the job task whose progress is to be reported
+     * @param jobTaskId identifies the job task in progress
      * @param estimatedPercentageCompleted an indication of progress on the job task
      * @throws JobReportingException if a failure occurs in connecting or reporting to a Job Database
      */
     @Override
     public void reportJobTaskProgress(final String jobTaskId, final int estimatedPercentageCompleted) throws JobReportingException
     {
-        LOG.trace("Received progress update message for task {}; taking no action", jobTaskId);
```
dermot-hardy, line 109: If the `estimatedPercentageCompleted` is 0% then don't bother making the database call.

worker-jobtracking/src/main/java/com/hpe/caf/worker/jobtracking/JobTrackingWorkerReporter.java
```diff
@@ -99,14 +100,42 @@ public JobTrackingWorkerReporter() throws JobReportingException {
...
+        LOG.trace("Received progress update message for task {} / {} .Progress: {};", taskId, partitionId, estimatedPercentageCompleted);
+
+        try (final Connection conn = getConnection()) {
+            try (final CallableStatement stmt = conn.prepareCall("{call report_progress(?,?,?,?)}")) {
+                stmt.setString(1, partitionId);
+                stmt.setString(2, jobId);
+                stmt.setString(3, taskId);
+                stmt.setDouble(4, estimatedPercentageCompleted);
+
+                executeToJobDependencyList(stmt);
```
dermot-hardy: The `report_progress` function shouldn't return a job dependency list.

job-service-db/src/main/resources/procedures/reportProgress.sql
```diff
@@ -17,9 +17,51 @@
...
     in_task_id VARCHAR(58),
     in_status job_status
 );
+CREATE OR REPLACE PROCEDURE report_progress(
+    in_partition_id VARCHAR(40),
+    in_job_id VARCHAR(58),
+    in_task_id VARCHAR(58),
+    in_task_completion DOUBLE PRECISION
+)
+    LANGUAGE plpgsql
```
dermot-hardy: Explicitly declare that the function doesn't return anything.

worker-jobtracking/src/main/java/com/hpe/caf/worker/jobtracking/JobTrackingWorkerReporter.java
```diff
@@ -99,14 +100,42 @@ public JobTrackingWorkerReporter() throws JobReportingException {
...
+            }
+        } catch (final SQLTransientException te) {
+            throw new JobReportingTransientException(
+                    MessageFormat.format(FAILED_TO_REPORT_COMPLETION, taskId, te.getMessage()), te);
```
dermot-hardy: `FAILED_TO_REPORT_COMPLETION`!

dermot-hardy (changes requested):

job-service-db/src/main/resources/procedures/reportProgress.sql
xbreizh: done

job-service-db/src/main/resources/procedures/reportProgress.sql
xbreizh: done

job-service-db/src/main/resources/procedures/reportProgress.sql
xbreizh: For the job id, I now use **internal_get_job_id()** to get it from the task but doesn't it add work in the function when we already have the info above?
The task doesn't provide the partition, so we need to pass it I think

job-service-db/src/main/resources/procedures/reportProgress.sql
xbreizh: Done, but won't we have an issue if we never get that missing 0.01 % ?

Shouldn't we check and re-route from **reportJobTasksProgress(final TrackingReportTask trackingReportTask)** ?
I added a simple **_log.warn**_ for now if showing 100 completion but status not "_Complete_"
`
_if(trackingReport.estimatedPercentageCompleted==100 && trackingReportStatus!= TrackingReportStatus.Complete){
                    LOG.warn("The task estimatedPercentageCompleted but the status isn't Complete");
                }_
`

For the <0% or >100% Exception, I added it to the sql script, but I thought it would be good to catch it asap so added it to **reportJobTaskProgress()** too

worker-jobtracking/src/main/java/com/hpe/caf/worker/jobtracking/JobTrackingWorkerReporter.java
xbreizh: done

worker-jobtracking/src/main/java/com/hpe/caf/worker/jobtracking/JobTrackingWorkerReporter.java
xbreizh: True, done

job-service-db/src/main/resources/procedures/reportProgress.sql
xbreizh: It was a procedure, which returns void by default, I changed back to function returning void explicitly

worker-jobtracking/src/main/java/com/hpe/caf/worker/jobtracking/JobTrackingWorkerReporter.java
xbreizh: changed to FAILED_TO_REPORT_PROGRESS

dermot-hardy (commented):
I think we could do with some tests being added for this.

dermot-hardy (approved):

xbreizh:
> I think we could do with some tests being added for this.

Should we merge now or delay until we can add some tests?

dermot-hardy:
Ideally wait until the tests are ready so it all goes in together.  But it's good to know that this bit could go in if we come under pressure for it.  Thanks, Dermot.

xbreizh:
> Ideally wait until the tests are ready so it all goes in together. But it's good to know that this bit could go in if we come under pressure for it. Thanks, Dermot.

Modified the existing testTrackingReportTask() to make sure it uses reportJobTasksComplete()
For the Integration testing, it seems to be covered by testTrackingReportTasks() within JobTrackingWorkerIT

dermot-hardy:
> Modified the existing testTrackingReportTask() to make sure it uses reportJobTasksComplete()

But that's got nothing to do with this Jira.  The test that I was expecting to be added as part of this PR is a test for the SCMOD-11919 changes to support progress reporting messages.

worker-jobtracking/src/test/java/com/hpe/caf/worker/jobtracking/JobTrackingWorkerFactoryTest.java
```diff
@@ -88,10 +90,32 @@ public void testProxiedCompleteTask() throws Exception {
     @Test
     public void testTrackingReportTask() throws Exception
     {
+        final WorkerTask wtd = Mockito.mock(WorkerTask.class);
```
dermot-hardy: So far as I can see the changes to this test are not related to SCMOD-11919.  Move them to a separate PR where they can be considered on their own merits.  If there is an issue with them we don't want that holding up the SCMOD-11919 changes.

https://github.com/salemove/github-review-helper/blob/master/doc/rules.md

xbreizh: removed and stored on a separate branch

worker-jobtracking/src/test/java/com/hpe/caf/worker/jobtracking/JobTrackingWorkerFactoryTest.java
```diff
@@ -30,14 +30,15 @@
...
+
+import static junit.framework.TestCase.assertEquals;
 import static org.mockito.Mockito.verify;
 import org.mockito.runners.MockitoJUnitRunner;
 
 import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.Map;
-import static org.junit.Assert.assertEquals;
```
dermot-hardy: Why are we changing from `org.junit.Assert.assertEquals` to `junit.framework.TestCase.assertEquals` in this module?

xbreizh: sorry -> removed

worker-jobtracking-container/src/test/java/com/hpe/caf/worker/jobtracking/JobTrackingWorkerIT.java
```diff
@@ -105,6 +112,67 @@ public void testTrackingReportTasks() throws Exception
...
+        final String jobTaskId = jobDatabase.createJobId();
+        jobDatabase.createJobTask(defaultPartitionId, jobTaskId, "");
+
+        // Validate Waiting status
+        DBJob job = jobDatabase.getJob(defaultPartitionId, jobTaskId);
+        assertEquals(JobStatus.Waiting, job.getStatus());
+
+        // Report the final task fully completed
+        final String finalTaskId = jobTaskId+".2*";
+        jobDatabase.reportTaskProgress(defaultPartitionId, finalTaskId, 100);
```
dermot-hardy: These are the worker integration tests but they seem to be bypassing the worker and updating the database directly.  It seems to defeat the purpose.  I would think the new tests should be more like the pre-existing `testTrackingReportTasks()` test (i.e. they send a message to the worker and then only contact the database to verify that the worker has done what is expected).

dermot-hardy (changes requested):

worker-jobtracking/src/test/java/com/hpe/caf/worker/jobtracking/JobTrackingWorkerFactoryTest.java
```diff
@@ -26,18 +26,17 @@
...
-import static org.mockito.Mockito.verify;
 import org.mockito.runners.MockitoJUnitRunner;
-
-import java.util.Collections;
-import java.util.Date;
-import java.util.HashMap;
-import java.util.Map;
 import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.verify;
+import java.util.Date;
```
dermot-hardy: I know you've just re-organised the imports here and the change it harmless, but it still takes time for me to look at it and realize that that is what's happened.  This is not necessary as part of the SCMOD-11919 changes.  There is no issue reorganizing the imports in another PR that is for that purpose and that would do it across the whole project if it is annoying to you.

Please take a look at these articles to help to understand where I am coming from:
https://github.com/salemove/github-review-helper/blob/master/doc/rules.md
https://mtlynch.io/code-review-love/

I do hate to keep banging on about the same thing.

xbreizh (commented):
Here I change the "percentageComplete" in JobReportingExpectation"from float to int in order to match "int estimatedPercentageCompleted" from TrackingReport

worker-jobtracking-container/src/test/java/com/hpe/caf/worker/jobtracking/JobTrackingWorkerIT.java
```diff
@@ -98,15 +101,60 @@ public void testTrackingReportTasks() throws Exception
         jobDatabase.createJobTask(defaultPartitionId, jobTaskId, "testProxiedActiveMessage");
         for (int i = 0; i <= 4; i++) {
             final JobStatus status = i < 4 ? JobStatus.Waiting : JobStatus.Completed;
+            final int percentageCompleted = i < 4 ?  0 : 100;
```
dermot-hardy: I'd suggest using an `if` statement to avoid repeating the `i < 4` and help clarify the link between the two variables:

```suggestion
            final JobStatus status;
            final int percentageCompleted;
            if (i < 4) {
                status = JobStatus.Waiting;
                percentageCompleted = 0;
            } else {
                status = JobStatus.Completed;
                percentageCompleted = 100;
            }
```

Note that `status` and `percentageCompleted` are still declared `final` when done like this, so that they can only be initialized once and cannot be used until they have been initialized.

worker-jobtracking-container/src/test/java/com/hpe/caf/worker/jobtracking/JobTrackingWorkerIT.java
```diff
@@ -98,15 +101,60 @@ public void testTrackingReportTasks() throws Exception
...
         }
     }
 
-    private JobReportingExpectation getExpectation(final String jobTaskId, final JobStatus status){
-        final int percentageCompleted = status.equals(JobStatus.Completed) ? 100 : 0;
+    @Test
+    public void testTrackingReportProgressTasks() throws Exception
+    {
+        final String jobTaskId = jobDatabase.createJobId();
+        jobDatabase.createJobTask(defaultPartitionId, jobTaskId, "testProxiedActiveMessage");
```
dermot-hardy: Should this be:
```suggestion
        jobDatabase.createJobTask(defaultPartitionId, jobTaskId, "testTrackingReportProgressTasks");
```

worker-jobtracking-container/src/test/java/com/hpe/caf/worker/jobtracking/JobTrackingWorkerIT.java
```diff
@@ -98,15 +101,60 @@ public void testTrackingReportTasks() throws Exception
...
+                ()->jobDatabase.reportTaskProgress(defaultPartitionId, null, 18));
+
+        final String expectedMessage = "Task identifier has not been specified";
+
+        Assert.assertTrue(exception1.getMessage().contains(expectedMessage));
+        Assert.assertTrue(exception2.getMessage().contains(expectedMessage));
+    }
+
+    @Test
+    @Description("Throws an exception if percentage_complete is invalid")
```
dermot-hardy: Is this preferred over using JavaDoc in some situations?  Does it display the description when the test fails or something like that?  I don't think I've seen it before.  Where did you come across it?

dermot-hardy (approved):

worker-jobtracking-container/src/test/java/com/hpe/caf/worker/jobtracking/JobTrackingWorkerIT.java
xbreizh: Sorry, just habit I had, using @Description or @DisplayName (org.junit.jupiter.api) to add a quickly readable description of what's being done. It makes it easier in the test view on the IDE. But as the convention on that project is javadoc, I ll replace with it. Also, I tested and surefire doesn't pick any DisplayName, javadoc or Description. Opening the javadoc, we can see @DisplayName and @Description but a little less clean than classic javadoc, especially if you don't know they exist.

dermot-hardy (approved):

PR #204, opened 2020-12-15
SCMOD-10792: Don't delete dependent jobs until they're published

xbreizh:
https://portal.digitalsafe.net/browse/SCMOD-10792

### DIFF
--- a/job-service-db/src/main/resources/procedures/changelog.xml
+++ b/job-service-db/src/main/resources/procedures/changelog.xml
@@ -64,7 +64,7 @@
         </createProcedure>
     </changeSet>
     <changeSet id="create_delete_dependent_jobs_proc" runOnChange="true" author="Gamma Team">
-        <createProcedure path="procedures/deleteDependentJobs.sql"
+        <createProcedure path="procedures/deleteDependentJob.sql"
                          procedureName="delete_dependent_job"
                          schemaName="public">
         </createProcedure>
--- a/job-service-db/src/main/resources/procedures/changelog.xml
+++ b/job-service-db/src/main/resources/procedures/changelog.xml
@@ -63,6 +63,12 @@
                          schemaName="public">
         </createProcedure>
     </changeSet>
+    <changeSet id="create_delete_dependent_job_proc" runOnChange="true" author="Gamma Team">
+        <createProcedure path="procedures/deleteDependentJob.sql"
+                         procedureName="delete_dependent_job"
+                         schemaName="public">
+        </createProcedure>
+    </changeSet>
     <changeSet id="create_delete_task_table_proc" runOnChange="true" author="Gamma Team">
         <createProcedure path="procedures/deleteTaskTable.sql"
                          procedureName="internal_delete_task_table"
--- a/job-service-db/src/main/resources/procedures/changelog.xml
+++ b/job-service-db/src/main/resources/procedures/changelog.xml
@@ -63,8 +63,8 @@
                          schemaName="public">
         </createProcedure>
     </changeSet>
-    <changeSet id="create_delete_dependent_job_proc" runOnChange="true" author="Gamma Team">
-        <createProcedure path="procedures/deleteDependentJob.sql"
+    <changeSet id="create_delete_dependent_jobs_proc" runOnChange="true" author="Gamma Team">
+        <createProcedure path="procedures/deleteDependentJobs.sql"
                          procedureName="delete_dependent_job"
                          schemaName="public">
         </createProcedure>
--- a/job-service-db/src/main/resources/procedures/changelog.xml
+++ b/job-service-db/src/main/resources/procedures/changelog.xml
@@ -63,7 +63,7 @@
                          schemaName="public">
         </createProcedure>
     </changeSet>
-    <changeSet id="create_delete_dependent_jobs_proc" runOnChange="true" author="Gamma Team">
+    <changeSet id="create_delete_dependent_job_proc" runOnChange="true" author="Gamma Team">
         <createProcedure path="procedures/deleteDependentJob.sql"
                          procedureName="delete_dependent_job"
                          schemaName="public">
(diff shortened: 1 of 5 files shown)


job-service-db/src/main/resources/procedures/getDependentJobs.sql
```diff
@@ -58,10 +58,5 @@ BEGIN
     SELECT *
     FROM tmp_dependent_jobs;
```
dermot-hardy: There is really no need for the temporary table now at all, since with this change it is only used once.  You can simplify the code to get rid of it.

Also update the volatility constraint to mark the function as `STABLE`.

job-service-db/src/main/resources/procedures/deleteDependentJobs.sql
```diff
@@ -0,0 +1,35 @@
...
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+--
+
+/*
+ *  Name: delete_dependent_jobs
```
dermot-hardy: Also rename the file in line with this change.

```suggestion
 *  Name: delete_dependent_job
```

job-service-db/src/main/resources/procedures/deleteDependentJobs.sql
```diff
@@ -0,0 +1,35 @@
...
+ *
+ *  Description:
+ *  Deletes dependent jobs related to a specific job
+ */
+CREATE OR REPLACE FUNCTION delete_dependent_job(
+    partition_id varchar(40),
+    job_id varchar(48))
```
dermot-hardy: I don't really like the `in_` prefix convention either but since it's used everywhere else just stick with it for now.

```suggestion
    in_partition_id VARCHAR(40),
    in_job_id VARCHAR(48))
```

job-service-db/src/main/resources/procedures/deleteDependentJobs.sql
```diff
@@ -0,0 +1,35 @@
...
+RETURNS void
+LANGUAGE 'plpgsql'
```
dermot-hardy: Specify the volatility constraint.

```suggestion
LANGUAGE plpgsql VOLATILE
```

job-service-db/src/main/resources/procedures/deleteDependentJobs.sql
```diff
@@ -0,0 +1,35 @@
...
+AS $$
+
+BEGIN
+	DELETE FROM job_task_data jtd
+	WHERE delete_dependent_job.partition_id = jtd.partition_id
+	AND delete_dependent_job.job_id = jtd.job_id;
```
dermot-hardy: Use spaces instead of tabs for the indentation (as that seems to be the convention in this project).

job-service-db/src/main/resources/procedures/changelog.xml
```diff
@@ -63,6 +63,12 @@
                          schemaName="public">
         </createProcedure>
     </changeSet>
+    <changeSet id="create_delete_dependent_jobs_proc" runOnChange="true" author="Gamma Team">
+        <createProcedure path="procedures/deleteDependentJobs.sql"
```
dermot-hardy:
```suggestion
        <createProcedure path="procedures/deleteDependentJob.sql"
```

job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/DatabasePoller.java
```diff
@@ -84,7 +84,11 @@ public static void pollDatabaseForJobsToRun() {
                     try {
                         queueServices = QueueServicesFactory.create(jtd.getTaskPipe(),codec);
                         LOG.debug(MessageFormat.format("Sending task data to the target queue {0} ...", workerAction.toString()));
+                        if (jtd.getJobId().contains("xla")) {
+                            throw new RuntimeException("Rory testing SCMOD-10792");
+                        }
```
dermot-hardy, line 89: What on Earth is this in our actual code? 🤯

dermot-hardy (changes requested):

job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/DatabasePoller.java
xbreizh: just for testing, to be removed on merge of course

job-service-db/src/main/resources/procedures/getDependentJobs.sql
xbreizh: done

job-service-db/src/main/resources/procedures/deleteDependentJobs.sql
xbreizh: replaced with 4 spaces

dermot-hardy (approved):

job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/DatabasePoller.java
dermot-hardy: Hi @xbreizh, I can see that you forgot to remove this and attempted to fix it with #205.  In case you're not aware your #205 commit has not removed it (see 99f546c - there are no changes).  This is because you reused the same branch for #205 instead of creating a new one after the merge of #204.  You should always delete your local branch after it is merged and not re-use it.  Anyway I just wanted to let you know in case you missed it that the #205 commit was ineffectual.

PR #208, opened 2020-12-21
SCMOD-11850: Optimized tests to use conditional wait statement

pintusingh135:
Replaced the Thread.sleep methods in caf-job-service automation test cases with a conditional wait method for a better execution time.

### DIFF
--- a/job-service-acceptance-tests/src/test/java/com/hpe/caf/jobservice/acceptance/JobServiceEndToEndIT.java
+++ b/job-service-acceptance-tests/src/test/java/com/hpe/caf/jobservice/acceptance/JobServiceEndToEndIT.java
@@ -1351,14 +1351,15 @@ public class JobServiceEndToEndIT {
      * @throws InterruptedException
      */
     private void waitUntilJobCompletes(String jobId) throws ApiException, InterruptedException {
-        long currentTime= System.currentTimeMillis();
-        long endTime = currentTime+5000;
-        while(System.currentTimeMillis() < endTime) {
-            Job job=jobsApi.getJob(defaultPartitionId, jobId, jobCorrelationId);
-            String currentJobStatus= job.getStatus();
-            LOG.info("Current "+jobId+" status: "+currentJobStatus);
-            if(currentJobStatus.equalsIgnoreCase("Completed")){
+        for (int counter = 0; counter < 60; counter++) {
+            Job job = jobsApi.getJob(defaultPartitionId, jobId, jobCorrelationId);
+            String currentJobStatus = job.getStatus();
+            LOG.info("Job " + jobId + " current status: " + currentJobStatus);
+            if (currentJobStatus.equalsIgnoreCase("Completed")) {
                 break;
+            }
+            else {
+                Thread.sleep(500);
             }
         }
     }
--- a/job-service-acceptance-tests/src/test/java/com/hpe/caf/jobservice/acceptance/JobServiceEndToEndIT.java
+++ b/job-service-acceptance-tests/src/test/java/com/hpe/caf/jobservice/acceptance/JobServiceEndToEndIT.java
@@ -142,7 +142,7 @@ public class JobServiceEndToEndIT {
                 new JobServiceEndToEndITExpectation(
                         false,
                         exampleWorkerMessageOutQueue,
-                        defaultPartitionId,
+                    defaultPartitionId,
                         jobId,
                         jobCorrelationId,
                         ExampleWorkerConstants.WORKER_NAME,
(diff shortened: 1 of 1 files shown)


job-service-acceptance-tests/src/test/java/com/hpe/caf/jobservice/acceptance/JobServiceEndToEndIT.java
```diff
@@ -1329,4 +1341,19 @@ public void run() {
...
         return timer;
     }
+
+    /**
+     * Desc: Waits till the job is Completed
+     * @param jobId
+     * @throws ApiException
+     * @throws InterruptedException
+     */
+    private void waitUntilJobCompletes(String jobId) throws ApiException, InterruptedException {
```
michael-bryson, line 1353: What happens if there is an issue with the job and it never completes, will this get stuck in the loop? Why not use a for loop to ensure you exit out after a certain number of tries.
```suggestion
    private void waitUntilJobCompletes(String jobId) throws ApiException, InterruptedException {
        for (int i = 0; i < 60; i++) {
            Job job = jobsApi.getJob(defaultPartitionId, jobId, jobCorrelationId);
            String currentJobStatus = job.getStatus();
            LOG.info("Job " + jobId + " current status: " + currentJobStatus);
            if (currentJobStatus.equalsIgnoreCase("completed")) {
                break;
            }
            else {
                Thread.sleep(500);
            }
        }
    }
```

job-service-acceptance-tests/src/test/java/com/hpe/caf/jobservice/acceptance/JobServiceEndToEndIT.java
```diff
@@ -240,13 +240,15 @@ public void testTargetPipeForJobWithNoAndWithCompletedPrerequisiteJobs(final Str
             createJob(job2Id, true);
         }
 
-        Thread.sleep(1000); // Add short delay to allow previous jobs to complete
+        //Wait for the job to complete
```
michael-bryson: Should `waitUntilJobCompletes(job1Id);` be in this test also?

job-service-acceptance-tests/src/test/java/com/hpe/caf/jobservice/acceptance/JobServiceEndToEndIT.java
```diff
@@ -350,13 +352,15 @@ public void testJobWithPrerequisiteJobsWhichHaveCompleted() throws Exception
...
```
michael-bryson, line 356: Should `waitUntilJobCompletes(job1Id);` be in this test also?

michael-bryson (changes requested):

job-service-acceptance-tests/src/test/java/com/hpe/caf/jobservice/acceptance/JobServiceEndToEndIT.java
```diff
@@ -1329,4 +1343,23 @@ public void run() {
...
+
+    /**
+     * Desc: Waits till the job is Completed
+     * @param jobId
+     * @throws ApiException
+     * @throws InterruptedException
+     */
+    private void waitUntilJobCompletes(String jobId) throws ApiException, InterruptedException {
+        long currentTime= System.currentTimeMillis();
+        long endTime = currentTime+5000;
```
michael-bryson: Why are you doing this and not using a `for loop` as I suggested? Any tests which have a delay set for the prerequisite jobs of over 5 seconds will fail using this method.

michael-bryson (changes requested):
Make sure you put a proper title on this pull request.

job-service-acceptance-tests/src/test/java/com/hpe/caf/jobservice/acceptance/JobServiceEndToEndIT.java
pintusingh135: Added waitUntilJobCompletes() method for job1Id

job-service-acceptance-tests/src/test/java/com/hpe/caf/jobservice/acceptance/JobServiceEndToEndIT.java
pintusingh135: Added waitUntilJobCompletes() method for job1Id

job-service-acceptance-tests/src/test/java/com/hpe/caf/jobservice/acceptance/JobServiceEndToEndIT.java
pintusingh135: Replaced the While loop with a for loop

michael-bryson (approved):

PR #211, opened 2020-12-22
Further optimization of tests

michael-bryson:
@dermot-hardy  I found this hard coded wait in the job tracking worker tests so added this to only wait until the verifyJobStatus returns without exception. Not sure if this is ok or not but it seems to take some time off the build.


michael-bryson:
Dev build: http://sou-jenkins2.swinfra.net/job/JobService/view/Developer/job/JobService~job-service~mb-test~CI/

PR #212, opened 2021-01-04
SCMOD-12164: Try to get more info from add to queue failure

dgibson-microfocus:
https://portal.digitalsafe.net/browse/SCMOD-12164

### DIFF
--- a/job-service/src/main/java/com/hpe/caf/services/job/api/JobsPut.java
+++ b/job-service/src/main/java/com/hpe/caf/services/job/api/JobsPut.java
@@ -207,8 +207,8 @@ public final class JobsPut {
             //  Get database helper instance.
             try {
                 QueueServices queueServices = QueueServicesFactory.create(config, jobTask.getTaskPipe(),codec);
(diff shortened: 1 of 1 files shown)


job-service/src/main/java/com/hpe/caf/services/job/api/JobsPut.java
```diff
@@ -205,10 +207,10 @@ public static String createOrUpdateJob(final String partitionId, String jobId, N
             //  Get database helper instance.
             try {
                 QueueServices queueServices = QueueServicesFactory.create(config, jobTask.getTaskPipe(),codec);
-                LOG.debug("createOrUpdateJob: Sending task data to the target queue...");
```
dermot-hardy, line 208: Is moving this debug line down intentional?

dermot-hardy (approved):

job-service/src/main/java/com/hpe/caf/services/job/api/JobsPut.java
dgibson-microfocus: Oh, no it wasn't. I'll fix it now.

PR #213, opened 2021-01-04
SCMOD-12164: Add message publish confirm and retry

dgibson-microfocus:
https://portal.digitalsafe.net/browse/SCMOD-12164

### DIFF
--- a/job-service/src/main/java/com/hpe/caf/services/job/queue/QueueServices.java
+++ b/job-service/src/main/java/com/hpe/caf/services/job/queue/QueueServices.java
@@ -27,6 +27,8 @@ import com.rabbitmq.client.Channel;
 import com.rabbitmq.client.Connection;
 import com.rabbitmq.client.MessageProperties;
 import org.apache.commons.codec.binary.Base64;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import javax.ws.rs.core.UriBuilder;
 import java.io.IOException;
@@ -43,6 +45,7 @@ import java.util.concurrent.TimeoutException;
  */
 public final class QueueServices {
 
+    private static final Logger LOG = LoggerFactory.getLogger(QueueServices.class);
     private final Connection connection;
     private final Channel publisherChannel;
     private final String targetQueue;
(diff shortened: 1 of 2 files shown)


job-service/src/main/java/com/hpe/caf/services/job/queue/QueueServices.java
```diff
@@ -124,8 +128,20 @@ public void sendMessage(
...
+            LOG.warn("Failed to publish message. Will retry", e);
+            publishMessage(taskMessageBytes);
+        }
+    }
+
+    private void publishMessage(final byte[] taskMessageBytes)
+            throws IOException, InterruptedException, TimeoutException
+    {
         publisherChannel.basicPublish(
                 "", targetQueue, MessageProperties.TEXT_PLAIN, taskMessageBytes);
```
dermot-hardy, line 143: Should this also be `MessageProperties.PERSISTENT_TEXT_PLAIN`?

dermot-hardy (approved):

job-service/src/main/java/com/hpe/caf/services/job/queue/QueueServices.java
dgibson-microfocus: Yes probably, assuming this wouldn't cause disk issues. (Also worth noting none of the workers send with `PERSISTENT_TEXT_PLAIN`)

dermot-hardy: Really, what are the implications?  Do we need to raise a Jira for it?

dgibson-microfocus: So the persistent flag means messages will be written to disk immediately on arrival to the queue regardless of current memory levels. This means on durable queues any persistent messages will be recovered if the node restarts as opposed to transient messages which are only ever stored in memory (unless near overflow) and are lost if the node goes down.

This is all separate from mirrored queues (which we have in prod) and only really applies to single node recovery. It only comes into play if multiple nodes die before the messages can be replicated. (WHich seems very unlikely)

dermot-hardy: Ok.  I assume the acknowledgement is never received until the message has been secured.  If so then it sounds like it's not an issue.

PR #215, opened 2021-01-12
SCMOD-11998: Release job-service 3.5.0

xbreizh:
https://portal.digitalsafe.net/browse/SCMOD-11998


xbreizh:
I didn't add **[SCMOD-11843](https://portal.digitalsafe.net/browse/SCMOD-11843)**: Fix deadlock as it's a fix for a bug we introduced in that same release

dermot-hardy (approved):

PR #218, opened 2021-01-15
SCMOD-12309: Add replace function on update_job_progress

xbreizh:
https://portal.digitalsafe.net/browse/SCMOD-12309


dgibson-microfocus (approved):
I assume you've tested the upgrade?

michael-bryson (approved):

PR #219, opened 2021-01-15
SCMOD-11252: Replace use of ProcessEnvironment class

rorytorneymf:
https://portal.digitalsafe.net/browse/SCMOD-11252

Using the `ProcessEnvironment `class to set environment variables in unit tests does not work sometimes, due to the class returning different fields in different versions of Java:

https://stackoverflow.com/a/62855645/12177456

We can work around it using something like this:

https://github.houston.softwaregrp.net/caf/caf-storage/pull/274/files#diff-6eb71c520c088c4c17059d20c5fa3a94R50

but using a JUnit rule here seems to be a little simpler.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -165,6 +165,11 @@
                 <version>3.5.0-SNAPSHOT</version>
             </dependency>
             <dependency>
+                <groupId>com.github.stefanbirkner</groupId>
+                <artifactId>system-rules</artifactId>
+                <version>1.19.0</version>
+            </dependency>
+            <dependency>
                 <groupId>com.github.workerframework</groupId>
                 <artifactId>worker-example-shared</artifactId>
                 <version>1.8.0-249</version>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -164,6 +164,11 @@
                 <artifactId>worker-jobtracking-shared</artifactId>
                 <version>3.5.0-SNAPSHOT</version>
             </dependency>
+            <dependency>
+                <groupId>com.github.stefanbirkner</groupId>
```
rorytorneymf, line 168: @dermot-hardy I wasn't sure if this should go into the https://github.com/CAFapi/caf-dependency-management-bom/blob/master/pom.xml or not - should all new dependencies go in there?

dermot-hardy: What you've done is fine.  I think we'll need a rethink on `caf-dependency-management-bom`.

dermot-hardy (approved):

PR #222, opened 2021-01-27
SCMOD-12257: Correlation Id Support

simranjeetc:
https://portal.digitalsafe.net/browse/SCMOD-12257


dermot-hardy (approved):
Changes need to be made for package names, artifact coordinates, etc., but nonetheless I'm approving these changes in principle.

simranjeetc:
@dermot-hardy Added worker-api version under dependencyManagement in commit ca8d189 

The build fails if this is not added explicitly as I observed that 4.0.0-XYZ version of worker-api was getting used instead of 4.1.0-SNAPSHOT. Mentioning it here so that you can let me know if this is not right way to handle this.

dermot-hardy (approved):

PR #223, opened 2021-02-10
SCMOD-12505: Update job types to support taskScript

rorytorneymf:
Jira: https://portal.digitalsafe.net/browse/SCMOD-12505

Previously, when creating a new job type, users would supply a `taskDataScript` property.

Now, users should supply a `taskScript` property, which should represent the entire task, not just the `taskData` portion of the task.


job-service-container/src/test/resources/job-type-definitions/basic.yaml
```diff
@@ -14,14 +14,17 @@
...
+      "taskData": {
+        "config": .configuration,
+        "taskQueue": .taskPipe,
+        "targetQueue": .targetPipe,
+        "partitionIdent": .partitionId,
+        "jobIdent": .jobId,
+        "reqParams": .parameters
+      },
+      "taskPipe": .taskPipe,
+      "targetPipe": .targetPipe
```
dermot-hardy: There can't be a `.taskPipe` and a `.targetPipe`.  The trigger for doing this work is that we need to be able to choose a target pipe based on one of the parameters so by definition there could be multiple potential target pipes.  These need to come from the configuration object.

job-service-container/src/test/resources/job-type-definitions/basic.yaml
```diff
@@ -14,14 +14,17 @@
...
-  {
-    "config": .configuration,
-    "taskQueue": .taskPipe,
-    "targetQueue": .targetPipe,
-    "partitionIdent": .partitionId,
-    "jobIdent": .jobId,
-    "reqParams": .parameters,
-  }
+taskScript: |
+    { "taskClassifier": "basic classifier",
```
dermot-hardy: Adjust the formatting here:
```suggestion
  {
    "taskClassifier": "basic classifier",
```

job-service-container/src/test/resources/job-type-definitions/invalid-output.yaml
```diff
@@ -14,7 +14,10 @@
...
 
-taskClassifier: classifier
-taskApiVersion: 1
-taskDataScript: |
-  ["this is bad", "should be object"]
+#taskClassifier: classifier
+#taskApiVersion: 1
+#taskDataScript: |
+#  ["this is bad", "should be object"]
+#
```
dermot-hardy: Delete this comment

job-service-container/src/test/resources/job-type-definitions/wrong-ext.notyaml
```diff
@@ -14,6 +14,5 @@
 # limitations under the License.
 #
 
-taskClassifier: classifier
-taskApiVersion: 1
-taskDataScript: {}
+taskScript: |
+    {}
```
dermot-hardy: Should this be `taskScript: {}`?  I'm guessing it's just irrelevant?

job-service/src/test/resources/job-type-definitions/config.yaml
```diff
@@ -14,11 +14,14 @@
...
-taskApiVersion: 1
 configurationProperties:
   - name: prop_a
     description: the first property
   - name: prop_b
-taskDataScript: |
-  { "cfg": .configuration }
+taskScript: |
+  { "taskClassifier": "basic classifier",
+    "taskApiVersion": 74,
```
dermot-hardy: Not sure if it's important but this was `classifier / 1` before.

job-service/src/test/resources/job-type-definitions/invalid-syntax.yaml
```diff
@@ -14,6 +14,10 @@
...
-taskClassifier: { this: is invalid
-taskApiVersion: 1
-taskDataScript: .
+taskScript: |
+    { "taskClassifier": { this: is invalid,
+      "taskApiVersion": 74,
+      "taskData": {"cfg": .configuration},
+      "taskPipe": "basic task pipe",
+      "targetPipe": "basic target pipe"
+    }
```
dermot-hardy, line 26: I don't think this is still testing the same thing.  Before the yaml as a whole was invalid but now the yaml is valid but the script in invalid.  It's probably still a good test but it is different.

job-service/src/test/resources/job-type-definitions/missing-config-name.yaml
```diff
@@ -14,10 +14,16 @@
...
-taskDataScript: .
+taskScript: |
+  { "taskClassifier": "basic classifier",
+    "taskApiVersion": 74,
+    "taskData": {"cfg": .configuration},
+    "taskPipe": "basic task pipe",
+    "targetPipe": "basic target pipe"
+  }
+
+
```
dermot-hardy: Drop the extra newlines

job-service/src/test/resources/job-type-definitions/schema.yaml
```diff
@@ -14,8 +14,13 @@
...
 
-taskClassifier: classifier
-taskApiVersion: 1
 jobParametersSchema:
   type: string
-taskDataScript: .
+taskScript: |
+  { "taskClassifier": "basic classifier",
+    "taskApiVersion": 74,
+    "taskData": {"cfg": .configuration},
```
dermot-hardy: The `taskData` is completely different.  I can't tell from the name of the yaml file what this is trying to test so I don't know if it's important or not.

job-service/src/test/resources/job-type-definitions/empty-targetpipe.yaml
```diff
@@ -14,6 +14,10 @@
...
-taskClassifier: classifier
-taskApiVersion: not a number
-taskDataScript: .
+taskScript: |
+    { "taskClassifier": "basic classifier",
+      "taskApiVersion": 74,
+      "taskData": {"cfg": .configuration},
+      "taskPipe": "basic task pipe",
+      "targetPipe": ""
+    }
```
dermot-hardy: I'm not sure what the test is here but note that empty target pipe is valid (in fact it is the most common value for `targetPipe` to have).

job-service/src/main/java/com/hpe/caf/services/job/jobtype/JsltTaskBuilder.java
```diff
@@ -56,16 +56,16 @@
...
      *                            before passing them to the script
-     * @param taskDataScript the uncompiled JSLT script
+     * @param taskScript the uncompiled JSLT script
      * @throws InvalidJobTypeDefinitionException
      */
-    public JsltTaskDataBuilder(
+    public JsltTaskBuilder(
         final String jobTypeId,
         final String taskPipe,
         final String targetPipe,
```
dermot-hardy: I would have thought that these should no longer be being passed in.

job-service/src/main/java/com/hpe/caf/services/job/jobtype/JobType.java
```diff
@@ -81,21 +65,68 @@ public String getId() {
...
+    ) throws BadRequestException, InvalidJobTypeDefinitionException
+    {
+        final WorkerAction workerAction = new WorkerAction();
+        final JsonNode task = taskBuilder.build(partitionId, jobId, parameters);
         // job-put expects `Map` (or `String`), but only does a shallow type check
         try {
-            task.setTaskData(objectMapper.convertValue(taskData, Map.class));
+            // taskClassifier
+            final JsonNode taskClassifier = getNonNullPropertyFromTask(task, "taskClassifier");
+            if (!taskClassifier.isTextual() || taskClassifier.asText().isEmpty()) {
```
dermot-hardy: Rather than doing all this manual validation, define a JSON Schema for the expected output format and include it as a resource.  Then at runtime validate the `task` json node against it.  This is much the same way that the parameters passed by the user are validated against the schema in the job type.

job-service/src/main/java/com/hpe/caf/services/job/jobtype/DefaultDefinitionParser.java
```diff
@@ -100,35 +98,9 @@ public JobType parse(final String id, final InputStream definitionStream)
...
-        public Integer getTaskApiVersion(final String id) throws InvalidJobTypeDefinitionException {
-            if (taskApiVersion == null) {
-                throw new InvalidJobTypeDefinitionException(
-                    id + ": missing property: taskApiVersion");
-            }
-            return taskApiVersion;
-        }
+        private String taskScript;
 
         public String getTaskPipe(final String id, final AppConfig appConfig)
```
dermot-hardy: I see the others have been removed so not sure why this continues to exist.

job-service/src/main/java/com/hpe/caf/services/job/jobtype/DefaultDefinitionParser.java
```diff
@@ -66,13 +66,11 @@ public JobType parse(final String id, final InputStream definitionStream)
         final String targetPipe = definition.getTargetPipe(id, appConfig);
         final ParametersValidator parametersValidator =
             new JsonSchemaParametersValidator(id, definition.getParametersSchema());
-        final TaskDataBuilder taskDataBuilder = new JsltTaskDataBuilder(
+        final TaskBuilder taskBuilder = new JsltTaskBuilder(
             id, taskPipe, targetPipe, definition.getConfiguration(id, appConfig),
```
dermot-hardy: I don't get `taskPipe` and `targetPipe` continuing to be passed in here.

dermot-hardy (changes requested):

job-service/src/test/resources/job-type-definitions/schema.yaml
rorytorneymf: I've renamed the yaml and test to make it clearer what's being tested here, which is the job params

job-service-container/src/test/resources/job-type-definitions/basic.yaml
rorytorneymf: So is the idea the `*TASK_PIPE*` and `*TARGET_PIPE*` env vars are going to be removed? And you would have to specify those instead in the parameters json when creating the job? E.g.

```
"parameters": {
    "batches": ["batch-a", "batch-b"],
    "operationMode": "EXTRACT",
     ...
    "taskPipe": ...
    "targetPipe":...
  }
```

I'm just confused about how the "configuration object" you mention above comes into it - is that the same as the 'configurationProperties' in the yaml?

```
configurationProperties:
- name: WORKFLOW_WORKER_INPUT_QUEUE
  description: Input Queue for the Workflow Worker
```

Thanks

dermot-hardy: > So is the idea the `*TASK_PIPE*` and `*TARGET_PIPE*` env vars are going to be removed?

I know you're leaving the configuration changes for a later PR so they don't have to be removed now, but if you're leaving them then to access them for now they will have to be declared in the `configurationProperties` section and then referenced via the configuration object (i.e. using ` .configuration.TASK_PIPE`, and ` .configuration.TARGET_PIPE`).

job-service/src/main/java/com/hpe/caf/services/job/jobtype/JobType.java
```diff
@@ -81,21 +64,15 @@ public String getId() {
...
-        task.setTaskClassifier(taskClassifier);
-        task.setTaskApiVersion(taskApiVersion);
-        task.setTaskPipe(taskPipe);
-        task.setTargetPipe(targetPipe);
-        final JsonNode taskData = taskDataBuilder.build(partitionId, jobId, parameters);
-        // job-put expects `Map` (or `String`), but only does a shallow type check
+    ) throws BadRequestException, InvalidJobTypeDefinitionException
+    {
+        final JsonNode task = taskBuilder.build(partitionId, jobId, parameters);
+        JsonSchemaTaskScriptValidator.getInstance().validate(task);
```
rorytorneymf, line 70: Wasn't sure about this static `JsonSchemaTaskScriptValidator` class here, but couldn't really find a nicer way of doing this, due to the way these classes are initialized on startup.

job-service/src/test/resources/job-type-definitions/config.yaml
```diff
@@ -14,11 +14,13 @@
 # limitations under the License.
 #
 
-taskClassifier: classifier
-taskApiVersion: 1
 configurationProperties:
-  - name: prop_a
-    description: the first property
-  - name: prop_b
```
dermot-hardy, line 22: I'm not sure what this is testing but is it OK that `prop_a` and `prop_b` have been removed.

job-service/src/test/java/com/hpe/caf/services/job/jobtype/JsltTaskBuilderTest.java
```diff
@@ -26,51 +26,40 @@
...
+        config.put("cfg key", "cfg val");
+        config.put("TASK_PIPE", "task pipe");
+        config.put("TARGET_PIPE", "target pipe");
+        final TaskBuilder builder = new JsltTaskBuilder(
+            "type", config, paramValidatorSuccess, ".");
+        final JsonNode actualTask =
             builder.build("partition id", "job id", TextNode.valueOf("params"));
 
-        Assert.assertEquals("targetPipe should not be passed to script",
-            NullNode.getInstance(), actualTaskData);
```
dermot-hardy, line 59: import for `NullNode` no longer required.

job-service/src/main/java/com/hpe/caf/services/job/jobtype/DefaultDefinitionParser.java
```diff
@@ -175,7 +122,11 @@ public void setConfigurationProperties(
                 }
                 // preserve property name case in the output, even though the configuration lookup
                 // ignores it
-                configuration.put(propertyName, propertyValue);
+                if (propertyName.equals("TARGET_PIPE") && propertyValue.equals("")) {
+                    configuration.put(propertyName, null);
```
dermot-hardy, line 126: Is this the right place for this?

job-service/src/main/resources/jobtype/taskscript-schema.yaml
```diff
@@ -0,0 +1,37 @@
...
+  taskApiVersion:
+    type: integer
+    minimum: 1
+  taskData:
+    type: object
+  taskPipe:
+    type: string
+    minLength: 1
+  targetPipe:
+    type: ["string", "null"]
```
dermot-hardy: Should this also have a `minLength: 1` if empty strings are not allowed?

job-service/src/main/resources/jobtype/taskscript-schema.yaml
```diff
@@ -0,0 +1,37 @@
+#
```
dermot-hardy: This is in the `jobtype` package.  Use the same package name as the related source code - i.e. `com.hpe.caf.services.job.jobtype`.

dermot-hardy (approved):

job-service/src/test/resources/job-type-definitions/config.yaml
rorytorneymf: Yes its ok, since we're using the `TASK_PIPE` and `TARGET_PIPE` now in the examples and tests, we didn't need these extra configuration properties.

job-service/src/main/java/com/hpe/caf/services/job/jobtype/DefaultDefinitionParser.java
rorytorneymf: I think it's ok. The original code had this logic in this same class (although in a different method, `getTargetPipe` which has now been removed):

https://github.com/JobService/job-service/blob/develop/job-service/src/main/java/com/hpe/caf/services/job/jobtype/DefaultDefinitionParser.java#L149

```
public String getTargetPipe(final String id, final AppConfig appConfig)
            throws InvalidJobTypeDefinitionException
{
    final String targetPipe = appConfig.getJobTypeProperty(id, "target_pipe");
    if (targetPipe == null) {
        throw new InvalidJobTypeDefinitionException(id + ": target pipe is not configured");
    } else if (targetPipe.equals("")) {
        return null;
    } else {
        return targetPipe;
    }
}
```

I could move it to somewhere else such as the [JsltTaskBuilder ](https://github.com/JobService/job-service/blob/SCMOD-12505/job-service/src/main/java/com/hpe/caf/services/job/jobtype/JsltTaskBuilder.java) or [JobType](https://github.com/JobService/job-service/blob/SCMOD-12505/job-service/src/main/java/com/hpe/caf/services/job/jobtype/JobType.java) class, but at that point, we're working with a `JsonNode` rather than a string, so it's slightly more awkward (have to check if JsonNode has prop, have it check its type, cast it in order to set it again).

PR #224, opened 2021-02-16
SCMOD-12505: Make job type config more flexible

rorytorneymf:
Jira: https://portal.digitalsafe.net/browse/SCMOD-12505

The names used for the `configurationProperties` now have a direct mapping to the environment variables used to populate them. For example, if a configuration property is named `TASK_PIPE` in a job type yaml definition, then the value for that configuration property is expected to be available in an environment variable named `TASK_PIPE`. There is no longer a requirement that the env variable must start with `CAF_JOB_SERVICE_JOB_TYPE_<JOB-TYPE-ID>_...`

### DIFF
--- a/job-service/src/main/java/com/hpe/caf/services/job/jobtype/DefaultDefinitionParser.java
+++ b/job-service/src/main/java/com/hpe/caf/services/job/jobtype/DefaultDefinitionParser.java
@@ -115,18 +115,14 @@ public final class DefaultDefinitionParser implements DefinitionParser {
             final Map<String, String> configuration = new HashMap<>();
             for (final ConfigurationProperty property : properties) {
                 final String propertyName = property.getName(id, this);
-                final String propertyValue = appConfig.getJobTypeProperty(id, propertyName);
+                final String propertyValue = appConfig.getJobProperty(propertyName);
                 if (propertyValue == null) {
                     throw new InvalidJobTypeDefinitionException(
                         id + ": configuration is not available: " + propertyName);
                 }
                 // preserve property name case in the output, even though the configuration lookup
                 // ignores it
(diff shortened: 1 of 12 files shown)


job-service/src/main/java/com/hpe/caf/services/job/jobtype/DefaultDefinitionParser.java
```diff
@@ -115,18 +115,14 @@ public void setConfigurationProperties(
...
                 final String propertyName = property.getName(id, this);
-                final String propertyValue = appConfig.getJobTypeProperty(id, propertyName);
+                final String propertyValue = appConfig.getJobProperty(propertyName);
                 if (propertyValue == null) {
                     throw new InvalidJobTypeDefinitionException(
                         id + ": configuration is not available: " + propertyName);
                 }
                 // preserve property name case in the output, even though the configuration lookup
                 // ignores it
-                if (propertyName.equals("TARGET_PIPE") && propertyValue.equals("")) {
```
rorytorneymf, line 125: I've removed this as I don't think we need this functionality anymore. (allowing empty string for target pipe). Looking at Josephs PR it seemed to be a way to signify that the target pipe is not set:

https://github.com/JobService/job-service/pull/97#issuecomment-505791033

We don't need this anymore as there is no need now to provide an env var for the target pipe if you don't want it set to anything.

Also, we don't know what the target_pipe config might be called now, as it could be named anything the user wants, so this test `propertyName.equals("TARGET_PIPE")` won't really work now.

dermot-hardy: Great.

dermot-hardy (approved):

PR #225, opened 2021-02-17
SCMOD-11786: Return HTTP 503 for db connection errors

kyra-ohare:
Jira: https://portal.digitalsafe.net/browse/SCMOD-11786
Return HTTP status code 503 (instead of 500) when the service cannot get a connection to PostgresSQL.

### DIFF
--- a/job-service/src/main/java/com/hpe/caf/services/job/api/DatabaseHelper.java
+++ b/job-service/src/main/java/com/hpe/caf/services/job/api/DatabaseHelper.java
@@ -13,6 +13,21 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+/*
+* Copyright 2016-2021 Micro Focus or one of its affiliates.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*      http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
 package com.hpe.caf.services.job.api;
 
 import com.hpe.caf.services.db.client.DatabaseConnectionProvider;
@@ -24,6 +39,7 @@ import com.hpe.caf.services.job.api.generated.model.SortField;
 import com.hpe.caf.services.job.exceptions.BadRequestException;
 import com.hpe.caf.services.job.exceptions.ForbiddenException;
 import com.hpe.caf.services.job.exceptions.NotFoundException;
+import com.hpe.caf.services.job.exceptions.ServiceUnavailableException;
 import com.hpe.caf.services.job.util.JobTaskId;
 import org.codehaus.jettison.json.JSONObject;
 import org.slf4j.Logger;
@@ -42,11 +58,17 @@ import java.util.stream.Collectors;
 import org.codehaus.jettison.json.JSONException;
 
 /**
- * The DatabaseHelper class is responsible for database operations.
- */
+* The DatabaseHelper class is responsible for database operations.
+*/
 public final class DatabaseHelper
 {
     private static final String FAILURE_PROPERTY_MISSING = "Unknown";
+
+    // PostgreSQL Error Codes: https://www.postgresql.org/docs/current/errcodes-appendix.html
+    private static final String POSTGRES_CONNECTION_EXCEPTION_ERROR_CODE_PREFIX = "08";
+    private static final String POSTGRES_NO_DATA_ERROR_CODE = "02000";
+    private static final String POSTGRES_NO_DATA_FOUND_ERROR_CODE = "P0002";
+    private static final String POSTGRES_UNIQUE_VIOLATION_ERROR_CODE = "23505";
 
     private static AppConfig appConfig;
 
@@ -140,6 +162,13 @@ public final class DatabaseHelper
                     array.free();
                 }
             }
+        } catch (final SQLException se) {
+            final String sqlState = se.getSQLState();
+            if (sqlState.startsWith(POSTGRES_CONNECTION_EXCEPTION_ERROR_CODE_PREFIX)) {
+                throw new ServiceUnavailableException(se.getMessage());
+            } else {
+                throw se;
+            }
         }
 
         //  Convert arraylist to array of jobs.
@@ -176,6 +205,13 @@ public final class DatabaseHelper
                 if (rs.next()) {
                     jobsCount = rs.getLong(1);
                 }
+            }
+        } catch (final SQLException se) {
+            final String sqlState = se.getSQLState();
+            if (sqlState.startsWith(POSTGRES_CONNECTION_EXCEPTION_ERROR_CODE_PREFIX)) {
+                throw new ServiceUnavailableException(se.getMessage());
+            } else {
+                throw se;
             }
         }
 
@@ -221,19 +257,21 @@ public final class DatabaseHelper
                     }
                 }
             }
-        } catch (SQLException se) {
+        } catch (final SQLException se) {
             //  Determine source of SQL exception and throw appropriate error.
-            String sqlState = se.getSQLState();
+            final String sqlState = se.getSQLState();
 
-            switch (sqlState) {
-                case "02000":
-                    //  Job id has not been provided.
-                    throw new BadRequestException(se.getMessage());
-                case "P0002":
-                    //  No data found for the specified job id.
-                    throw new NotFoundException(se.getMessage());
-                default:
-                    throw se;
+            if (sqlState.equals(POSTGRES_NO_DATA_ERROR_CODE)) {
+                //  Job id has not been provided.
+                throw new BadRequestException(se.getMessage());
+            } else if (sqlState.equals(POSTGRES_NO_DATA_FOUND_ERROR_CODE)) {
+                //  No data found for the specified job id.
+                throw new NotFoundException(se.getMessage());
+            } else if (sqlState.startsWith(POSTGRES_CONNECTION_EXCEPTION_ERROR_CODE_PREFIX)) {
+                // Connection exception.
+                throw new ServiceUnavailableException(se.getMessage());
+            } else  {
+                throw se;
             }
         }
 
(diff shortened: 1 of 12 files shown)


rorytorneymf:
Apologies, I forgot about swagger:

https://github.com/JobService/job-service/blob/develop/job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml

Think we need to update this to document that the endpoints can return a 503 now.

rorytorneymf:
@kleyton-ohare Looks like you accidentally reverted your previous commit. Can you add these tests back in? Thanks

image

job-service/src/main/java/com/hpe/caf/services/job/exceptions/ServiceUnavailableException.java
```diff
@@ -0,0 +1,32 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hpe.caf.services.job.exceptions;
+
+/**
+ * Custom exception implemented for the job service api. Exceptions of this type map directly onto http 503 status codes.
+ */
+public class ServiceUnavailableException extends Exception
```
dermot-hardy: Can this class be `final`?
```suggestion
public final class ServiceUnavailableException extends Exception
```

job-service/src/main/java/com/hpe/caf/services/job/api/DatabaseHelper.java
```diff
@@ -42,12 +43,18 @@
 import org.codehaus.jettison.json.JSONException;
 
 /**
- * The DatabaseHelper class is responsible for database operations.
- */
+* The DatabaseHelper class is responsible for database operations.
+*/
```
dermot-hardy: Put those spaces back.

job-service/src/main/java/com/hpe/caf/services/job/api/DatabaseHelper.java
```diff
@@ -221,19 +242,21 @@ public Job getJob(final String partitionId, String jobId) throws Exception {
...
+            if (sqlState.equals(POSTGRES_NO_DATA_ERROR_CODE)) {
+                //  Job id has not been provided.
+                throw new BadRequestException(se.getMessage());
+            } else if (sqlState.equals(POSTGRES_NO_DATA_FOUND_ERROR_CODE)) {
+                //  No data found for the specified job id.
+                throw new NotFoundException(se.getMessage());
+            } else if (sqlState.startsWith(POSTGRES_CONNECTION_EXCEPTION_ERROR_CODE_PREFIX)) {
+                // Connection exception.
+                throw new ServiceUnavailableException(se.getMessage(), se);
+            } else  {
```
dermot-hardy:
```suggestion
            } else {
```

job-service/src/main/java/com/hpe/caf/services/job/api/DatabaseHelper.java
```diff
@@ -383,20 +422,22 @@ public void deleteJob(final String partitionId, String jobId) throws Exception {
...
+                throw new BadRequestException(se.getMessage());
+            } else if (sqlState.equals(POSTGRES_NO_DATA_FOUND_ERROR_CODE)) {
+                //  No data found for the specified job id.
+                throw new NotFoundException(se.getMessage());
+            } else if (sqlState.startsWith(POSTGRES_CONNECTION_EXCEPTION_ERROR_CODE_PREFIX)) {
+                // Connection exception
+                throw new ServiceUnavailableException(se.getMessage(), se);
+            } else {
+                throw se;
+            }
```
dermot-hardy:
```suggestion
            }
```

job-service/src/main/java/com/hpe/caf/services/job/api/DatabaseHelper.java
```diff
@@ -481,20 +528,22 @@ public void cancelJob(final String partitionId, String jobId) throws Exception {
...
```
dermot-hardy:
```suggestion
            }
```

job-service/src/main/java/com/hpe/caf/services/job/api/DatabaseHelper.java
```diff
@@ -221,19 +242,21 @@ public Job getJob(final String partitionId, String jobId) throws Exception {
...
+
+            if (sqlState.equals(POSTGRES_NO_DATA_ERROR_CODE)) {
+                //  Job id has not been provided.
+                throw new BadRequestException(se.getMessage());
+            } else if (sqlState.equals(POSTGRES_NO_DATA_FOUND_ERROR_CODE)) {
+                //  No data found for the specified job id.
+                throw new NotFoundException(se.getMessage());
+            } else if (sqlState.startsWith(POSTGRES_CONNECTION_EXCEPTION_ERROR_CODE_PREFIX)) {
+                // Connection exception.
+                throw new ServiceUnavailableException(se.getMessage(), se);
```
dermot-hardy: I see this pattern repeating over and over.  Would it be a good idea to have a `mapSqlException()` utility function which takes the `SQLException` and returns the `Exception` to be thrown?

dermot-hardy (approved):

job-service/src/main/java/com/hpe/caf/services/job/api/DatabaseHelper.java
rorytorneymf: @dermot-hardy @kleyton-ohare Just be aware that all the catch blocks are not the same, so changing them to all perform the same exception processing is changing the existing behaviour now.

For example, in the original code for the `canJobBeProgressed` function, if the error code is P0002 the method returns false:

```
        } catch (SQLException e) {
            final String sqlState = e.getSQLState();
            if (sqlState.equals("P0002")) {
                // job missing - return false
            } else {
                throw e;
            }
        }
        return canBeProgressed;
```

Now the method will throw a NotFoundException instead of returning false.

```
 else if (sqlState.equals(POSTGRES_NO_DATA_FOUND_ERROR_CODE)) {
            //  No data found for the specified job id.
            return new NotFoundException(se.getMessage());
        }
```

There are other examples as well - methods that may have thrown an SQLException previously may now throw one of the other exceptions, `NotFoundException`, `BadRequestException`.

rorytorneymf: @kleyton-ohare I had a quick chat with Dermot regarding this. 

It may be useful to look through the code history to try and determine why some of the catch blocks are different, i.e. if it's intentional, or maybe if some catch blocks were missed when updating others etc.

If you find that the existing code is correct and we should maintain the same behaviour, then instead of having one mapException method, we could maybe have 2 or 3 mapException methods to encapsulate the common behaviour.

job-service/src/main/java/com/hpe/caf/services/job/api/DatabaseHelper.java
```diff
@@ -403,7 +415,15 @@ public boolean canJobBeProgressed(final String partitionId, final String jobId)
...
-            throw mapSqlException(se);
+            final String sqlState = se.getSQLState();
+            if (sqlState.equals(POSTGRES_NO_DATA_FOUND_ERROR_CODE)) {
+                // job missing - return false
+            } else if (sqlState.startsWith(POSTGRES_CONNECTION_EXCEPTION_ERROR_CODE_PREFIX)) {
+                // Connection exception
+                throw new ServiceUnavailableException(se.getMessage());
+            } else {
+                throw se;
+            }
```
kyra-ohare: It is not possible to create a method to encapsulate this block because the new method will complain the missing return statement from sqlState.equals(POSTGRES_NO_DATA_FOUND_ERROR_CODE)

job-service/src/main/java/com/hpe/caf/services/job/api/DatabaseHelper.java
```diff
@@ -187,12 +187,12 @@ public long getJobsCount(final String partitionId, String jobIdStartsWith, Strin
...
             }
         } catch (final SQLException se) {
-            throw mapSqlException(se);
+            throw mapSqlConnectionException(se);
         }
 
         return jobsCount;
     }
-
+
```
kyra-ohare: I have fixed this already

kyra-ohare (commented):
The two catches for methods canJobBeProgressed and isJobActive were not possible to encapsulate

job-service/src/main/java/com/hpe/caf/services/job/api/DatabaseHelper.java
rorytorneymf: You can encapsulate:

```
final String sqlState = se.getSQLState();
if (sqlState.equals(POSTGRES_NO_DATA_FOUND_ERROR_CODE)) {
   // job missing - return false
} else if (sqlState.startsWith(POSTGRES_CONNECTION_EXCEPTION_ERROR_CODE_PREFIX)) {
   // Connection exception
   throw new ServiceUnavailableException(se.getMessage());
}  else {
   throw e;	
}

```
into a method that looks like this:

```
private static void throwIfUnexpectedException(final SqlException sqlException) throws Exception {
    final String sqlState = se.getSQLState();
    if (sqlState.equals(POSTGRES_NO_DATA_FOUND_ERROR_CODE)) {
       // job missing - don't throw anything, return void
    } else if (sqlState.startsWith(POSTGRES_CONNECTION_EXCEPTION_ERROR_CODE_PREFIX)) {
       // Connection exception
       throw new ServiceUnavailableException(se.getMessage());
    }  else {
       throw e;	
    }
}
```

kleyton-ohare:
> I think you can include the original SqlException in the constructor here so we get the original cause in the stack trace in the logs:
> 
> `return new ServiceUnavailableException(se.getMessage(), se);`

kyra-ohare:
Added three utility methods to map SQLException in different behaviours.

rorytorneymf (approved):
Looks fine to me Kleyton, good work. The next steps are:

1) Merge the master branch into this branch to make sure you're up to date:

```
# git checkout master
# git pull

# git checkout <branch>
# git merge master

# Resolve conflicts...

# git add .
# git commit -m "Merge master into <branch>"
```

2) Merge this PR
3) Monitor the Jenkins master build to make sure it passes after you merge
4) Monitor the  darwin-job-service Jenkins master build to make sure it passes after (3) passes
5) Move the Jira to Awaiting QA

Use https://docs.google.com/spreadsheets/d/1taaQ0Bm6U9TPsmMaiWM9c3vKyubOvEwod38nuWOu5vM/edit#gid=1261077722 to find links to all the Jenkins builds

PR #226, opened 2021-02-18
SCMOD-10943: Filter recursive call to avoid querying entire system

dgibson-microfocus:
https://portal.digitalsafe.net/browse/SCMOD-10943

### DIFF
--- a/job-service-db/src/main/resources/procedures/processFailedDependentJobs.sql
+++ b/job-service-db/src/main/resources/procedures/processFailedDependentJobs.sql
@@ -37,10 +37,11 @@ BEGIN
             SELECT partition_id, dependent_job_id, job_id
             FROM job_dependency
             WHERE partition_id = in_partition_id AND job_id = in_job_id
-            UNION ALL
+            UNION
             SELECT adj.partition_id, adj.dependent_job_id, jd.job_id
             FROM all_job_dependencies adj
             INNER JOIN job_dependency jd ON adj.partition_id = jd.partition_id AND adj.job_id = jd.dependent_job_id
+            WHERE (jd.job_id <> '') IS NOT TRUE
         )
         SELECT job_id FROM all_job_dependencies;
 
--- a/job-service-db/src/main/resources/procedures/processFailedDependentJobs.sql
+++ b/job-service-db/src/main/resources/procedures/processFailedDependentJobs.sql
@@ -37,7 +37,7 @@ BEGIN
             SELECT partition_id, dependent_job_id, job_id
             FROM job_dependency
             WHERE partition_id = in_partition_id AND job_id = in_job_id
-            UNION ALL
+            UNION
             SELECT adj.partition_id, adj.dependent_job_id, jd.job_id
             FROM all_job_dependencies adj
             INNER JOIN job_dependency jd ON adj.partition_id = jd.partition_id AND adj.job_id = jd.dependent_job_id
--- a/job-service-db/src/main/resources/procedures/processFailedDependentJobs.sql
+++ b/job-service-db/src/main/resources/procedures/processFailedDependentJobs.sql
@@ -36,14 +36,14 @@ BEGIN
         WITH RECURSIVE all_job_dependencies AS (
             SELECT partition_id, dependent_job_id, job_id
             FROM job_dependency
-            WHERE partition_id = in_partition_id AND job_id = in_job_id
+            WHERE partition_id = in_partition_id AND dependent_job_id = in_job_id
             UNION
             SELECT adj.partition_id, adj.dependent_job_id, jd.job_id
             FROM all_job_dependencies adj
             INNER JOIN job_dependency jd ON adj.partition_id = jd.partition_id AND adj.job_id = jd.dependent_job_id
             WHERE adj.partition_id = in_partition_id
         )
-        SELECT job_id FROM all_job_dependencies;
+        SELECT DISTINCT job_id FROM all_job_dependencies
 
     -- Ensure that no other `job_dependency` deletion can run until we've committed.
     -- Lock ALL possibly conflicting rows up-front to avoid deadlocks.
--- a/job-service-db/src/main/resources/procedures/processFailedDependentJobs.sql
+++ b/job-service-db/src/main/resources/procedures/processFailedDependentJobs.sql
@@ -36,13 +36,13 @@ BEGIN
         WITH RECURSIVE all_job_dependencies AS (
             SELECT partition_id, dependent_job_id, job_id
             FROM job_dependency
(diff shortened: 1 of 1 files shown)


dermot-hardy (approved):

job-service-db/src/main/resources/procedures/processFailedDependentJobs.sql
```diff
@@ -36,13 +36,13 @@ BEGIN
         WITH RECURSIVE all_job_dependencies AS (
             SELECT partition_id, dependent_job_id, job_id
             FROM job_dependency
+            WHERE partition_id = in_partition_id AND job_id = in_job_id
             UNION ALL
             SELECT adj.partition_id, adj.dependent_job_id, jd.job_id
             FROM all_job_dependencies adj
```
dermot-hardy, line 42: Does this definitely work?  I see it's going back to `all_job_dependencies` which is now different.

dgibson-microfocus: It should, as previously it was building the table for every job querying that where as now it just builds the table for this job. 
It seemed to on the one call I tried. I don't see any IT for it which is slightly concerning though.

dermot-hardy: Make sure to test it with multiple levels of dependencies.

dermot-hardy (changes requested):
I don't understand this anymore.

job-service-db/src/main/resources/procedures/processFailedDependentJobs.sql
```diff
@@ -36,13 +36,14 @@ BEGIN
         WITH RECURSIVE all_job_dependencies AS (
             SELECT partition_id, dependent_job_id, job_id
             FROM job_dependency
-            UNION ALL
+            WHERE partition_id = in_partition_id AND job_id = in_job_id
+            UNION
```
dermot-hardy, line 40: Why this this moving to `UNION`?  Surely that is a more expensive operator than `UNION ALL`?

job-service-db/src/main/resources/procedures/processFailedDependentJobs.sql
```diff
@@ -36,13 +36,14 @@ BEGIN
...
             SELECT adj.partition_id, adj.dependent_job_id, jd.job_id
             FROM all_job_dependencies adj
             INNER JOIN job_dependency jd ON adj.partition_id = jd.partition_id AND adj.job_id = jd.dependent_job_id
+            WHERE (jd.job_id <> '') IS NOT TRUE
```
dermot-hardy: Is this just the same as `WHERE jd.job_id = ''`?  Why are we doing this?

dermot-hardy (changes requested):

job-service-db/src/main/resources/procedures/processFailedDependentJobs.sql
dgibson-microfocus: I meant to write the equivalent of `WHERE jd.job_id IS NOT NULL AND jd.job_id <> ''` and ended up flipping the meaning by mistake. This obviously meant the iterative part of the query never returned results and I mistakenly believed I'd solved the performance issue as the query plan/stats where suddenly much better.

job-service-db/src/main/resources/procedures/processFailedDependentJobs.sql
```diff
@@ -36,13 +36,14 @@ BEGIN
         WITH RECURSIVE all_job_dependencies AS (
             SELECT partition_id, dependent_job_id, job_id
             FROM job_dependency
+            WHERE partition_id = in_partition_id AND job_id = in_job_id
             UNION ALL
             SELECT adj.partition_id, adj.dependent_job_id, jd.job_id
             FROM all_job_dependencies adj
             INNER JOIN job_dependency jd ON adj.partition_id = jd.partition_id AND adj.job_id = jd.dependent_job_id
+            WHERE adj.partition_id = in_partition_id
```
dgibson-microfocus, line 44: Adding this `WHERE` allows the recursive clause to actually use the index on `job_dependency` instead of doing a seq scan of the table reach iteration.

dgibson-microfocus:
@dermot-hardy The query as it is now is much better than it's previous version (See query plans below) but will still not scale for relatively large dependency trees. (e.g. the query won't work on most of the jobs in `jobservice_scrap2` on victor) 

Original recursive CTE plan:
image

New recursive CTE plan:
image

Note those are just the estimate plans as the original query blows up.

job-service-db/src/main/resources/procedures/processFailedDependentJobs.sql
dgibson-microfocus: I've tested this on a job with 3 levels of dependents and the full hierarchy was correctly returned.

job-service-db/src/main/resources/procedures/processFailedDependentJobs.sql
dgibson-microfocus: Using `UNION` drastically reduced the number of iterations the loop does. To the point where `UNION ALL` would used 100s of GB of disk and not finish while `UNION` was sub-second.

job-service-db/src/main/resources/procedures/processFailedDependentJobs.sql
```diff
@@ -36,13 +36,14 @@ BEGIN
...
+            WHERE partition_id = in_partition_id AND dependent_job_id = in_job_id
+            UNION
             SELECT adj.partition_id, adj.dependent_job_id, jd.job_id
             FROM all_job_dependencies adj
             INNER JOIN job_dependency jd ON adj.partition_id = jd.partition_id AND adj.job_id = jd.dependent_job_id
+            WHERE adj.partition_id = in_partition_id
         )
-        SELECT job_id FROM all_job_dependencies
-        WHERE dependent_job_id = in_job_id AND partition_id = in_partition_id;
+        SELECT DISTINCT job_id FROM all_job_dependencies
```
dgibson-microfocus: `SELECT DISTINCT` over `SELECT`as even when using `UNION` there are still duplicate `job_ids` in the result set. So I've removed them here to make the rest of the function quicker.

dermot-hardy (changes requested):
The build is failing.

Also, is the comment about the re-worked query not working on most of the jobs in `jobservice_scrap2` on victor still applicable?

dgibson-microfocus:
> 
> 
> The build is failing.
> 
> Also, is the comment about the re-worked query not working on most of the jobs in `jobservice_scrap2` on victor still applicable?

The build is failing due to a missing `;` and no, that no longer applies.

dermot-hardy (approved):

PR #234, opened 2021-03-08
SCMOD-12730: Add job pause, resume and status endpoints

rorytorneymf:

Adding 3 new endpoints to pause, resume, and get the status of the job. The pause and resume endpoints currently just update the status of the job in the job service database. 

Future PRs will cover additional updates to store and recover messages when a job is paused and resumed. See the Jira linked above for full implementation details.


job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
```diff
@@ -386,8 +465,7 @@ definitions:
         format: int64
         description: The time the job status or progress last changed
       status:
-        type: string
-        description: String value of the job status enum, possible values are Completed, Active, Waiting, Failed, Paused and Cancelled.
+        $ref: "#/definitions/job-status"
```
rorytorneymf, line 469: This change was not strictly necessary but felt it was cleaner to use an enum vs a string here now for the response from the existing endpoints, as thats what the new `status` endpoint is using.

This does mean a breaking change to the job-service-client though, as the generated client code will have to handle a `JobStatus` enum now rather than a `String`:

https://github.com/JobService/job-service-client/pull/3/files#diff-d5ca635f7b96abc24783be438f79d8ec92d70a7a32b64ae7b2049d159697c017R11

The alternative is to just leave it as a `String` and use a `String` rather than an enum for the new status endpoint as well.

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
```diff
@@ -218,6 +217,36 @@ paths:
...
+    get:
+      tags:
+        - Jobs
+      summary: Gets the status of the job
+      description: Gets the status of the specified job.
+      operationId: getJobStatus
+      responses:
+        200:
+          schema:
+            $ref: "#/definitions/job-status"
```
rorytorneymf: The new status endpoint just returns a string in the response, e.g. "Active", "Cancelled" etc. It does not wrap this in a JSON object like:

```
{
  "status": "Active"
}
```

I think this is ok, but just wanted to highlight it. It's consistent with the 2 other endpoints here that just return a single value (the `isActive` and `jobstats/count` endpoints), and results in a bit less boilerplate code around wrapping the response and unwrapping it in the client code etc.

I believe its fine from a JSON validity point of view:

https://stackoverflow.com/questions/7487869/is-this-simple-string-considered-valid-json

job-service-db/src/main/resources/procedures/pauseJob.sql
```diff
@@ -0,0 +1,67 @@
...
+    WHERE partition_id = in_partition_id
+        AND job_id = in_job_id
+    FOR UPDATE;
+
+    IF NOT FOUND THEN
+        RAISE EXCEPTION 'job_id {%} not found', in_job_id USING ERRCODE = 'P0002'; -- sqlstate no_data_found
+    END IF;
+
+    IF v_job_status NOT IN ('Active', 'Paused', 'Waiting') THEN
+        RAISE EXCEPTION 'job_id {%} cannot be paused as it has a status of {%}. Only jobs with a status of Active or Waiting can be paused.',
```
dermot-hardy: Goes over the 138-character guideline.

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
```diff
@@ -218,6 +217,36 @@ paths:
...
+          across different services.
+      - $ref: '#/parameters/jobId'
+    get:
+      tags:
+        - Jobs
+      summary: Gets the status of the job
+      description: Gets the status of the specified job.
+      operationId: getJobStatus
+      responses:
+        200:
```
dermot-hardy, line 237: According to https://editor.swagger.io/:

image

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
```diff
@@ -403,7 +481,17 @@ definitions:
           type: string
         description: |
           A map of meta data associated with this job. Note that the keys can only include alphanumeric, '_', '-' and
-          ':' characters
+          ':' characters
```
dermot-hardy: Why is this line changing?

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
dermot-hardy: I'm not bothered by the client breaking change.

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
dermot-hardy: Ok.  It probably wasn't the best idea to start down the line of non-object responses but the decision makes sense from a consistency perspective.

dermot-hardy:
```suggestion
            $ref: "#/definitions/job-status"
```

job-service-db/src/main/resources/procedures/resumeJob.sql
```diff
@@ -0,0 +1,67 @@
...
+    FROM job
+    WHERE partition_id = in_partition_id
+        AND job_id = in_job_id
+    FOR UPDATE;
+
+    IF NOT FOUND THEN
+        RAISE EXCEPTION 'job_id {%} not found', in_job_id USING ERRCODE = 'P0002'; -- sqlstate no_data_found
+    END IF;
+
+    IF v_job_status <> 'Paused' THEN
```
dermot-hardy: We allow a paused job to be Paused without complaining but we don't allow a resumed job to be Resumed.  It seems inconsistent.

job-service/src/main/java/com/hpe/caf/services/job/api/JobsStatus.java
```diff
@@ -0,0 +1,82 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hpe.caf.services.job.api;
+
+import com.hpe.caf.services.job.api.*;
```
dermot-hardy: Unused import.

job-service-acceptance-tests/src/test/java/com/hpe/caf/jobservice/acceptance/JobServiceEndToEndIT.java
```diff
@@ -1173,6 +1177,148 @@ public void testJobDeletionWithLabels() throws Exception {
...
+        createJob(jobId, true);
+        waitUntilJobCompletes(jobId);
+        try {
+            pauseJob(jobId);
+        } catch (final ApiException e) {
+            Assert.assertEquals(e.getCode(), 400, "Unexpected HTTP response code");
+            Assert.assertEquals(e.getMessage(),
+                                "{\"message\":\"ERROR: job_id {" + jobId + "} cannot be paused as it has a status of {Completed}. "
+                                + "Only jobs with a status of Active or Waiting can be paused.\\n  "
+                                + "Where: PL/pgSQL function pause_job(character varying,character varying) line 23 at RAISE\"}",
```
dermot-hardy: It's not a great idea to test the exact error message like this.  In this case even adding a blank line to the function would cause the tests to break, nevermind how the message might change with a bigger change like a PostgreSQL upgrade.

dermot-hardy (approved):

job-service-db/src/main/resources/procedures/resumeJob.sql
rorytorneymf: Updated to allow a resumed job to be resumed as well.

job-service-acceptance-tests/src/test/java/com/hpe/caf/jobservice/acceptance/JobServiceEndToEndIT.java
rorytorneymf: Updated to use `contains` rather than `equals` to just check the error message for the subset of text of interest rather than the entire thing.

PR #235, opened 2021-03-08
SCMOD-8730: Support expiry policies

xbreizh:
https://portal.digitalsafe.net/browse/SCMOD-8730


xbreizh:
@dermot-hardy  I think that I understand what you mean with the RFC 3339 but that would mean passing a date directly so the calculation would have to be made on client's end.
What I was thinking of was to make it easier on client's side and have them specify the date reference as a string then a number of days as integer and the calculation would be done whenever we run the checkExpiration job.
But I can go with a date if you prefer, Swagger uses RFC339 as [default already](https://swagger.io/specification/) 
Here, I use the same object in and out (for newJob and job), but I could return an expirationDate instead for the job. I suppose that it would make more sense.
I tried having a default object to return but failed to find it so I am passing the different statuses and set a default for each element.

An example on json would be:

```json{
  "name": "Ingest documents submitted in batches A and B",
  "description": "This job ingests all of the documents that were in the A and B batches.",
  "expiration-policy": {
    "Active": {
      Policy:{
        "action": "delete",
        "referenceDate": "createDate",
        "duration": "70"
      }
    },
    "Failed": {
      Policy:{
        "action": "keep",
        "referenceDate": "lastUpdateDate",
        "duration": "7"
      }
    },
    "Completed": {
      Policy:{
        "action": "delete",
        "referenceDate": "lastUpdateDate",
        "duration": "7"
      }
    },
    "Waiting": {
      Policy:{
        "action": "delete",
        "referenceDate": "lastUpdateDate",
        "duration": "7"
      }
    },
    "Cancelled": {
      Policy:{
        "action": "delete",
        "referenceDate": "createDate",
        "duration": "7"
      }
    },
    "Paused": {
      Policy:{
        "action": "delete",
        "referenceDate": "lastUpdateDate",
        "duration": "7"
      }
    }  
  }, 
  "labels": {
    "repoId": "x37",
    "runNumber": "12",
    "workUnitId": "371"
  },
  "type": "batch-ingestion-job",
  "parameters": {
    "batches": ["batch-a", "batch-b"],
    "operationMode": "EXTRACT",
    "grammarMap": {
      "pii": []
    },
    "extractMaxMatches": 100
  }
}```

for the whole policy setup. But using the default, we should be able to pass only:

```json{
  "name": "Ingest documents submitted in batches A and B",
  "description": "This job ingests all of the documents that were in the A and B batches.",
  "expiration-policy": {
    "Active": {
      Policy:{
        "action": "delete",
        "referenceDate": "createDate",
        "duration": "70"
      }
    },
    "Failed": {
      Policy:{
        "action": "keep",
        "referenceDate": "lastUpdateDate",
        "duration": "7"
      }
    }
  }, 
  "labels": {
    "repoId": "x37",
    "runNumber": "12",
    "workUnitId": "371"
  },
  "type": "batch-ingestion-job",
  "parameters": {
    "batches": ["batch-a", "batch-b"],
    "operationMode": "EXTRACT",
    "grammarMap": {
      "pii": []
    },
    "extractMaxMatches": 100
  }
}```

File: job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml [hunk 1]
```diff
@@ -298,6 +298,8 @@ definitions:
       description:
         type: string
         description: The description of the job
+      expiration-policy:
```
dermot-hardy: Rightly or wrongly the other properties in this object use camelCase, so we wouldn't want a new one to use kebab-case.

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
```diff
@@ -335,6 +337,42 @@ definitions:
...
+        $ref: "#/definitions/policy"
+      "Failed":
+        $ref: "#/definitions/policy"
+      "Cancelled":
+        $ref: "#/definitions/policy"
+      "Waiting":
+        $ref: "#/definitions/policy"
+      "Paused":
+        $ref: "#/definitions/policy"
+    description: The expiration policy to be applied on the job
```
dermot-hardy, line 451: How can I specify a default policy here?

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
```diff
@@ -335,6 +337,42 @@ definitions:
...
+      referenceDate:
+        type: string
+        enum:
+          - createDate
+          - lastUpdateDate
+        default: createDate
+      duration:
+        type: integer
+        format: int32
+        default: 5
```
dermot-hardy: How about we combine `referenceDate` and `duration` into an `expiryTime`?  It could be an exact time like `2021-04-12T23:20:50.52Z`, or it could be an expression like `lastUpdateTime+P1D`.

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
```diff
@@ -335,6 +337,42 @@ definitions:
...
+      action:
+        type: string
+        enum:
+          - keep
+          - delete
+        default: keep
```
dermot-hardy: Would `expire` be a better word than `keep`?

dermot-hardy (changes requested):

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
xbreizh: Done

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
xbreizh: Done

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
xbreizh: That would mean that they have to provide the date itself and not just its reference? I adapted the code to reflect that and now we have an expiryTime as a date-time (which is RFC 3339)

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
xbreizh: Done

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
xbreizh: Done

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
xbreizh: Done

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
```diff
@@ -335,6 +337,41 @@ definitions:
         example:
           "tag:4": "4"
           owner: "bob"
+  expirationPolicy:
+    type: object
+    properties:
```
dermot-hardy, line 436: Aren't you introducing a new Expired status, so shouldn't it be listed here?  Is it special since the `expire` action makes no sense for it?

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
```diff
@@ -335,6 +337,41 @@ definitions:
...
+      "Default":
+        $ref: "#/definitions/policy"
+    description: The expiration policy to be applied on the job
+  policy:
+    type: object
+    properties:
+      expiryTime:
+        type: string
+        format: date-time
+        default: 2222-04-12T23:20:50.52Z
```
dermot-hardy: You would never have a default expiry time.

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
```diff
@@ -335,6 +337,41 @@ definitions:
...
+        $ref: "#/definitions/policy"
+      "Default":
+        $ref: "#/definitions/policy"
+    description: The expiration policy to be applied on the job
+  policy:
+    type: object
+    properties:
+      expiryTime:
+        type: string
+        format: date-time
```
dermot-hardy: Declaring this as a `date-time` would mean that it could only accept exact times.  I could imagine that we might want to pass an exact expiry time but equally I think we should also support accepting an expiry expression like `lastUpdateTime+P1D`.

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
```diff
@@ -335,6 +337,41 @@ definitions:
...
+        default: 2222-04-12T23:20:50.52Z
+      action:
+        $ref: "#/definitions/action"
+    description: The expiration details to be applied on the job
+  action:
```
dermot-hardy: Too easily confused with `worker-action`.

dermot-hardy (changes requested):

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
xbreizh: I wasn't to list it here, but I supposed we could have a policy to expire the jobs and keep them for a while, then delete them after another period.
If we don't pass a policy for a status, then nothing is set for it.

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
xbreizh: Replaced with regex

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
xbreizh: used ´pattern: '^((lastUpdateDate)|(createDate))(\+P)([1-9]\d?|[12]\d{2}|3[0-5]\d|36[0-5])D$'´ 
action taken between 1 and 365 days. We can change that, I just thought we might need a max duration

By the way, I noticed that the pattern is not being passed to the classes. It has been raised some time ago [here](https://github.com/swagger-api/swagger-codegen/issues/7980) but I don't think that anyone works on it.

The way the classes are manually copied over, we can easily add it (it hasn't been done so far apparently)

I made a test with openAPI but same result -> restored back to Swagger 2 as the upgrade is not part of that Jira

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
xbreizh: now expirationOperation

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
```diff
@@ -335,6 +337,55 @@ definitions:
...
+        enum:
+          - delete
+    description: The expiration details to be applied on the job
+  policy:
+    type: object
+    properties:
+      expiryTime:
+        type: string
+        pattern: '^((lastUpdateDate)|(createDate))(\+P)([1-9]\d?|[12]\d{2}|3[0-5]\d|36[0-5])D$'
+        default: "createDate+10D"
```
dermot-hardy: Why do we have a default expiry?

From the Jira:
https://portal.digitalsafe.net/browse/SCMOD-8730?focusedCommentId=3096664&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-3096664
> The default should be no expiry...

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
```diff
@@ -335,6 +337,55 @@ definitions:
...
+        type: string
+        enum:
+          - delete
+    description: The expiration details to be applied on the job
+  policy:
+    type: object
+    properties:
+      expiryTime:
+        type: string
+        pattern: '^((lastUpdateDate)|(createDate))(\+P)([1-9]\d?|[12]\d{2}|3[0-5]\d|36[0-5])D$'
```
dermot-hardy: - The contract uses [`lastUpdateTime`](https://github.com/JobService/job-service/blob/v3.5.0-source/job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml#L370) and [`createTime`](https://github.com/JobService/job-service/blob/v3.5.0-source/job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml#L366).  It makes no sense to use `lastUpdateDate` and `createDate` as they are not the property names presented to the consumer.
- I'm not going to study this regular-expression but I can see that it ends in `D`, so I assume it only supports days.  I thought we were going to use the RFC 3339 duration format.  I should be able to specify `lastUpdateTime+P20M` to expire after 20 minutes but this regular expression is clearly not going to accept that.
- What happened with supporting an exact time?  In [this](https://github.com/JobService/job-service/pull/235#discussion_r590916801) comment I said: "It could be an exact time like `2021-04-12T23:20:50.52Z`, or it could be an expression like `lastUpdateTime+P1D`".  In [this](https://github.com/JobService/job-service/pull/235#discussion_r592019221) one I said: "I could imagine that we might want to pass an exact expiry time...".  These comments appear to have been ignored and you're now trying to lock us in to a duration instead of supporting both like I repeatedly said.
- Finally, just for the record I can see the attractiveness of specifying a pattern if you can get it right, but the reality is that it is a nice-to-have.  I'd be perfectly happy with a few decent examples in the description.

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
```diff
@@ -335,6 +337,55 @@ definitions:
...
+    description: The expiration policy to be applied on the job
+  deletePolicy:
+    type: object
+    properties:
+      expiryTime:
+        type: string
+        pattern: '^((lastUpdateDate)|(createDate))(\+P)([1-9]\d?|[12]\d{2}|3[0-5]\d|36[0-5])D$'
+        default: "createDate+10D"
+      expirationOperation:
+        type: string
```
dermot-hardy, line 461: You created a separate object (definition) for this will the normal policy but you haven't done that here.  I think it would be better to have the symmetry of either doing it in both places or not doing it in either.

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
dermot-hardy: There is no need for a maximum duration.  There is a need to be able to specify less than a day.

dermot-hardy: > By the way, I noticed that the pattern is not being passed to the classes. It has been raised some time ago here but I don't think that anyone works on it.

Yes the tooling support can vary quite a bit depends on what features you use.

> The way the classes are manually copied over, we can easily add it (it hasn't been done so far apparently)

Best just to enforce it in the implementation.

> I made a test with openAPI but same result -> restored back to Swagger 2 as the upgrade is not part of that Jira

Thank you.

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
```diff
@@ -335,6 +337,55 @@ definitions:
...
+    type: object
+    properties:
+      expiryTime:
+        type: string
+        pattern: '^((lastUpdateDate)|(createDate))(\+P)([1-9]\d?|[12]\d{2}|3[0-5]\d|36[0-5])D$'
+        default: "createDate+10D"
+      expirationOperation:
+        $ref: "#/definitions/expirationOperation"
+    description: The expiration details to be applied on the job
+  expirationOperation:
```
dermot-hardy: You're mixing up definition names and property names.  Definition names are internal and flat.  Property names are public and have a context.  Just because you changed this to `expirationOperation` it doesn't follow that the property name should have changed.

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
xbreizh: Replaced with "no expiry"

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
xbreizh: - Updated for  lastUpdateTime and createTime.
- Adapted regex for accepting RFC3339 format date or lastUpdateTime+P1[DYHM] pattern (https://regex101.com/r/badcKz/1/)
- Added a few examples
Comments haven't been ignored, I just couldn't find how to pass either a date or a string combination as a value, that's why I wanted to have a chat with you. 
Passing a final date I get that, but I don t think we would add the client to provide us with the createTime or lastUpdateTime as we don t want to check that, so I went for a reference to it instead that we can easily convert then into a date. and the string format was conflicting with the date-time which is already RFC3339
Another solution could be to have an object taking a string and a date, then in the treatment, we use the date if provided, otherwise the string, otherwise "no expiry"
I guess we could also take any sort of string  in, and only provide some examples but I understand it's good practice to catch an issue early on

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
xbreizh: extracted the object as expiryTimePattern

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
xbreizh: replaced with operation

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
xbreizh: Added options for hour, minute and year in addition to day

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
```diff
@@ -415,6 +417,56 @@ definitions:
...
+      expirationOperation:
+        type: string
+        enum:
+          - delete
+    description: The expiration details to be applied on the job
+  expiryTimePattern:
+    type: string
+    pattern: '^(?<fullyear>\d{4})-(?<month>0[1-9]|1[0-2])-(?<mday>0[1-9]|[12][0-9]|3[01])T(?<hour>[01][0-9]|2[0-3]):(?<minute>[0-5][0-9]):(?<second>[0-5][0-9]|60)(?<secfrac>\.[0-9]+)?(Z|(\+|-)(?<offset_hour>[01][0-9]|2[0-3]):(?<offset_minute>[0-5][0-9]))|(((lastUpdateTime)|(createTime))(\+P)[1-9]\d*[DYHM])|(no expiry)$'
+    example: 2021-04-12T23:20:50.52Z, lastUpdateTime+P1D, createTime+P90M, no expiry
+    default: no expiry
```
dermot-hardy: Just don't specify the default in the contract at all:
```suggestion
```

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
```diff
@@ -415,6 +417,56 @@ definitions:
...
+      expiryTime:
+        $ref: "#/definitions/expiryTimePattern"
+      expirationOperation:
+        type: string
+        enum:
+          - delete
+    description: The expiration details to be applied on the job
+  expiryTimePattern:
+    type: string
+    pattern: '^(?<fullyear>\d{4})-(?<month>0[1-9]|1[0-2])-(?<mday>0[1-9]|[12][0-9]|3[01])T(?<hour>[01][0-9]|2[0-3]):(?<minute>[0-5][0-9]):(?<second>[0-5][0-9]|60)(?<secfrac>\.[0-9]+)?(Z|(\+|-)(?<offset_hour>[01][0-9]|2[0-3]):(?<offset_minute>[0-5][0-9]))|(((lastUpdateTime)|(createTime))(\+P)[1-9]\d*[DYHM])|(no expiry)$'
```
dermot-hardy: If we're going to have a token for no expiry I'd rather that it is "none":
```suggestion
    pattern: '^(?<fullyear>\d{4})-(?<month>0[1-9]|1[0-2])-(?<mday>0[1-9]|[12][0-9]|3[01])T(?<hour>[01][0-9]|2[0-3]):(?<minute>[0-5][0-9]):(?<second>[0-5][0-9]|60)(?<secfrac>\.[0-9]+)?(Z|(\+|-)(?<offset_hour>[01][0-9]|2[0-3]):(?<offset_minute>[0-5][0-9]))|(((lastUpdateTime)|(createTime))(\+P)[1-9]\d*[DYHM])|(none)$'
```

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
dermot-hardy: The regex may still need some adjustment - I notice that it still matched when I added "xx" at the start:

image

dermot-hardy: Note that when you submit a PR that you can add comments yourself about things you're not sure about or want to discuss further.  Without comments I can't tell the difference between being ignored and not knowing how to do something that's been discussed.  Some examples from others:
https://github.com/JobService/job-service/pull/234#discussion_r592315769 (Rory)
https://github.com/JobService/job-service/pull/234#discussion_r592320909 (Rory)
https://github.houston.softwaregrp.net/Verity/darwin-job-service/pull/97#discussion_r463281 (David)

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
```diff
@@ -415,6 +417,56 @@ definitions:
         example:
           "tag:4": "4"
           owner: "bob"
+  expirationPolicy:
```
dermot-hardy, line 434: Should we shorten this to `expiry`?

dermot-hardy, line 434: Should we shorten this to `expiry`?

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
```diff
@@ -415,6 +417,56 @@ definitions:
...
+        $ref: "#/definitions/deletePolicy"
+      "Default":
+        $ref: "#/definitions/policy"
+    description: The expiration policy to be applied on the job
+  deletePolicy:
+    type: object
+    properties:
+      expiryTime:
+        $ref: "#/definitions/expiryTimePattern"
+      expirationOperation:
```
dermot-hardy: For consistency:
```suggestion
      operation:
```

dermot-hardy (approved):
Approval relates to contract only.

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
xbreizh: Renamed

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
xbreizh: Renamed

xbreizh: Renamed

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
xbreizh: Done and fixed the issues about extra characters before or after the expression

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
xbreizh: Removed

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
dermot-hardy: Oops I meant to put that comment on the property not the definition.  `<excuses>`With GitHub going down earlier I was really struggling to get those comments applied so I guess that's how I screwed up.`</excuses>`

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
```diff
@@ -378,6 +378,8 @@ definitions:
       description:
         type: string
         description: The description of the job
+      expirationPolicy:
```
dermot-hardy: Should we shorten this to `expiry`?  (And you can put the definition back to `expirationPolicy` if that makes more sense.)

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
```diff
@@ -415,6 +417,58 @@ definitions:
...
+    properties:
+      expiryTime:
+        $ref: "#/definitions/expiryTimePattern"
+      operation:
+        type: string
+        enum:
+          - delete
+    description: The expiration details to be applied on the job
+  expiryTimePattern:
+    type: object
```
dermot-hardy: Why turn this into an object?

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
dermot-hardy: I've commented on the correct line that I meant to [here](https://github.com/JobService/job-service/pull/235#discussion_r593487072).

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
xbreizh: Done

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
xbreizh: If I pass it as a String to a new definition, it becomes an object and I get the following on compilation:

> incompatible types: java.lang.String cannot be converted to com.hpe.caf.services.job.client.model.ExpiryTimeDefinition

I couldn't manage to cast it either. So from that, I think that we can either create a proper object or duplicate the definition.
Found a report of that issue [here](https://github.com/swagger-api/swagger-parser/issues/243) or [here](https://github.com/swagger-api/swagger-codegen/issues/2314)

dermot-hardy (approved):
Approval relates to contract only.

job-service/pom.xml
```diff
@@ -269,6 +274,12 @@
             <version>7.1.0</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter</artifactId>
+            <version>RELEASE</version>
```
dermot-hardy: Why are you specifying `<version>RELEASE</version>` here?

image

Is your IDE warning you about this?

job-service/pom.xml
```diff
@@ -97,6 +97,11 @@
             <groupId>com.github.jobservice</groupId>
             <artifactId>job-service-util</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
+            <scope>test</scope>
+        </dependency>
```
dermot-hardy: Why is this up here instead of down with the other `junit-jupiter` dependency that you're adding?

job-service/src/main/java/com/hpe/caf/services/job/api/generated/model/Policy.java
```diff
@@ -0,0 +1,127 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hpe.caf.services.job.api.generated.model;
+//
+// Source code recreated from a .class file by IntelliJ IDEA
```
dermot-hardy: Where did this file come from?

job-service-db/src/main/resources/procedures/upsertJobPolicy.sql
```diff
@@ -0,0 +1,37 @@
...
+    in_status job_status,
+    in_reference_date VARCHAR(58),
+    in_duration INT*/
+)
+RETURNS VOID
+LANGUAGE plpgsql
+AS $$
+BEGIN
+
+    RAISE NOTICE 'Hello world';
```
dermot-hardy: Resubmit this for review with an explanation of what parts of it you want reviewed and what should be ignored.

dermot-hardy (changes requested):

job-service/pom.xml
xbreizh: No, I am not getting any warning about it. I see there was a case raised some time ago [here](https://youtrack.jetbrains.com/issue/IDEA-181023)

job-service/pom.xml
xbreizh: sorry, moved to the right place with the other test dependencies

job-service/src/main/java/com/hpe/caf/services/job/api/generated/model/Policy.java
xbreizh: I am not entirely sure, but from what I found, this could have been recreated if at a point I deleted it and it was still referenced elsewhere in the code. The IDE would recreate it from compile sources.
Maybe more likely, there was an issue with some cache being corrupt. I cleared those to make sure.

job-service-db/src/main/resources/procedures/upsertJobPolicy.sql
xbreizh: My bad, I have worked so far on the policy validation and that file is for the next step: persisting it (at that point, I just wanted to see if I was reaching out). I will rework the PR to be worthy of being reviewed, then resubmit

job-service/pom.xml
dermot-hardy: Surprising that the two free IDEs both warn about this and yet the commercial one doesn't.  Ok.

xbreizh:
@dermot-hardy DBConnection can be ignored (extra-spaces removal done by the IDE). Also, I saw that the acceptance tests have been disabled as part of [SCMOD-13004](https://portal.digitalsafe.net/browse/SCMOD-13004) where it says that we might refactor those. 
So, at the moment, I have only integrated the policies to the existing tests (when creating the jobs).
I manually checked that it works, but should I write the testing for it or should I hold off until 13004 is solved?
Also, there are quite a lot of changes since the last review, so please let me know if you want me to get you through it. Thx

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
```diff
@@ -415,6 +417,53 @@ definitions:
...
+        $ref: "#/definitions/deletePolicy"
+      "Default":
+        $ref: "#/definitions/policy"
+    description: The expiration policy to be applied on the job
+  deletePolicy:
+    type: object
+    properties:
+      expiryTime:
+        type: string
+        pattern: '^(?<fullyear>\d{4})-(?<month>0[1-9]|1[0-2])-(?<mday>0[1-9]|[12][0-9]|3[01])T(?<hour>[01][0-9]|2[0-3]):(?<minute>[0-5][0-9]):(?<second>[0-5][0-9]|60)(?<secfrac>\.[0-9]+)?(Z|(\+|-)(?<offset_hour>[01][0-9]|2[0-3]):(?<offset_minute>[0-5][0-9]))$|^(lastUpdateTime|createTime)\\+((P)\d+[DYM])$|^(lastUpdateTime|createTime)\\+(PT)?(\d+[HM])$|^none$'
```
dermot-hardy: The pattern changes don't seem to be right - 3 of the 5 examples don't match:
https://regex101.com/r/qJh6BR/1

job-service/src/main/java/com/hpe/caf/services/job/api/generated/model/ExpirationPolicy.java
```diff
@@ -0,0 +1,157 @@
...
+        } else {
+            return false;
+        }
+    }
+
+    public int hashCode() {
+        return Objects.hash(new Object[]{this.active, this.completed, this.failed, this.cancelled, this.waiting, this.paused, this.expired, this.defaultPolicy});
+    }
+
+    public String toString() {
```
dermot-hardy, line 134: The `@Override` annotations are missing.  Does IntelliJ not warn you about this?

image

job-service/src/main/java/com/hpe/caf/services/job/api/generated/model/ExpirationPolicy.java
```diff
@@ -0,0 +1,157 @@
...
+ */
+package com.hpe.caf.services.job.api.generated.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.hpe.caf.services.job.api.generated.model.DeletePolicy.ExpirationOperationEnum;
+import io.swagger.annotations.ApiModel;
+import io.swagger.annotations.ApiModelProperty;
+
+import java.util.ArrayList;
+import java.util.List;
```
dermot-hardy: Unused imports:

image

Is this file actually auto-generated?

job-service/src/main/java/com/hpe/caf/services/job/api/generated/model/ExpirationPolicy.java
```diff
@@ -0,0 +1,157 @@
...
+import java.util.Objects;
+
+@ApiModel(
+        description = "The expiration policy to be applied on the job"
+)
+public class ExpirationPolicy {
+    private static final int EXPIRATION_AFTER_LAST_UPDATE = 0;
```
dermot-hardy: This field is apparently unused.

image

job-service-db/src/main/resources/changelog-master.xml
```diff
@@ -34,5 +34,6 @@
    <include file="v3.4.0/changelog.xml" relativeToChangelogFile="true" />
    <include file="v3.5.0/changelog.xml" relativeToChangelogFile="true" />
    <include file="v4.0.0/changelog.xml" relativeToChangelogFile="true" />
+   <include file="v4.0.1/changelog.xml" relativeToChangelogFile="true" />
```
dermot-hardy: Why are we not just using the `v4.0.0` changelog?  Are you predicting that we're not going to get this functionality in?

job-service-db/src/main/resources/v4.0.1/changelog.xml
```diff
@@ -0,0 +1,45 @@
...
+                                       http://www.liquibase.org/xml/ns/dbchangelog
+                                       http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.3.xsd">
+    <changeSet id="create_expiration_operation_enum_type" author="Gamma Team">
+        <sql>CREATE TYPE EXPIRATION_OPERATION AS ENUM('Expire', 'Delete')</sql>
+    </changeSet>
+    <changeSet id="update_job_status_enum_type_add_expired" author="Gamma Team" runInTransaction="false">
+        <sql>ALTER TYPE JOB_STATUS ADD VALUE 'Expired';</sql>
+    </changeSet>
+    <changeSet id="create_policy_type" author="Gamma Team">
+        <sql>CREATE TYPE JOB_POLICY AS (partition_id VARCHAR(40), job_id VARCHAR(48), job_status JOB_STATUS, operation EXPIRATION_OPERATION, expiration_time VARCHAR(58))</sql>
```
dermot-hardy: Split across multiple lines to stay within the 138-character guideline.

job-service-db/src/main/resources/procedures/changelog.xml
```diff
@@ -295,6 +307,12 @@
                          schemaName="public">
         </createProcedure>
     </changeSet>
+    <changeSet id="create_upsert_joc_policy_proc" runOnChange="true" author="Gamma Team">
```
dermot-hardy:
```suggestion
    <changeSet id="create_upsert_job_policy_proc" runOnChange="true" author="Gamma Team">
```

dermot-hardy (changes requested):
Comments so far...

job-service-db/src/main/resources/procedures/changelog.xml
xbreizh: Updated

job-service-db/src/main/resources/v4.0.1/changelog.xml
xbreizh: Done

job-service-db/src/main/resources/changelog-master.xml
xbreizh: True, I got lost with that one. Merged with 4.0.0. I am bad at predicting anyways

job-service/src/main/java/com/hpe/caf/services/job/api/generated/model/ExpirationPolicy.java
xbreizh: Removed

job-service/src/main/java/com/hpe/caf/services/job/api/generated/model/ExpirationPolicy.java
xbreizh: Removed extra imports. Not autogenerated. I believe that none of those from the job service module are auto-generated

job-service/src/main/java/com/hpe/caf/services/job/api/generated/model/ExpirationPolicy.java
xbreizh: No warning. Also, there is no @Override annotation on the generated classes under module job-service-internal-client in target/classes/com/hpe/caf/services/job/client/model.
But, we do have them in the classes created under the generated/model folder in the job-service module (Job, NewJob ...)
I now added them to ExpirationPolicy too to match it

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
xbreizh: looks like an extra / slipped in there when I pasted it. Fixed here: https://regex101.com/r/joR6PN/1

xbreizh: Should we add Seconds to be compliant with RFC3339. I haven't so far as we may not be able to process as fast as required, or we could set a minimum duration as info only, so we can take any value but specify that the minimum we need is 60 seconds for example?

job-service/src/main/java/com/hpe/caf/services/job/utilities/ExpirationPolicyHelper.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    private static void validateExpiryTime(final Policy policy)
+    {
+        final String expiryTime = policy.getExpiryTime();
+        try {
+            DateHelper.validate(expiryTime);
+        } catch (final BadRequestException e) {
+            LOG.error("invalid expiry_time {}", expiryTime);
+        }
```
dermot-hardy: Is it correct that this is just logging here?  What is the intention?

dermot-hardy (changes requested):

job-service/src/main/java/com/hpe/caf/services/job/utilities/ExpirationPolicyHelper.java
xbreizh: We should rethrow the exception, done in the new code.
I remodelled so we check and pass the expiration policy earlier, reaching an exception earlier if any issue.
Also, moved the date validation from the helper onto the PolicyBuilder where it makes more sense I believe.

job-service-db/src/main/resources/procedures/applyJobExpirationPolicy.sql
```diff
@@ -0,0 +1,94 @@
...
+CREATE OR REPLACE FUNCTION apply_job_expiration_policy(
+)
+RETURNS VOID
+LANGUAGE plpgsql
+VOLATILE
+AS
+$$
+BEGIN
+    -- create inner function that deletes or expires the job according to its status policy
+    CREATE OR REPLACE FUNCTION delete_or_expire_job(p_id varchar, j_id varchar, j_operation EXPIRATION_OPERATION)
```
dermot-hardy: This should not be being created on the fly.

job-service-db/src/main/resources/procedures/applyJobExpirationPolicy.sql
```diff
@@ -0,0 +1,94 @@
...
+                        WHEN jep.expiration_time IS NOT NULL AND LEFT(jep.expiration_time, 1) = 'l' THEN ((
+                                j.last_update_date + split_part(jep.expiration_time, '+', 2)::interval)::timestamp)
+                            <= now() AT TIME ZONE 'UTC'
+                        WHEN jep.expiration_time IS NOT NULL AND LEFT(jep.expiration_time, 1) = 'c'
+                            THEN ((j.create_date + split_part(jep.expiration_time, '+', 2)::interval)::timestamp)
+                            <= now() AT TIME ZONE 'UTC'
+                        ELSE (jep.expiration_time::timestamp) <= now() AT TIME ZONE 'UTC'
+                        END           expired
+             FROM job j
+                      FULL OUTER JOIN job_expiration_policy jep
```
dermot-hardy: This appears to be calculating expiry dates for statuses other than the current one.

job-service-db/src/main/resources/procedures/applyJobExpirationPolicy.sql
```diff
@@ -0,0 +1,94 @@
...
+AS
+$$
+BEGIN
+    -- create inner function that deletes or expires the job according to its status policy
+    CREATE OR REPLACE FUNCTION delete_or_expire_job(p_id varchar, j_id varchar, j_operation EXPIRATION_OPERATION)
+    RETURNS VOID AS
+    $delete_or_expire_job$
+    BEGIN
+        IF j_operation = 'Expire' THEN
+            UPDATE job j
```
dermot-hardy: Create an `internal_expire_job` similar to `cancel_job` and call that from here.

job-service-db/src/main/resources/procedures/applyJobExpirationPolicy.sql
```diff
@@ -0,0 +1,94 @@
...
+-- limitations under the License.
+--
+
+/*
+ *  Name: apply_job_expiration_policy
+ *
+ *  Description:
+ *  Expires or deletes a job based on its expiration policy.
+ */
+CREATE OR REPLACE FUNCTION apply_job_expiration_policy(
```
dermot-hardy: Go with a procedure for this.

job-service/src/main/java/com/hpe/caf/services/job/api/DatabaseHelper.java
```diff
@@ -239,6 +251,52 @@ public Job getJob(final String partitionId, String jobId) throws Exception {
         return job;
     }
 
+    private ExpirationPolicy retrieveExpirationPolicy(final ExpirationPolicy expirationPolicy, final ResultSet rs) throws SQLException {
```
dermot-hardy: Reconsider as discussed.

dermot-hardy (changes requested):

job-service-db/src/main/resources/procedures/applyJobExpirationPolicy.sql
xbreizh: moved as external function

job-service-db/src/main/resources/procedures/applyJobExpirationPolicy.sql
xbreizh: Created expire_job and delete_or_expire_job

job-service-db/src/main/resources/procedures/applyJobExpirationPolicy.sql
xbreizh: converted into procedure

job-service/src/main/java/com/hpe/caf/services/job/api/DatabaseHelper.java
xbreizh: Done

xbreizh:
@dermot-hardy There was a bug on DatabaseHelper.getJobs() that made the function return null policies. Now working.
I also added support for flagging a System default policy so we can easily track it down. I set it only for variable expiration_time (ex: none, lastUpdateTime and createTime related) as I believe it does not make much sense to apply a fix date as system default.
I tried refactoring apply_job_expiration_policy() but failing to do it so far. I ll have a fresh look again tomorrow.
Please let me know if you think of anything else. Thx!

xbreizh:
@dermot-hardy I worked some more on the apply_job_expiration_policy()
plan: https://explain.depesz.com/s/p7tA  /  https://explain.dalibo.com/plan/nkM

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
```diff
@@ -415,6 +417,53 @@ definitions:
         example:
           "tag:4": "4"
           owner: "bob"
+  expirationPolicy:
+    type: object
+    properties:
+      "Active":
+        $ref: "#/definitions/policy"
```
dermot-hardy: Let's change this to `expirablePolicy`.

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
```diff
@@ -415,6 +417,53 @@ definitions:
...
+        $ref: "#/definitions/deletePolicy"
+      "Cancelled":
+        $ref: "#/definitions/deletePolicy"
+      "Waiting":
+        $ref: "#/definitions/policy"
+      "Paused":
+        $ref: "#/definitions/policy"
+      "Expired":
+        $ref: "#/definitions/deletePolicy"
+      "Default":
```
dermot-hardy: Scrap default policy given that it is no longer applicable to half the statuses.

job-service/src/main/java/com/hpe/caf/services/job/utilities/PolicyBuilder.java
```diff
@@ -0,0 +1,140 @@
...
+import com.hpe.caf.services.job.api.generated.model.ExpirationPolicy;
+import com.hpe.caf.services.job.api.generated.model.NewJob;
+import com.hpe.caf.services.job.api.generated.model.Policy;
+import com.hpe.caf.services.job.exceptions.BadRequestException;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+
+public final class PolicyBuilder
+{
+    private static final String SYSTEM_FLAG = "+System";
```
dermot-hardy: Policies flagged with this were not supplied by the user so they should not be returned to the user (unless with a flag that indicates that they have been inherited/implicitly applied).

job-service/src/main/java/com/hpe/caf/services/job/utilities/PolicyBuilder.java
```diff
@@ -0,0 +1,140 @@
...
+    private static void defineDeletePolicy(
+        final Supplier<DeletePolicy> policySupplier,
+        final Consumer<DeletePolicy> policyConsumer
+    ) throws BadRequestException
+    {
+        if (null == policySupplier.get()) {
+            final DeletePolicy policy = defineDefaultDeletePolicy();
+            policyConsumer.accept(policy);
+        } else {
+            policySupplier.get().setOperation(OperationEnum.DELETE);
```
dermot-hardy: Is an equivalent required in the defineExpirablePolicy() function?

job-service/src/main/java/com/hpe/caf/services/job/utilities/PolicyBuilder.java
```diff
@@ -0,0 +1,140 @@
...
+
+    private static Policy clonePolicy(final Policy policy)
+    {
+        final Policy newPolicy = new Policy();
+        newPolicy.setOperation(policy.getOperation());
+        newPolicy.setExpiryTime(policy.getExpiryTime());
+        return newPolicy;
+    }
+
+    private static DeletePolicy defineDefaultDeletePolicy()
```
dermot-hardy: Do this the same way as the other.

job-service-db/src/main/resources/procedures/expireJob.sql
```diff
@@ -0,0 +1,58 @@
...
+    WHERE partition_id = in_partition_id
+      AND job_id = in_job_id
+      AND status != 'Expired';
+
+    -- Drop any task tables relating to the job
+    PERFORM internal_drop_task_tables(in_partition_id, in_job_id);
+
+    -- Removes all related subtasks from completed_subtask_report table
+    PERFORM internal_cleanup_completed_subtask_report(in_partition_id, in_job_id);
+END
```
dermot-hardy, line 58: Should we be calling `internal_process_failed_dependent_jobs` when jobs expire?  I.e. what happens when a job that has dependent jobs waiting for it to complete expires?  Should they inherit the expiry?  Or should they be marked as failed?  (I'm inclined towards the latter - i.e. that they should fail).

dermot-hardy (changes requested):

job-service-db/src/main/resources/procedures/expireJob.sql
xbreizh: I am adding internal_process_failed_dependent_jobs on deleteOrExpireJob() as I think we should fail the related jobs in both cases.

job-service/src/main/java/com/hpe/caf/services/job/utilities/PolicyBuilder.java
xbreizh: Done

job-service/src/main/java/com/hpe/caf/services/job/utilities/PolicyBuilder.java
xbreizh: Added delete as default in the contract for deletePolicy (was already for the expirablePolicy)

job-service/src/main/java/com/hpe/caf/services/job/utilities/PolicyBuilder.java
xbreizh: Added policer

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
xbreizh: Done

job-service-contract/src/main/resources/com/hpe/caf/services/job/swagger.yaml
xbreizh: Done

xbreizh:
@dermot-hardy I refactored the [apply_job_expiration_policy](https://github.com/JobService/job-service/pull/235/commits/21b8fc0d84e8889e99755f9400609767f58658d5) by removing a layer (join table) and simplifying the selection.
Query plan [before](https://explain.depesz.com/s/LK97) 
Query plan [after](https://explain.depesz.com/s/vyUR) 
I hard-coded the value passed for [in_propagate_failures](https://github.com/JobService/job-service/pull/235/commits/b41dad3ce5b4aaf570da33f4f114794dbe051c04) as discussed, as it's being definitely removed with another JIRA
Last, I fixed a [failing test](https://github.com/JobService/job-service/pull/235/commits/7b53b684b7c99c1e32d818b183a3c7ff77d06915)

xbreizh:
New explain analyse
[Source 1](https://explain.dalibo.com/plan/cPA) 
[Source 2](https://explain.depesz.com/s/fnCe)

xbreizh:
New Explain Analyse
[Source 1](https://explain.depesz.com/s/WjSC) 
[Source 2](https://explain.dalibo.com/plan/oSh)

PR #237, opened 2021-03-12
SCMOD-12730: Updates to support pausing jobs

rorytorneymf:

- `statusCheckUrl` now points to the `status` endpoint rather than the `isActive` endpoint
- Worker framework updated to the latest version (via the caf-dependency-management-bom)
- `CAF_STATUS_CHECK_TIME` env var renamed to `CAF_STATUS_CHECK_INTERVAL_SECONDS`
- Removed `statusCheckTime` field from `TrackingInfo`, which was the time the next status check should be performed, and replaced it with 2 new fields: `lastStatusCheckTime`, which is the last time the status check was performed, and `statusCheckIntervalMillis`, which is the interval in milliseconds to wait between status checks. Passing these 2 new fields between workers instead of just the time the next status check should be performed will give the workers more flexibility in deciding how often to check the status of the task.


pom.xml
```diff
@@ -83,7 +83,7 @@
             <dependency>
                 <groupId>com.github.cafapi</groupId>
                 <artifactId>caf-dependency-management-bom</artifactId>
-                <version>2.1.0-SNAPSHOT</version>
+                <version>3.0.0-SCMOD-12730-SNAPSHOT</version>
```
rorytorneymf: Update before merging.

release-notes-4.0.0.md
```diff
@@ -15,6 +15,11 @@ ${version-number}
   - The names used for the `configurationProperties` now have a direct mapping to the environment variables used to populate them. For
   example, if a configuration property is named `TASK_PIPE` in a job type yaml definition, then the value for that configuration property
  is expected to be available in an environment variable named `TASK_PIPE`.
+- SCMOD-12730: The `statusCheckUrl` now points to the `status` endpoint instead of the `isActive` endpoint.
```
dermot-hardy: You can include that detail if you want but make the headline "SCMOD-12730: Workers built using framework versions prior to 5.0.0 are no longer supported."

dermot-hardy (approved):

rorytorneymf:
@dermot-hardy Re-requested review due to changes to the status check tracking info:

https://github.com/JobService/job-service/pull/237/commits/430351f5fb8e0d2a6fd9b114e485b0f461af485d

dermot-hardy (approved):
Yes go ahead renaming the environment variable - just make sure that you list it among the breaking changes as it will be important for consumers to know.

Don't forget to remove imports that you're no longer using:
image

PR #241, opened 2021-04-06
SCMOD-12730: Add stowed_task table 

rorytorneymf:

### DIFF
--- a/job-service-db/src/main/resources/v4.0.0/changelog.xml
+++ b/job-service-db/src/main/resources/v4.0.0/changelog.xml
@@ -68,7 +68,7 @@
             <column name="tracking_info_job_task_id" type="VARCHAR(255)">
                 <constraints nullable="false"/>
             </column>
-            <column name="tracking_info_last_status_check_time" type="TIMESTAMP">
+            <column name="tracking_info_last_status_check_time" type="BIGINT">
                 <constraints nullable="true"/>
             </column>
             <column name="tracking_info_status_check_interval_millis" type="BIGINT">
--- a/job-service-db/src/main/resources/v4.0.0/changelog.xml
+++ b/job-service-db/src/main/resources/v4.0.0/changelog.xml
@@ -35,9 +35,6 @@
     </changeSet>
     <changeSet id="create_stowed_task_table" author="Gamma Team">
         <createTable tableName="stowed_task">
-            <column name="id" type="SERIAL">
-                <constraints primaryKey="true" nullable="false" />
-            </column>
             <column name="partition_id" type="VARCHAR(40)">
                 <constraints nullable="false" />
             </column>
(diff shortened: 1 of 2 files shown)


release-notes-4.0.0.md
```diff
@@ -4,8 +4,9 @@
 ${version-number}
 
 #### New Features
-- SCMOD-12707: Added support for pausing and resuming jobs.
+- SCMOD-12730: Added support for pausing and resuming jobs.
   - Three new endpoints have been added to pause, resume, and get the status of a job.
+  - A new `stowed_task` table has been added to the database to stow tasks belonging to a paused job.
```
dermot-hardy: Drop this.  Too internal for release notes.  The "Added support for pausing and resuming jobs." covers it.

job-service-db/src/main/resources/v4.0.0/changelog.xml
```diff
@@ -33,4 +33,54 @@
             <column name="table_name" />
         </createIndex>
     </changeSet>
+    <changeSet id="create_stowed_task_table" author="Gamma Team">
+        <createTable tableName="stowed_task">
+            <column name="id" type="SERIAL">
+                <constraints primaryKey="true" nullable="false" />
+            </column>
```
dermot-hardy: What do we need this column for?

rorytorneymf: Since we had no natural primary key for this table (as we can have many tasks with the same partitionId and jobId), I added this as the general consensus seems to be that each table should have a primary key:

https://stackoverflow.com/questions/840162/should-each-and-every-table-have-a-primary-key

As a nice side effect, I was using the id of the row in error logging when there was some problem that occurred, in order to find out exactly which row caused the issue.

But it's not strictly necessary, and I can remove it if you think we don't need it?

dermot-hardy: The logging argument is appealing but on reflection I don't think it's sufficient to justify the column.  The advice that tables should generally have primary keys is decent advice, but that doesn't mean that you should invent an artificial one.  The natural primary key for this table would be (partition_id, task_id).  Unfortunately at the minute task_id is buried in the tracking_info byte array.  There is a good argument for extracting out the tracking info fields and storing them in their own right, and then just reconstructing the tracking info from them on resumption.  I think you should do that, but even then we may want to think carefully about whether we want to add the (partition_id, task_id) as a primary key, because if we do then we'll have to deal with the possibility of the paused message being processed twice.  At the minute it would just add two rows - not great but we can live with it.  If (partition_id, task_id) was a primary key then the database wouldn't allow that, and so we'd have to deal with that situation (which is easy for single inserts, but because of potential mixing with legitimate rows more complex in the bulk case).

rorytorneymf: Thanks, I've split out tracking info into its own columns now. When you say 'taskId' is that the same thing as `TrackingInfo.getJobTaskId`?
```
Values of jobTaskId should be the job id followed by period-separated subtask elements. For example J5.1.2 where the job id is
     * "J5".
```

dermot-hardy: Yes

dermot-hardy (approved):

PR #242, opened 2021-04-09
SCMOD-12972: Increase task_id to 70

xbreizh:
https://portal.digitalsafe.net/browse/SCMOD-12972


dermot-hardy (commented):
I don't understand why it is just these functions that have this issue.  If we need to increase the width of `task_id` for `report_complete` then why not also for `report_failure`, `report_progress`, etc.

xbreizh:
> report_progress



> I don't understand why it is just these functions that have this issue. If we need to increase the width of `task_id` for `report_complete` then why not also for `report_failure`, `report_progress`, etc.

I have changed the missing ones.

dermot-hardy (approved):

PR #244, opened 2021-04-15
SCMOD-12730: Send msg to resume queue when job resumed

rorytorneymf:

When a job is resumed, send a message to the `CAF_JOB_SERVICE_RESUME_JOB_QUEUE`. 

The message is a `DocumentWorkerTask` and looks something like this:

```
{
  "version": 3,
  "taskId": "89dfb12d-5de0-4b40-85ec-835e33ea447d",
  "taskClassifier": "DocumentWorkerTask",
  "taskApiVersion": 1,
  "taskData": "eyJjdXN0b21EYXRhIjp7ImpvYklkIjoiSjE2MTg0OTkxOTcyNjQtOTE2IiwicGFydGl0aW9uSWQiOiIwN2I2NjM2NS1lOTlmLTQ1OWEtYjM2NC1iMWFmZmJkYzMzYzYifX0=",
  "taskStatus": "NEW_TASK",
  "context": 1,
  "to": "worker-taskunstowing-in",
  "tracking": null,
  "sourceInfo": null,
  "priority": null,
  "correlationId": "1"
}
```

Decoded `taskData`:

```
{
  "customData": {
    "jobId": "J1618499197264-916",
    "partitionId": "07b66365-e99f-459a-b364-b1affbdc33c6"
  }
}
```

A new worker (probably named Task Unstowing Worker) will consume from this queue, and use the `jobId` and `partitionId` to read previously stowed task messages from the database, and dispatch them to the worker they are supposed to go to.


job-service/src/main/java/com/hpe/caf/services/job/api/JobsResume.java
```diff
@@ -56,14 +72,78 @@ public static void resumeJob(final String partitionId, String jobId) throws Exce
...
+            // Store the job status before resuming the job.
+            final Job.StatusEnum jobStatusBeforeResuming = databaseHelper.getJobStatus(partitionId, jobId);
+
             //  Resume the specified job.
-            LOG.debug("resume: Resuming the job...");
+            LOG.debug("resume: Updating the status of the job in the database...");
             databaseHelper.resumeJob(partitionId, jobId);
 
+            // Send a resume job message only if the job was paused (we allow resuming an active job, which is why we need to check this).
+            if (jobStatusBeforeResuming == Job.StatusEnum.PAUSED) {
```
rorytorneymf: There is a race condition here that can between lines 76 and 83, if another request came in that caused the status to change. 

I think it is ok if it did happen, as the new Task Unstowing worker would get a resume job message, not be able to find any matching tasks in the db (because the job is active/complete whatever), and can either just ignore the message or log a warning "Received a job resume message but unable to find any tasks to be resumed"

deploy/src/main/resources-filtered/jobservice-deployment.yaml
```diff
@@ -144,6 +144,11 @@ spec:
               configMapKeyRef:
                 name: jobservice-config
                 key: CAF_WEBSERVICE_URL
+          - name: CAF_JOB_SERVICE_RESUME_JOB_QUEUE
```
rorytorneymf, line 147: Not sure about the name `CAF_JOB_SERVICE_RESUME_JOB_QUEUE`. 

We have `CAF_WORKER_PAUSED_QUEUE` so maybe `CAF_JOB_SERVICE_RESUMED_QUEUE` might be a better name so that the two queues have a similar naming pattern, e.g. `<WORKER>_<ACTION>_QUEUE`

`CAF_WORKER_PAUSED_QUEUE`
`CAF_JOB_SERVICE_RESUMED_QUEUE`

What do you think?

dermot-hardy: Put "request" into the queue name: `CAF_JOB_SERVICE_RESUME_REQUEST_QUEUE`.

dermot-hardy: Or your original suggestion is fine: `CAF_JOB_SERVICE_RESUME_JOB_QUEUE`

rorytorneymf: Thanks, I'll leave it as is then: `CAF_JOB_SERVICE_RESUME_JOB_QUEUE`

job-service/src/main/java/com/hpe/caf/services/job/api/JobsResume.java
```diff
@@ -15,23 +15,39 @@
...
      *
-     * @param   partitionId the partition containing the id of the job to resume
-     * @param   jobId       the id of the job to resume
-     * @throws  Exception   bad request or database exceptions thrown
+     * @param partitionId the partition containing the id of the job to resume
+     * @param jobId the id of the job to resume
+     * @throws Exception bad request, database or queue exceptions thrown
      */
-    public static void resumeJob(final String partitionId, String jobId) throws Exception {
+    public static void resumeJob(final String partitionId, String jobId) throws Exception
```
dermot-hardy:
```suggestion
    public static void resumeJob(final String partitionId, final String jobId) throws Exception
```

job-service/src/main/java/com/hpe/caf/services/job/api/JobsResume.java
```diff
@@ -15,23 +15,39 @@
...
+    {
 
         try {
             LOG.debug("resume: Starting...");
```
dermot-hardy: You don't really need `resume: ` in all these log messages since `JobsResume` should be automatically in the log lines anyway.

job-service/src/main/java/com/hpe/caf/services/job/api/JobsResume.java
dermot-hardy: I think it would be safer to end out the message before updating the database status.  If the Job Service crashed at this point in the code the database status would be updated but the message wouldn't be sent.  Furthermore calling `/resume` again would do nothing because of this check.  (I suppose they could get it going again by re-pausing and then re-resuming but I don't think we should expect that level of smarts from the calling code.)

job-service/src/main/java/com/hpe/caf/services/job/api/JobsResume.java
```diff
@@ -56,14 +72,78 @@ public static void resumeJob(final String partitionId, String jobId) throws Exce
...
+        }
+    }
+
+    private static WorkerAction createResumeJobWorkerAction(
+        final String partitionId,
+        final String jobId,
+        final String resumeJobQueue) throws CodecException
+    {
+        final WorkerAction workerAction = new WorkerAction();
+        workerAction.setTaskClassifier("DocumentWorkerTask");
```
dermot-hardy:
```suggestion
        workerAction.setTaskClassifier(DocumentWorkerConstants.DOCUMENT_TASK_NAME);
```

job-service/src/main/java/com/hpe/caf/services/job/api/JobsResume.java
```diff
@@ -56,14 +72,78 @@ public static void resumeJob(final String partitionId, String jobId) throws Exce
...
+    private static Map<String, byte[]> createResumeJobTaskData(
+        final String partitionId,
+        final String jobId) throws CodecException
+    {
+        final DocumentWorkerDocumentTask documentWorkerDocumentTask = new DocumentWorkerDocumentTask();
+        final Map<String, String> customData = new HashMap<>();
+        customData.put("partitionId", partitionId);
+        customData.put("jobId", jobId);
+        documentWorkerDocumentTask.customData = customData;
+        return new ObjectMapper().convertValue(documentWorkerDocumentTask, Map.class);
```
dermot-hardy: I'm not sure I understand this line.  When I seen it I checked the build log to see if it was generating any warnings and I see this:
```
/C:/Source/JobService/job-service/job-service/src/main/java/com/hpe/caf/services/job/api/JobsResume.java: C:\Source\JobService\job-service\job-service\src\main\java\com\hpe\caf\services\job\api\JobsResume.java uses unchecked or unsafe operations.
/C:/Source/JobService/job-service/job-service/src/main/java/com/hpe/caf/services/job/api/JobsResume.java: Recompile with -Xlint:unchecked for details.
```

If it's right you should probably keep `ObjectMapper` as a class variable so that you can re-use it.

dermot-hardy (changes requested):

job-service/src/main/java/com/hpe/caf/services/job/api/JobsResume.java
dermot-hardy: I'm still not comfortable with this line:

image

deploy/src/main/resources-filtered/production-swarm/docker-stack.yml
```diff
@@ -11,6 +11,7 @@ services:
       CAF_STATUS_CHECK_INTERVAL_SECONDS: 5
       CAF_TRACKING_PIPE: jobtracking-in
       CAF_WEBSERVICE_URL: "http://${JOB_SERVICE_DOCKER_HOST}:${JOB_SERVICE_PORT}/job-service/v1"
+      CAF_JOB_SERVICE_RESUME_JOB_QUEUE: "${CAF_JOB_SERVICE_RESUME_JOB_QUEUE}"
```
dermot-hardy, line 14: xxx

job-service/src/main/java/com/hpe/caf/services/job/api/JobsResume.java
dermot-hardy: I'm still not comfortable with this line:

image

job-service/src/main/java/com/hpe/caf/services/job/api/JobsResume.java
```diff
@@ -15,55 +15,134 @@
...
+    private static Map<String, byte[]> createResumeJobTaskData(
+        final String partitionId,
+        final String jobId) throws CodecException
+    {
+        final DocumentWorkerDocumentTask documentWorkerDocumentTask = new DocumentWorkerDocumentTask();
+        final Map<String, String> customData = new HashMap<>();
+        customData.put("partitionId", partitionId);
+        customData.put("jobId", jobId);
+        documentWorkerDocumentTask.customData = customData;
+        return OBJECT_MAPPER.convertValue(documentWorkerDocumentTask, Map.class);
```
dermot-hardy: I'm still not comfortable with this line:

image

dermot-hardy (approved):

job-service/src/main/java/com/hpe/caf/services/job/api/JobsResume.java
rorytorneymf: I tried using a `TypeReference` here instead of a `Map`:

```
import com.fasterxml.jackson.core.type.TypeReference;
...
private static final TypeReference<Map<String, byte[]>> TASK_DATA_TYPE_REFERENCE
        = new TypeReference<Map<String, byte[]>>()
    {
    };
...
return OBJECT_MAPPER.convertValue(documentWorkerDocumentTask, TASK_DATA_TYPE_REFERENCE);
`

but seeing some errors, I'll have to investigate more:

java.lang.Exception: Failed to send a message to the testResumeJobQueue queue.
	at com.hpe.caf.services.job.api.JobsResumeTest.testResumeJob_Success(JobsResumeTest.java:86)
Caused by: java.lang.IllegalArgumentException: 
Cannot deserialize instance of `[B` out of START_OBJECT token
 at [Source: UNKNOWN; line: -1, column: -1] (through reference chain: java.util.LinkedHashMap["customData"])
	at com.hpe.caf.services.job.api.JobsResumeTest.testResumeJob_Success(JobsResumeTest.java:86)

rorytorneymf: I don't get those warnings in my IDE (I use netbeans), but I've changed this so that I'm using the object mapper to write to bytes, then as a UTF_8 string, the same way the`DatabasePoller` does it:

https://github.com/JobService/job-service/blob/develop/job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/DatabasePoller.java#L72

workerAction.setTaskData(new String(jtd.getTaskData(), StandardCharsets.UTF_8));

dermot-hardy: > I don't get those warnings in my IDE

Your branch would need to be using the latest `caf-common-parent`, that includes the [a5da1ff](https://github.com/CAFapi/caf-common-parent/commit/a5da1ffe44ff3754a778473b3fd0dd27408ea803) commit, to see them.

PR #245, opened 2021-04-22
SCMOD-12730: Add index + drop context column

rorytorneymf:

### DIFF
--- a/job-service-db/src/main/resources/v4.0.0/changelog.xml
+++ b/job-service-db/src/main/resources/v4.0.0/changelog.xml
@@ -101,7 +101,16 @@
                                  referencedTableName="job"
                                  referencedColumnNames="partition_id, job_id" />
     </changeSet>
-    <changeSet id="add_task_stowing_index" author="Gamma Team">
+    <changeSet id="drop_task_stowing_idx_partition_id_and_job_id" author="Gamma Team">
+        <preConditions onFail="CONTINUE">
+            <indexExists schemaName="public"
+                         tableName="stowed_task"
+                         indexName="idx_partition_id_and_job_id"
+                         columnNames="partition_id, job_id"/>
+        </preConditions>
+        <dropIndex indexName="idx_partition_id_and_job_id" tableName="stowed_task"/>
+    </changeSet>
+    <changeSet id="create_task_stowing_idx_partition_id_and_job_id_and_tracking_info_job_task_id" author="Gamma Team">
         <preConditions onFail="CONTINUE">
             <not>
                 <indexExists schemaName="public"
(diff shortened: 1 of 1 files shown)


dermot-hardy (changes requested):
I don't see the need to retain the `idx_partition_id_and_job_id` index given this one being added.

rorytorneymf:
> I don't see the need to retain the `idx_partition_id_and_job_id` index given this one being added.

I think we still need this existing index to fetch **all** the stowed tasks for a job:

https://github.com/JobService/worker-taskunstowing/blob/SCMOD-12730/worker-taskunstowing/src/main/java/com/github/jobservice/workers/taskunstowing/TaskUnstowingWorker.java#L98

```
stowedTaskRows = databaseClient.getStowedTasks(partitionId, jobId);
```

This additional index would then be used to delete a specific stowed task, after it had been unstowed:

https://github.com/JobService/worker-taskunstowing/blob/SCMOD-12730/worker-taskunstowing/src/main/java/com/github/jobservice/workers/taskunstowing/TaskUnstowingWorker.java#L129

```
databaseClient.deleteStowedTask(partitionId, jobId, taskMessage.getTracking().getJobTaskId());
```

dermot-hardy:
If you read up on the way PostgreSQL B-tree indexes work you should see why this change makes the `idx_partition_id_and_job_id` essentially redundant.

rorytorneymf:
> If you read up on the way PostgreSQL B-tree indexes work you should see why this change makes the `idx_partition_id_and_job_id` essentially redundant.

Thanks, I've dropped the old 2-column index now, and also dropped the context column as well as discussed.

job-service-db/src/main/resources/v4.0.0/changelog.xml
```diff
@@ -101,4 +101,38 @@
...
+    <changeSet id="drop_task_stowing_idx_partition_id_and_job_id" author="Gamma Team">
+        <preConditions onFail="CONTINUE">
+            <indexExists schemaName="public"
+                         tableName="stowed_task"
+                         indexName="idx_partition_id_and_job_id"
+                         columnNames="partition_id, job_id"/>
+        </preConditions>
+        <dropIndex indexName="idx_partition_id_and_job_id" tableName="stowed_task"/>
+    </changeSet>
+    <changeSet id="create_task_stowing_idx_partition_id_and_job_id_and_tracking_info_job_task_id" author="Gamma Team">
```
rorytorneymf, line 113: Should I also add another changeSet to make this the primary key?

```
<changeSet id="add_task_stowing_primary_key" author="Gamma Team">
    <addPrimaryKey schemaName="public" tableName="stowedTask" columnNames="partition_id, job_id, tracking_info_job_task_id" />
</changeSet>
```

dermot-hardy: That would have further implications for the calling code so you'd be better leaving that for another Jira.

PR #248, opened 2021-05-04
SCMOD-13413: Fixed server unavailable feedback

xbreizh:
https://portal.digitalsafe.net/browse/SCMOD-13413


job-service/src/main/java/com/hpe/caf/services/job/api/ApiExceptionMapper.java
```diff
@@ -40,14 +47,22 @@
      */
     @Override
     public Response toResponse(Exception exception) {
+        LOG.debug("exception found {}", exception.getClass());
```
rorytorneymf: Maybe get rid of this debug log, I assume it was just for testing purposes?

job-service/src/main/java/com/hpe/caf/services/job/api/ApiExceptionMapper.java
```diff
@@ -40,14 +47,22 @@
...
         } else if (exception instanceof NotFoundException) {
             httpStatus = Response.Status.NOT_FOUND;
         } else if (exception instanceof ForbiddenException) {
             httpStatus = Response.Status.FORBIDDEN;
-        } else if (exception instanceof ServiceUnavailableException) {
+        } else if (
+            exception instanceof ServiceUnavailableException ||
+            exception instanceof AlreadyClosedException ||
+            exception instanceof TimeoutException ||
+            (exception instanceof IOException )
```
rorytorneymf: Seems to be an extra ( bracket here

job-service/src/main/java/com/hpe/caf/services/job/api/JobsPut.java
```diff
@@ -224,7 +225,7 @@ public static String createOrUpdateJob(final String partitionId, String jobId, N
                 databaseHelper.reportFailure(partitionId, jobId, mapper.writeValueAsString(f));
 
                 //  Throw error message to user.
-                throw new Exception("Failed to add task data to the queue", ex);
+                throw new ServiceUnavailableException("Failed to add task data to the queue", ex);
```
rorytorneymf, line 228: Should this be a 503? I think this usually happens when we forget to configure the job service with some queue ENV var it needs

job-service/src/main/java/com/hpe/caf/services/job/api/ApiExceptionMapper.java
xbreizh: Removed

job-service/src/main/java/com/hpe/caf/services/job/api/JobsPut.java
xbreizh: I went with 503 as we are catching only `IOException | TimeoutException ` at that level. 
For any missing or invalid queue data, I understand that a BadRequestException (400) is thrown 

         //  Make sure the target queue name is not empty. Null targetPipe is valid
            if ("".equals(jobTask.getTargetPipe())) {
                LOG.error("createOrUpdateJob: Error - '{}'", ERR_MSG_TARGET_QUEUE_NOT_SPECIFIED);
                throw new BadRequestException(ERR_MSG_TARGET_QUEUE_NOT_SPECIFIED);
            }

            //  Make sure the task queue name has been provided.
            if (!ApiServiceUtil.isNotNullOrEmpty(jobTask.getTaskPipe())) {
                LOG.error("createOrUpdateJob: Error - '{}'", ERR_MSG_TASK_QUEUE_NOT_SPECIFIED);
                throw new BadRequestException(ERR_MSG_TASK_QUEUE_NOT_SPECIFIED);
            }

job-service/src/main/java/com/hpe/caf/services/job/api/ApiExceptionMapper.java
xbreizh: Removed

rorytorneymf (approved):

job-service/src/main/java/com/hpe/caf/services/job/api/ApiExceptionMapper.java
```diff
@@ -41,13 +44,20 @@
...
         } else if (exception instanceof NotFoundException) {
             httpStatus = Response.Status.NOT_FOUND;
         } else if (exception instanceof ForbiddenException) {
             httpStatus = Response.Status.FORBIDDEN;
-        } else if (exception instanceof ServiceUnavailableException) {
+        } else if (
+            exception instanceof ServiceUnavailableException ||
+            exception instanceof AlreadyClosedException ||
+            exception instanceof TimeoutException ||
+            exception instanceof IOException
```
dermot-hardy, line 60: How come you have to add all these other exceptions when you're already updated `JobsPut.java` to throw a `ServiceUnavailableException`?

job-service/src/main/java/com/hpe/caf/services/job/api/ApiExceptionMapper.java
```diff
@@ -41,13 +44,20 @@
     @Override
     public Response toResponse(Exception exception) {
         final Response.Status httpStatus;
-        if (exception instanceof BadRequestException) {
+        if (exception instanceof BadRequestException ||
+            exception instanceof UnrecognizedPropertyException
```
dermot-hardy, line 49: Why are we adding this?

job-service/src/main/java/com/hpe/caf/services/job/api/ApiExceptionMapper.java
```diff
@@ -15,23 +15,26 @@
...
+import java.io.IOException;
+import java.util.concurrent.TimeoutException;
 
 /**
  * The ApiExceptionMapper class maps exceptions thrown by the audit management api
  * to response http status codes.
  */
 @Provider
 public final class ApiExceptionMapper implements ExceptionMapper<Exception> {
-
```
dermot-hardy: Unnecessary whitespace change.

job-service/src/main/java/com/hpe/caf/services/job/api/ApiExceptionMapper.java
xbreizh: I added those as in createOrUpdateJob(), we contact the database several times before reaching [the part where we intent to create a job](https://github.com/JobService/job-service/blob/bea9a358ea4767e812d054a0f130d67ee2f998c1/job-service/src/main/java/com/hpe/caf/services/job/api/JobsPut.java#L208)
So, I just wanted to cover the different disconnection exceptions. From my understanding, the idea of that 
 ApiExceptionMapper is to categorize all exceptions likely to return a Service Unavailable code.

job-service/src/main/java/com/hpe/caf/services/job/api/ApiExceptionMapper.java
xbreizh: If we pass an extra parameter for creating the job, we get an UnrecognizedPropertyException, which extends IOException. So, the way it's, it would be returning a 503. Adding that check allows to return a 400 which seems more appropriate.

job-service/src/main/java/com/hpe/caf/services/job/api/ApiExceptionMapper.java
xbreizh: Sorry, restored

dermot-hardy (approved):

PR #251, opened 2021-05-11
SCMOD-13414: Added logs details

xbreizh:
https://portal.digitalsafe.net/browse/SCMOD-13414

### DIFF
--- a/job-service/src/main/java/com/hpe/caf/services/job/api/JobsGetById.java
+++ b/job-service/src/main/java/com/hpe/caf/services/job/api/JobsGetById.java
@@ -65,7 +65,7 @@ public final class JobsGetById {
             job = databaseHelper.getJob(partitionId, jobId);
 
             LOG.debug("getJobById: Done.");
-            LOG.info("Job progression {}/{} status {}  progression {}", partitionId, jobId, job.getStatus(), job.getPercentageComplete());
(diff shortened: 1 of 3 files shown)


job-service/src/main/java/com/hpe/caf/services/job/api/JobsGetById.java
```diff
@@ -65,6 +65,7 @@ public static Job getJob(final String partitionId, String jobId) throws Exceptio
             job = databaseHelper.getJob(partitionId, jobId);
 
             LOG.debug("getJobById: Done.");
+            LOG.info("Job progression {}/{} status {}  progression {}", partitionId, jobId, job.getStatus(), job.getPercentageComplete());
```
dermot-hardy: Did you mean to write progression twice?

dermot-hardy (approved):

job-service/src/main/java/com/hpe/caf/services/job/api/JobsGetById.java
xbreizh: Yes, like "Job progression" as title for what follows, then "progression" as details. But if that's confusing, I can change

xbreizh: Changing for:
            LOG.info("Job progress {}/{} status {}  progression {}", partitionId, jobId, job.getStatus(), job.getPercentageComplete());

PR #252, opened 2021-05-28
SCMOD-13870: Added rabbit Exception handling

xbreizh:

### DIFF
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -193,7 +193,7 @@
                             <goal>wget</goal>
                         </goals>
                         <configuration>
-                            <retries>3</retries>
+                            <retries>7</retries>
                             <outputFileName>${project.build.finalName}.tar.gz</outputFileName>
                             <overwrite>true</overwrite>
                             <skipCache>true</skipCache>
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -193,6 +193,7 @@
                             <goal>wget</goal>
                         </goals>
                         <configuration>
+                            <retries>3</retries>
                             <outputFileName>${project.build.finalName}.tar.gz</outputFileName>
                             <overwrite>true</overwrite>
                             <skipCache>true</skipCache>
@@ -200,7 +201,6 @@
                         </configuration>
                     </execution>
                 </executions>
-                <retryFailedDeploymentCount>3</retryFailedDeploymentCount>
             </plugin>
 
             <!-- In this project, only include the project file itself when checking license headers -->
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -184,7 +184,7 @@
             <plugin>
                 <groupId>com.googlecode.maven-download-plugin</groupId>
                 <artifactId>download-maven-plugin</artifactId>
-                <version>1.6.3</version>
+                <version>1.3.0</version>
                 <executions>
                     <execution>
                         <id>download-target-file</id>
@@ -193,6 +193,7 @@
                             <goal>wget</goal>
                         </goals>
                         <configuration>
+                            <retries>10</retries>
                             <outputFileName>${project.build.finalName}.tar.gz</outputFileName>
                             <overwrite>true</overwrite>
                             <skipCache>true</skipCache>
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -197,7 +197,6 @@
                             <overwrite>true</overwrite>
                             <skipCache>true</skipCache>
                             <url>http://${docker.host.address}:${webServerPort}/docs-site.tar.gz</url>
-                            <retries>10</retries>
                         </configuration>
                     </execution>
                 </executions>
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>com.github.jobservice</groupId>
         <artifactId>job-service-aggregator</artifactId>
-        <version>4.1.0-SNAPSHOT</version>
+        <version>5.0.0-SNAPSHOT</version>
     </parent>
 
     <properties>
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -184,7 +184,7 @@
             <plugin>
                 <groupId>com.googlecode.maven-download-plugin</groupId>
                 <artifactId>download-maven-plugin</artifactId>
-                <version>1.3.0</version>
+                <version>1.6.3</version>
                 <executions>
                     <execution>
                         <id>download-target-file</id>
@@ -193,7 +193,6 @@
                             <goal>wget</goal>
                         </goals>
                         <configuration>
-                            <retries>7</retries>
                             <outputFileName>${project.build.finalName}.tar.gz</outputFileName>
                             <overwrite>true</overwrite>
                             <skipCache>true</skipCache>
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -184,7 +184,7 @@
             <plugin>
                 <groupId>com.googlecode.maven-download-plugin</groupId>
                 <artifactId>download-maven-plugin</artifactId>
-                <version>1.6.2</version>
+                <version>1.3.0</version>
                 <executions>
                     <execution>
                         <id>download-target-file</id>
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -184,7 +184,7 @@
             <plugin>
                 <groupId>com.googlecode.maven-download-plugin</groupId>
                 <artifactId>download-maven-plugin</artifactId>
-                <version>1.3.0</version>
+                <version>1.6.2</version>
                 <executions>
                     <execution>
                         <id>download-target-file</id>
@@ -193,11 +193,11 @@
                             <goal>wget</goal>
                         </goals>
                         <configuration>
-                            <retries>10</retries>
                             <outputFileName>${project.build.finalName}.tar.gz</outputFileName>
                             <overwrite>true</overwrite>
                             <skipCache>true</skipCache>
                             <url>http://${docker.host.address}:${webServerPort}/docs-site.tar.gz</url>
+                            <retries>10</retries>
                         </configuration>
                     </execution>
                 </executions>
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -200,6 +200,7 @@
                         </configuration>
                     </execution>
                 </executions>
+                <retryFailedDeploymentCount>3</retryFailedDeploymentCount>
             </plugin>
 
             <!-- In this project, only include the project file itself when checking license headers -->
--- a/job-service-acceptance-tests/pom.xml
+++ b/job-service-acceptance-tests/pom.xml
@@ -630,7 +630,6 @@
                             <alias>job-service-caller</alias>
                             <name>${dockerJobServiceOrg}job-service-caller${dockerProjectVersion}</name>
                         </image>
-    
                         <!-- Run the Job Service Scheduled Executor -->
                         <image>
                             <alias>job-service-scheduled-executor</alias>
--- a/job-service-acceptance-tests/pom.xml
+++ b/job-service-acceptance-tests/pom.xml
@@ -455,46 +455,6 @@
                                 </log>
                             </run>
                         </image>
-    
-                        <!-- Run the Job Service Scheduled Executor -->
-                        <image>
-                            <alias>job-service-scheduled-executor</alias>
-                            <name>${dockerJobServiceOrg}job-service-scheduled-executor${dockerProjectVersion}</name>
-                            <run>
-                                <ports>
-                                    <port>${scheduler-worker.adminport}:8081</port>
-                                </ports>
-                                <env>
-                                    <JOB_SERVICE_DATABASE_APPNAME>Job Service Scheduled Executor</JOB_SERVICE_DATABASE_APPNAME>
-                                    <JOB_SERVICE_DATABASE_URL>${job.database.url}</JOB_SERVICE_DATABASE_URL>
-                                    <JOB_SERVICE_DATABASE_USERNAME>${job.database.username}</JOB_SERVICE_DATABASE_USERNAME>
-                                    <JOB_SERVICE_DATABASE_PASSWORD>${job.database.password}</JOB_SERVICE_DATABASE_PASSWORD>
-                                    <CAF_TRACKING_PIPE>jobtrackingworker-test-input-1</CAF_TRACKING_PIPE>
-                                    <CAF_STATUS_CHECK_INTERVAL_SECONDS>1</CAF_STATUS_CHECK_INTERVAL_SECONDS>
-                                    <CAF_WEBSERVICE_URL>http://job-service:8080/job-service/v1</CAF_WEBSERVICE_URL>
-                                    <CAF_RABBITMQ_HOST>rabbitmq</CAF_RABBITMQ_HOST>
-                                    <CAF_RABBITMQ_PORT>5672</CAF_RABBITMQ_PORT>
-                                    <CAF_RABBITMQ_USERNAME>guest</CAF_RABBITMQ_USERNAME>
-                                    <CAF_RABBITMQ_PASSWORD>guest</CAF_RABBITMQ_PASSWORD>
-                                    <CAF_SCHEDULED_EXECUTOR_PERIOD>2</CAF_SCHEDULED_EXECUTOR_PERIOD>
-                                    <CAF_WORKER_INPUT_QUEUE>jobservicescheduler-in</CAF_WORKER_INPUT_QUEUE>
-                                </env>
-                                <links>
-                                    <link>rabbitmq</link>
-                                    <link>${job.database.host}</link>
-                                </links>
-                                <log>
-                                    <enabled>true</enabled>
-                                </log>
-                                <wait>
-                                    <http>
-                                        <url>http://${docker.host.address}:${scheduler-worker.adminport}/healthcheck</url>
-                                    </http>
-                                    <time>200000</time>
-                                    <shutdown>500</shutdown>
-                                </wait>
-                            </run>
-                        </image>
 
                         <!-- Run the Job Service -->
                         <image>
@@ -670,7 +630,46 @@
                             <alias>job-service-caller</alias>
                             <name>${dockerJobServiceOrg}job-service-caller${dockerProjectVersion}</name>
                         </image>
-
+    
+                        <!-- Run the Job Service Scheduled Executor -->
+                        <image>
+                            <alias>job-service-scheduled-executor</alias>
+                            <name>${dockerJobServiceOrg}job-service-scheduled-executor${dockerProjectVersion}</name>
+                            <run>
+                                <ports>
+                                    <port>${scheduler-worker.adminport}:8081</port>
+                                </ports>
+                                <env>
+                                    <JOB_SERVICE_DATABASE_APPNAME>Job Service Scheduled Executor</JOB_SERVICE_DATABASE_APPNAME>
+                                    <JOB_SERVICE_DATABASE_URL>${job.database.url}</JOB_SERVICE_DATABASE_URL>
+                                    <JOB_SERVICE_DATABASE_USERNAME>${job.database.username}</JOB_SERVICE_DATABASE_USERNAME>
+                                    <JOB_SERVICE_DATABASE_PASSWORD>${job.database.password}</JOB_SERVICE_DATABASE_PASSWORD>
+                                    <CAF_TRACKING_PIPE>jobtrackingworker-test-input-1</CAF_TRACKING_PIPE>
+                                    <CAF_STATUS_CHECK_INTERVAL_SECONDS>1</CAF_STATUS_CHECK_INTERVAL_SECONDS>
+                                    <CAF_WEBSERVICE_URL>http://job-service:8080/job-service/v1</CAF_WEBSERVICE_URL>
+                                    <CAF_RABBITMQ_HOST>rabbitmq</CAF_RABBITMQ_HOST>
+                                    <CAF_RABBITMQ_PORT>5672</CAF_RABBITMQ_PORT>
+                                    <CAF_RABBITMQ_USERNAME>guest</CAF_RABBITMQ_USERNAME>
+                                    <CAF_RABBITMQ_PASSWORD>guest</CAF_RABBITMQ_PASSWORD>
+                                    <CAF_SCHEDULED_EXECUTOR_PERIOD>2</CAF_SCHEDULED_EXECUTOR_PERIOD>
+                                    <CAF_WORKER_INPUT_QUEUE>jobservicescheduler-in</CAF_WORKER_INPUT_QUEUE>
+                                </env>
+                                <links>
+                                    <link>rabbitmq</link>
+                                    <link>${job.database.host}</link>
+                                </links>
+                                <log>
+                                    <enabled>true</enabled>
+                                </log>
+                                <wait>
+                                    <http>
+                                        <url>http://${docker.host.address}:${scheduler-worker.adminport}/healthcheck</url>
+                                    </http>
+                                    <time>200000</time>
+                                    <shutdown>500</shutdown>
+                                </wait>
+                            </run>
+                        </image>
                     </images>
                 </configuration>
             </plugin>
--- a/job-service-acceptance-tests/pom.xml
+++ b/job-service-acceptance-tests/pom.xml
@@ -455,6 +455,46 @@
                                 </log>
                             </run>
                         </image>
+    
+                        <!-- Run the Job Service Scheduled Executor -->
+                        <image>
+                            <alias>job-service-scheduled-executor</alias>
+                            <name>${dockerJobServiceOrg}job-service-scheduled-executor${dockerProjectVersion}</name>
+                            <run>
+                                <ports>
+                                    <port>${scheduler-worker.adminport}:8081</port>
+                                </ports>
+                                <env>
+                                    <JOB_SERVICE_DATABASE_APPNAME>Job Service Scheduled Executor</JOB_SERVICE_DATABASE_APPNAME>
+                                    <JOB_SERVICE_DATABASE_URL>${job.database.url}</JOB_SERVICE_DATABASE_URL>
+                                    <JOB_SERVICE_DATABASE_USERNAME>${job.database.username}</JOB_SERVICE_DATABASE_USERNAME>
+                                    <JOB_SERVICE_DATABASE_PASSWORD>${job.database.password}</JOB_SERVICE_DATABASE_PASSWORD>
+                                    <CAF_TRACKING_PIPE>jobtrackingworker-test-input-1</CAF_TRACKING_PIPE>
+                                    <CAF_STATUS_CHECK_INTERVAL_SECONDS>1</CAF_STATUS_CHECK_INTERVAL_SECONDS>
+                                    <CAF_WEBSERVICE_URL>http://job-service:8080/job-service/v1</CAF_WEBSERVICE_URL>
+                                    <CAF_RABBITMQ_HOST>rabbitmq</CAF_RABBITMQ_HOST>
+                                    <CAF_RABBITMQ_PORT>5672</CAF_RABBITMQ_PORT>
+                                    <CAF_RABBITMQ_USERNAME>guest</CAF_RABBITMQ_USERNAME>
+                                    <CAF_RABBITMQ_PASSWORD>guest</CAF_RABBITMQ_PASSWORD>
+                                    <CAF_SCHEDULED_EXECUTOR_PERIOD>2</CAF_SCHEDULED_EXECUTOR_PERIOD>
+                                    <CAF_WORKER_INPUT_QUEUE>jobservicescheduler-in</CAF_WORKER_INPUT_QUEUE>
+                                </env>
+                                <links>
+                                    <link>rabbitmq</link>
+                                    <link>${job.database.host}</link>
+                                </links>
+                                <log>
+                                    <enabled>true</enabled>
+                                </log>
+                                <wait>
+                                    <http>
+                                        <url>http://${docker.host.address}:${scheduler-worker.adminport}/healthcheck</url>
+                                    </http>
+                                    <time>200000</time>
+                                    <shutdown>500</shutdown>
+                                </wait>
+                            </run>
+                        </image>
 
                         <!-- Run the Job Service -->
                         <image>
@@ -478,6 +518,7 @@
                                     <CAF_RABBITMQ_PORT>5672</CAF_RABBITMQ_PORT>
                                     <CAF_RABBITMQ_USERNAME>guest</CAF_RABBITMQ_USERNAME>
                                     <CAF_RABBITMQ_PASSWORD>guest</CAF_RABBITMQ_PASSWORD>
+                                    <CAF_SCHEDULER_INPUT_QUEUE>jobservicescheduler-in</CAF_SCHEDULER_INPUT_QUEUE>
                                     <CAF_JOB_SERVICE_SUSPENDED_PARTITIONS_REGEX>^tenant-acme-c(.?rp|om|o)$</CAF_JOB_SERVICE_SUSPENDED_PARTITIONS_REGEX>
                                     <CAF_JOB_SERVICE_RESUME_JOB_QUEUE>worker-taskunstowing-in</CAF_JOB_SERVICE_RESUME_JOB_QUEUE>
                                 </env>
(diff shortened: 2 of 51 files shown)


xbreizh:
Hi @dermot-hardy so here, I decomposed in several steps:
1- [Added retry on createConnection](https://github.com/JobService/job-service/pull/252/commits/8b14fbe1688f026fd0180606534c09e2c3fd4699)
2- Returning 200 if subtask reported in the database but failing to send a message -> not required anymore as we want to get rid of the sendMessage()
3- Fill job_task_data on each job creation -> done (needs refactoring).
4- Add scheduler function to pickup job_task_data data automatically
The dilemna that I have at the moment here, is that the [getJobCanBeProgressed()](https://github.com/JobService/job-service/blob/develop/job-service-db/src/main/resources/procedures/getJobCanBeProgressed.sql) checks whether a job is in that table in order to know if it's available to run, when with that new way of working, any new job without dependency would be in that table and available for running. So, in order to differentiate and filter, I added  a new column ["Type"](https://github.com/JobService/job-service/blob/3191789e7b76480a3ee3317d4520ccbe183d8b57/job-service-db/src/main/resources/v4.0.0/changelog.xml#L107) with default value as Depending, so passing "Not Depending" on job creation.
In order to have the passing within job-service-container, I added the [job-service-scheduled-executor](https://github.com/JobService/job-service/pull/252/commits/4c254c79b4b3ee8cff29b865c786e61bc11e45cf) and also [changed the build order](https://github.com/JobService/job-service/pull/252/commits/3191789e7b76480a3ee3317d4520ccbe183d8b57) so that it would be available to run.
With that move, we manage to keep the tests passing as before (I removed the sendMessage section from them so we have the behaviour that we seek)
There's only one resistance at the mn that I am [working on](https://github.com/JobService/job-service/blob/3191789e7b76480a3ee3317d4520ccbe183d8b57/job-service-container/src/test/java/com/hpe/caf/services/job/api/JobServiceIT.java#L740). I am unsure if we need to keep it and fix or just get rid of it. I commented it out for now and try to see if I can fix it
5- Remove reportFailure() and associated -> quick cleanup
TODO functionality to poke scheduler on job creation

xbreizh:
@dermot-hardy About the build, it tends to randomly throw a EOFException:
`[ERROR] Failed to execute goal com.googlecode.maven-download-plugin:download-maven-plugin:1.3.0:wget (download-target-file) on project job-service-docs: IO Error: EOFException -> [Help 1]
`
I discussed it with @tonymcveigh and we 

> Could we implement a retry in case in requests it too quickly or decides it's EOF too quickly?
Also, 

> Are there other build doing something similar? Do they use wget?

If we insist a couple of times, the build finally passes, but it might be worth fixing it, either by spending some time on investigating the root source or by using a quick fix as described above. What do you think?

xbreizh:
I removed the extra column that I previously added.
I also temporarily added a Sleep on 2 of the tests in order to let the scheduler wake up and do its work
The next step is to implement a trigger on that scheduler method (pollDatabaseForJobsToRun()) on job creation (job without dependencies)

docs/pom.xml
```diff
@@ -197,6 +197,7 @@
                             <overwrite>true</overwrite>
                             <skipCache>true</skipCache>
                             <url>http://${docker.host.address}:${webServerPort}/docs-site.tar.gz</url>
+                            <retries>10</retries>
```
dermot-hardy: This is nothing to do with this Jira.  Raise a separate Jira/PR for it if it is required.

pom.xml
```diff
@@ -247,11 +247,11 @@
         <module>job-service-db-client</module>
         <module>job-service</module>
         <module>job-service-admin</module>
-        <module>job-service-container</module>
         <module>job-service-internal-client</module>
         <module>job-service-html</module>
         <module>job-service-scheduled-executor</module>
         <module>job-service-scheduled-executor-container</module>
+        <module>job-service-container</module>
```
dermot-hardy, line 254: Why is this being moved?

job-service-container/src/test/java/com/hpe/caf/services/job/api/JobServiceIT.java
```diff
@@ -737,7 +737,7 @@ public void testGetJobsCountFromDifferentPartition() throws ApiException {
      * appropriate web method. It then consumes the message from the queue and asserts the result is as expected.
      * @throws Exception
      */
-    @Test
+   // @Test
```
dermot-hardy: Why is this being commented out?

If it is no longer relevant then why are we keeping it around?  If it can be removed than does that allow any other functions (`testMessagesPutOnQueue` perhaps) to be removed?

job-service/src/main/java/com/hpe/caf/services/job/queue/QueueServicesFactory.java
```diff
@@ -42,7 +45,14 @@
      */
     public static QueueServices create(final AppConfig configuration, final String targetQueue, final Codec codec) throws IOException, TimeoutException {
         //  Create connection and channel for publishing messages.
-        Connection connection = createConnection(configuration);
+        Connection connection;
+        try {
+            //  Create connection and channel for publishing messages.
+            connection = createConnection(configuration);
+        } catch (IOException | TimeoutException e) {
```
dermot-hardy:
```suggestion
        } catch (final IOException | TimeoutException e) {
```

job-service/src/test/java/com/hpe/caf/services/job/api/JobsPutTest.java
```diff
@@ -145,7 +142,6 @@ public void setup() throws Exception {
         JobTypes.initialise(() -> Collections.singletonList(basicJobType));
 
         //  Mock QueueServices calls.
-        doNothing().when(mockQueueServices).sendMessage(any(), any(), any(), any(), anyBoolean());
```
dermot-hardy, line 148: Can the `anyBoolean()` import now be removed?

worker-jobtracking-container/src/test/java/com/hpe/caf/worker/jobtracking/JobDatabase.java
```diff
@@ -26,6 +28,7 @@
 import java.text.MessageFormat;
 import java.time.Instant;
 import java.util.Date;
+import java.util.List;
```
dermot-hardy: Why is this being added?

worker-jobtracking-container/src/test/java/com/hpe/caf/worker/jobtracking/JobDatabase.java
```diff
@@ -59,13 +62,24 @@ public void createJobTask(
...
         try(Connection connection = getConnection();
-            CallableStatement stmt = connection.prepareCall("{call create_job(?,?,?,?,?,?)}")) {
+            CallableStatement stmt = connection.prepareCall("{call create_job(?,?,?,?,?,?,?,?,?,?,?,?,?)}")) {
+
             stmt.setString(1, partitionId);
-            stmt.setString(2, jobId);
-            stmt.setString(3, name);
-            stmt.setString(4, description);
-            stmt.setString(5, data);
-            stmt.setInt(6, jobHash);
```
dermot-hardy: Why are these lines changing?

job-service/src/main/java/com/hpe/caf/services/job/api/JobsGetById.java
```diff
@@ -65,8 +65,10 @@ public static Job getJob(final String partitionId, String jobId) throws Exceptio
             job = databaseHelper.getJob(partitionId, jobId);
 
             LOG.debug("getJobById: Done.");
-            LOG.info("Job progress {}/{} status {}  progression {}", partitionId, jobId, job.getStatus(), job.getPercentageComplete());
-            return job;
+            if (null != job) {
+                LOG.info("Job progress {}/{} status {}  progression {}", partitionId, jobId, job.getStatus(),
+                        job.getPercentageComplete());
+            }            return job;
```
dermot-hardy: Merge `develop` into this branch to get rid of this change.

job-service/src/main/java/com/hpe/caf/services/job/api/JobsPut.java
```diff
@@ -75,6 +63,7 @@
      */
     public static String createOrUpdateJob(final String partitionId, String jobId, NewJob job) throws Exception {
         try {
+            final boolean hasDependencies;
```
dermot-hardy: Surprising to see this from a non-C developer (because in C variables have to be declared at the start of functions).  In modern languages they should be declared as close to their initial usage as is sensible (i.e. just above the `jobCreated` declaration in this case).

job-service/src/main/java/com/hpe/caf/services/job/api/JobsPut.java
```diff
@@ -188,46 +177,28 @@ public static String createOrUpdateJob(final String partitionId, String jobId, N
...
+                        getTaskDataBytes(jobTask, codec), jobTask.getTaskPipe(), jobTask.getTargetPipe(),
+                        job.getDelay(), job.getLabels());
             }
 
             if (!jobCreated) {
                 return "update";
             }
 
-            if (!databaseHelper.canJobBeProgressed(partitionId, jobId)) {
+            if (hasDependencies && !databaseHelper.canJobBeProgressed(partitionId, jobId)) {
```
dermot-hardy: What is the purpose of this?  It looks like you are returning `"create"` either way.

dermot-hardy (changes requested):

job-service/src/main/java/com/hpe/caf/services/job/api/JobsPut.java
xbreizh: Moved down.

job-service/src/main/java/com/hpe/caf/services/job/api/JobsGetById.java
xbreizh: Done

worker-jobtracking-container/src/test/java/com/hpe/caf/worker/jobtracking/JobDatabase.java
xbreizh: restored

worker-jobtracking-container/src/test/java/com/hpe/caf/worker/jobtracking/JobDatabase.java
xbreizh: Removed unused import

job-service/src/test/java/com/hpe/caf/services/job/api/JobsPutTest.java
xbreizh: Removed unused import

job-service-container/src/test/java/com/hpe/caf/services/job/api/JobServiceIT.java
xbreizh: I think that this test is no longer relevant, but I wasn't 100% sure so wanted to check it with you. If we get rid of it, then yes testMessagesPutOnQueue() can be removed too.

pom.xml
xbreizh: [from](https://github.com/JobService/job-service/pull/252#issuecomment-850564347) I added the job-service-scheduled-executor to be available in job-service-container's tests. So, I had to reverse the building order so it gets available there

docs/pom.xml
xbreizh: I added it here in an attempt to get the build stable wrt that above comment (we keep hitting job-service-docs: IO Error: EOFException) but indeed, unrelated to this jira so removed.

job-service/src/main/java/com/hpe/caf/services/job/api/JobsPut.java
xbreizh: True, strangely we were doing it before and I just tried to adapt. I guess there was a use for it some time ago, but you're right, not anymore and it can be removed. Also, that's the only place in the code where we use canJobBeProgressed() so I cleaned that one too

xbreizh:
There is some correction required in the acceptance-tests that I can't figure at the moment.
The manual tests are working as expected -> we can see in the scheduler logs that we trigger it as soon as the job is created , whereas there's a delay when the cron runs.
I added some details in the existing logs to determine the source of the trigger (job creation or cron)

xbreizh:
It seems that the process got slightly slower, I had to keep some delay in a [test](https://github.com/JobService/job-service/blob/49187cdbcb7bcccc3a8fdf839e7471dfbea948d8/job-service-acceptance-tests/src/test/java/com/hpe/caf/jobservice/acceptance/JobServiceEndToEndIT.java#L711) to have it passing.
For it, I used a new library, [awaitability](http://www.awaitility.org/) so I can set a max value for the waiting duration instead of a fixed one. I imagine that in some cases (maybe not that exact one as the duration is minimal) it can help speeding up the tests?

job-service-acceptance-tests/pom.xml
```diff
@@ -455,6 +455,46 @@
                                 </log>
                             </run>
                         </image>
+    
+                        <!-- Run the Job Service Scheduled Executor -->
```
dermot-hardy: Why is this being moved?

job-service-container/pom.xml
```diff
@@ -556,6 +556,35 @@
                                 </wait>
                             </run>
                         </image>
+
+                        <!-- Run the Job Service Scheduled Executor -->
+                        <image>
+                            <alias>job-service-scheduled-executor</alias>
+                            <name>${dockerJobServiceOrg}job-service-scheduled-executor${dockerProjectVersion}</name>
+                            <run>
```
dermot-hardy, line 565: You should probably have a `wait` section in here also.

job-service-scheduled-executor-container/pom.xml
```diff
@@ -171,6 +171,11 @@
                                     </inline>
                                 </assembly>
                             </build>
+                            <run>
+                                <env>
+                                    <CAF_WORKER_INPUT_QUEUE>jobservicescheduler-in</CAF_WORKER_INPUT_QUEUE>
+                                </env>
+                            </run>
```
dermot-hardy: This doesn't make any sense.  Did you mean to put the `env` element inside `build`, for it to become the default?

dermot-hardy (changes requested):

job-service-scheduled-executor-container/pom.xml
xbreizh: Moved into build

job-service-container/pom.xml
xbreizh: Added

job-service-acceptance-tests/pom.xml
xbreizh: The idea was to have it created before the jobservice container. Moved it back down

job-service-acceptance-tests/pom.xml
```diff
@@ -629,12 +630,15 @@
                             <alias>job-service-caller</alias>
                             <name>${dockerJobServiceOrg}job-service-caller${dockerProjectVersion}</name>
                         </image>
-
+
```
dermot-hardy: Pointless whitespace change

job-service-acceptance-tests/pom.xml
```diff
@@ -656,9 +661,15 @@
...
+                                <wait>
+                                    <http>
+                                        <url>http://${docker.host.address}:${scheduler-worker.adminport}/healthcheck</url>
+                                    </http>
+                                    <time>200000</time>
+                                    <shutdown>500</shutdown>
+                                </wait>
                             </run>
                         </image>
-
```
dermot-hardy, line 661: Pointless whitespace change

job-service-config/src/main/java/com/hpe/caf/services/configuration/AppConfig.java
```diff
@@ -15,16 +15,16 @@
...
-import org.springframework.context.annotation.Configuration;
-import org.springframework.context.annotation.PropertySource;
-import org.springframework.core.env.Environment;
-
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Locale;
 import java.util.regex.Pattern;
 
+import org.springframework.beans.factory.annotation.Autowired;
```
dermot-hardy: Put the imports back where they were - moving them is not for this Jira

job-service-config/src/main/java/com/hpe/caf/services/configuration/AppConfig.java
```diff
@@ -120,4 +120,12 @@ public String getResumeJobQueue()
     {
         return environment.getProperty("CAF_JOB_SERVICE_RESUME_JOB_QUEUE");
     }
+
```
dermot-hardy:
```suggestion

```

job-service/src/main/java/com/hpe/caf/services/job/queue/QueueServices.java
```diff
@@ -68,6 +68,7 @@ public void sendMessage(
         final String partitionId, String jobId, WorkerAction workerAction, AppConfig config, final boolean includeTrackingInfo
     ) throws IOException, InterruptedException, TimeoutException
     {
+        LOG.info("sending message");
```
dermot-hardy: Do we need this?  Surely it will fire quite a lot and it doesn't tell us much.  If you think it should stay then should it be downgraded to debug, and should it include at least some details about the message being sent?

job-service/src/main/java/com/hpe/caf/services/job/queue/QueueServices.java
```diff
@@ -122,7 +123,6 @@ public void sendMessage(
                 trackingInfo,
                 null,
                 MDC.get(MDC_KEY));
-
```
dermot-hardy: Put this blank line back.

job-service/src/main/java/com/hpe/caf/services/job/queue/QueueServices.java
```diff
@@ -141,9 +141,10 @@ public void sendMessage(
         }
     }
 
-    private void publishMessage(final byte[] taskMessageBytes)
+    public void publishMessage(final byte[] taskMessageBytes)
             throws IOException, InterruptedException, TimeoutException
     {
+        LOG.info("target queue: {}", targetQueue);
```
dermot-hardy: Do we need this? Surely it will fire quite a lot and it doesn't tell us much. If you think it should stay then should it be downgraded to debug?

job-service/src/main/java/com/hpe/caf/services/job/api/JobsPut.java
```diff
@@ -162,13 +164,6 @@ public static String createOrUpdateJob(final String partitionId, String jobId, N
                 throw new BadRequestException(ERR_MSG_TASK_DATA_DATATYPE_ERROR);
             }
 
-            //  Load serialization class.
-            Codec codec = ModuleLoader.getService(Codec.class);
-
-            //  Get app config settings.
-            LOG.debug("createOrUpdateJob: Reading database and RabbitMQ connection properties...");
-            AppConfig config = AppConfigProvider.getAppConfigProperties();
```
dermot-hardy: Making these static will have an effect on what happens if there is an error.  Does it need to be done?

job-service/src/main/java/com/hpe/caf/services/job/api/JobsPut.java
```diff
@@ -190,44 +185,22 @@ public static String createOrUpdateJob(final String partitionId, String jobId, N
...
-                mapper.setDateFormat(df);
-                databaseHelper.reportFailure(partitionId, jobId, mapper.writeValueAsString(f));
-
-                //  Throw error message to user.
-                throw new ServiceUnavailableException("Failed to add task data to the queue", ex);
-            }
-
+    
+            triggerScheduler();
+
```
dermot-hardy:
```suggestion

            triggerScheduler();

```

job-service/src/main/java/com/hpe/caf/services/job/api/JobsPut.java
```diff
@@ -236,17 +209,33 @@ public static String createOrUpdateJob(final String partitionId, String jobId, N
...
-            LOG.warn("Error on connection close to RabbitMQ", e);
+        try (final QueueServices queueServices = QueueServicesFactory.create(config, config.getSchedulerQueue(), codec)){
+            
+            LOG.debug("createOrUpdateJob: Triggering scheduler to send data to the target queue");
+            
+            final TaskMessage taskMessage = new TaskMessage(
+                    UUID.randomUUID().toString(),
+                    DocumentWorkerConstants.DOCUMENT_TASK_NAME,
+                    1,
+                    OBJECT_MAPPER.writeValueAsBytes(new DocumentWorkerDocumentTask()),
```
dermot-hardy: Can we just use `codec.serialise()` for this?  I'm not sure why we're introducing `OBJECT_MAPPER`.

job-service/src/main/java/com/hpe/caf/services/job/api/JobsPut.java
```diff
@@ -262,16 +251,41 @@ private static void closeQueueConnection(final QueueServices queueServices)
...
+            taskDataBytes = serializingData(taskDataObj);
         } else {
             throw new RuntimeException("The taskData is an unexpected type");
         }
 
         return taskDataBytes;
     }
-
+    
+    private static byte[] serializingData(final Object taskDataObj)
```
dermot-hardy: You don't normally use the `-ing` form of the verb when naming methods.  Can this be changed to `serializeData`?

job-service/src/main/java/com/hpe/caf/services/job/api/JobsPut.java
```diff
@@ -236,17 +209,33 @@ public static String createOrUpdateJob(final String partitionId, String jobId, N
...
+                    DocumentWorkerConstants.DOCUMENT_TASK_NAME,
+                    1,
+                    OBJECT_MAPPER.writeValueAsBytes(new DocumentWorkerDocumentTask()),
+                    TaskStatus.NEW_TASK,
+                    Collections.emptyMap(),
+                    config.getSchedulerQueue());
+            final byte[] taskMessageBytes = serializingData(taskMessage);
+            queueServices.publishMessage(taskMessageBytes);
+        } catch (final Exception ex) {
+           LOG.error("fail to ping the scheduler {}", ex.getMessage());
```
dermot-hardy: Downgrade to `WARN`.  Also start log statements with capital letters like a sentence, and use past tense where appropriate (i.e. `fail` --> `Failed`).

job-service/src/main/java/com/hpe/caf/services/job/api/JobsPut.java
```diff
@@ -236,17 +209,33 @@ public static String createOrUpdateJob(final String partitionId, String jobId, N
...
         }
     }
-
-    private static byte[] getTaskDataBytes(final WorkerAction workerAction, final Codec codec)
+    
+    private static byte[] getTaskDataBytes(final WorkerAction workerAction)
```
dermot-hardy: Were the changes to this method necessary?

job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/ScheduledExecutor.java
```diff
@@ -29,28 +31,43 @@
...
     public static void main(final String[] args)
     {
+        runJobs();
+    }
+
+    private static void runJobs() {
         // Create a scheduler to process scheduled tasks.
         final ScheduledExecutorService scheduler = Executors.newScheduledThreadPool(1);
-
+
```
dermot-hardy: Pointless whitespace change

job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/ScheduledExecutor.java
```diff
@@ -29,28 +31,43 @@
...
-
+        
+        final Runnable task = () -> runAvailableJobs("Auto");
+    
         //  Poll the Job Service database using the specified polling period configuration to specify how often the
         //  scheduled task is run.
         scheduler.scheduleWithFixedDelay(task, 20, ScheduledExecutorConfig.getScheduledExecutorPeriod(),
                 TimeUnit.SECONDS);
-
+
```
dermot-hardy: Pointless whitespace change

dermot-hardy (approved):

job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/ScheduledExecutor.java
```diff
@@ -29,28 +31,43 @@
...
+     *
+     * @param origin the trigger's origin. It can be "Auto" or "Manual"
+     */
+    public static void runAvailableJobs(final String origin)
+    {
+        try {
+            final Instant start = Instant.now();
+            DatabasePoller.pollDatabaseForJobsToRun();
+            final Instant end = Instant.now();
+            LOG.debug("Total time taken to execute scheduler for {} task in ms {}", origin, Duration.between(start, end).toMillis());
```
dermot-hardy: Consider using `isDebugEnabled()` when there is extra logic for the logging.

```suggestion
            if (LOG.isDebugEnabled()) {
                final Instant start = Instant.now();
                DatabasePoller.pollDatabaseForJobsToRun();
                final Instant end = Instant.now();
                LOG.debug("Total time taken to execute scheduler for {} task in ms {}", origin, Duration.between(start, end).toMillis());
            } else {
                DatabasePoller.pollDatabaseForJobsToRun();
            }
```

xbreizh: Ok, I didn't know that one. Thx

job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/ScheduledExecutor.java
xbreizh: Removed

job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/ScheduledExecutor.java
xbreizh: Removec

job-service/src/main/java/com/hpe/caf/services/job/api/JobsPut.java
xbreizh: Aiming to reduce the number of argument by making the codec global. Restoring

job-service/src/main/java/com/hpe/caf/services/job/api/JobsPut.java
xbreizh: Done

job-service/src/main/java/com/hpe/caf/services/job/api/JobsPut.java
xbreizh: Changed

job-service/src/main/java/com/hpe/caf/services/job/api/JobsPut.java
xbreizh: replaced

job-service/src/main/java/com/hpe/caf/services/job/api/JobsPut.java
xbreizh: The idea behind it was to help reducing the number of parameters passed between functions. I don't get what the downside would be in case of error? More risk of error if we call the function several times instead of storing into a parameter and pass it along?

job-service/src/main/java/com/hpe/caf/services/job/queue/QueueServices.java
xbreizh: Removed

job-service/src/main/java/com/hpe/caf/services/job/queue/QueueServices.java
xbreizh: No, that can go I think, I added it for debugging. Removed

job-service-config/src/main/java/com/hpe/caf/services/configuration/AppConfig.java
xbreizh: Sorry, it's just the IDE ordering alphabetically. Done

job-service-acceptance-tests/pom.xml
xbreizh: Done

job-service/src/main/java/com/hpe/caf/services/job/api/JobsPut.java
dermot-hardy: The problem is the context in which it is initialized.  By doing this you moved initialization from when the `createOrUpdateJob()` call is made to a time before any of the class methods are called (possibly even to service startup).  When it was when the `createOrUpdateJob()` call is made the error could be bubbled back to the client whereas that cannot happen during class initialization.  If you just want to reduce the parameters between functions without changing the initialization context then one way would be to use a nested class.  For example [`RandomSamplingWorker.Task`](https://github.houston.softwaregrp.net/Verity/worker-sampling/blob/v1.1.1/worker-sampling/src/main/java/com/microfocus/apollo/workers/sampling/RandomSamplingWorker.java#L101) takes that approach.  Note that the parameters are grouped into a nested object.  They are then used by [`process()`](https://github.houston.softwaregrp.net/Verity/worker-sampling/blob/v1.1.1/worker-sampling/src/main/java/com/microfocus/apollo/workers/sampling/RandomSamplingWorker.java#L141), [`getSamplingDispatchJobParametersDigest()`](https://github.houston.softwaregrp.net/Verity/worker-sampling/blob/v1.1.1/worker-sampling/src/main/java/com/microfocus/apollo/workers/sampling/RandomSamplingWorker.java#L232), [`createSamplingDispatchJobParameters()`](https://github.houston.softwaregrp.net/Verity/worker-sampling/blob/v1.1.1/worker-sampling/src/main/java/com/microfocus/apollo/workers/sampling/RandomSamplingWorker.java#L248), etc., all without being repeatedly passed around.

dermot-hardy (approved):

job-service/src/main/java/com/hpe/caf/services/job/api/JobsPut.java
xbreizh: Oh ok, thanks for the explanation.

job-service-acceptance-tests/src/test/java/com/hpe/caf/jobservice/acceptance/JobServiceEndToEndIT.java
```diff
@@ -680,6 +680,7 @@ public void testCreateJobNoDelayNoPreReq() throws Exception {
         //retrieve job using web method
         final Job retrievedJob = jobsApi.getJob(partitionId, jobId, jobCorrelationId);
         LOG.info("--testCreateJobNoDelayNoPreReq job {} in partition: {}", retrievedJob.getId(), partitionId);
+        Thread.sleep(100); // Allows the scheduler to perform
```
dermot-hardy: You could consider removing this test altogether.  It's trying to check an internal implementation detail so it's not a good end-to-end test.

job-service-acceptance-tests/src/test/java/com/hpe/caf/jobservice/acceptance/JobServiceEndToEndIT.java
```diff
@@ -706,6 +707,8 @@ public void testCreateJobNoDelayAndSomePreReq() throws Exception {
         LOG.info("--testCreateJobNoDelayAndSomePreReq job {} in partition: {}, canRun? {}", job1Id, partitionId, canRun);
         assertEquals(canRun, false, "Job "+job1Id+" is eligible to run despite incomplete prerequisite jobs");
 
+        Thread.sleep(100); // Allows the scheduler to perform
+
         JobServiceDatabaseUtil.assertJobTaskDataRowDoesNotExist(preReqJobId);
```
dermot-hardy: Same with this part of this test (i.e. the `final` assert that you've had to add a delay for).

job-service-acceptance-tests/src/test/java/com/hpe/caf/jobservice/acceptance/JobServiceEndToEndIT.java
xbreizh: Removed test

job-service-acceptance-tests/src/test/java/com/hpe/caf/jobservice/acceptance/JobServiceEndToEndIT.java
xbreizh: Removed test

PR #253, opened 2021-06-09
SCMOD-13865: Remove propagate_failures

xbreizh:
https://portal.digitalsafe.net/browse/SCMOD-13865


dermot-hardy (approved):

pom.xml
```diff
@@ -27,7 +27,7 @@
 
     <groupId>com.github.jobservice</groupId>
     <artifactId>job-service-aggregator</artifactId>
-    <version>4.1.0-SNAPSHOT</version>
+    <version>5.0.0-SNAPSHOT</version>
```
dermot-hardy, line 30: What about all the other pom files?

dermot-hardy (changes requested):

pom.xml
xbreizh: Done

PR #254, opened 2021-06-15
SCMOD-7044: Use HTTP data store in integration tests

rorytorneymf:
https://portal.digitalsafe.net/browse/SCMOD-7044

### DIFF
--- a/job-service-container/pom.xml
+++ b/job-service-container/pom.xml
@@ -138,11 +138,6 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>com.github.workerframework</groupId>
-            <artifactId>worker-store-fs</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
             <groupId>org.apache.httpcomponents</groupId>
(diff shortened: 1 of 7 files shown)


dermot-hardy (approved):

job-service-container/pom.xml
```diff
@@ -137,11 +137,6 @@
             <artifactId>${caf.worker-queue.impl}</artifactId>
             <scope>test</scope>
         </dependency>
-        <dependency>
```
rorytorneymf, line 140: When the  `job-service-container` tests startup they look for a data store implementation (they're not using the worker testing framework though, I'm not entirely sure how they work).

I thought I would go ahead and remove the worker-store-fs test dependency here anyway so that worker-store-http is used instead:

```
-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running TestSuite
...
[main] INFO com.hpe.caf.util.ModuleLoader - Detected component implementation HttpDataStoreProvider
```

dermot-hardy (approved):

PR #257, opened 2021-07-16
SCMOD-14400: Adapt test to consider scheduled-executor

xbreizh:
https://portal.digitalsafe.net/browse/SCMOD-14400


xbreizh:
Here, that [test ](https://github.com/JobService/job-service/blob/64cc248323806c8d4ce97f2e423951697666fd34/job-service-container/src/test/java/com/hpe/caf/services/job/api/JobServiceIT.java#L1087)tends to fail as the number of deleted tables differs from the expected one.
That's probably due to the fact that we added the scheduled-executor in those tests, which also performs that same action.
We could:

- mitigate the impact of the scheduled-executor by having it performing at a lower rate

- or we can remove the check for the number of tables removes.

I opted for the latter, given that we also check the ultimate goal, which is that there is no [remaining table](https://github.com/JobService/job-service/blob/64cc248323806c8d4ce97f2e423951697666fd34/job-service-container/src/test/java/com/hpe/caf/services/job/api/JobServiceIT.java#L1132)

dermot-hardy (approved):

PR #261, opened 2021-07-23
 SCMOD-13870: Restore single threading

xbreizh:


dermot-hardy (changes requested):
This is not right at all.  This would still execute on two threads.  In fact in would keep creating new thread pools each time a message is received.

PR #264, opened 2021-08-02
SCMOD-13870: Making scheduler resilient to rabbitMQ disconnections

xbreizh:

### DIFF
--- a/job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/DatabasePoller.java
+++ b/job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/DatabasePoller.java
@@ -70,7 +70,7 @@ public class DatabasePoller
             LOG.error(MessageFormat.format("Exception caught when loading the serialization class. {0}", e.getMessage()));
         }
     }
-    
+
     private static void sendMessageToQueueMessaging(final Codec codec, final JobTaskData jtd, final WorkerAction workerAction)
     {
         try (final QueueServices queueServices= QueueServicesFactory.create(jtd.getTaskPipe(), codec)){
--- a/job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/DatabasePoller.java
+++ b/job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/DatabasePoller.java
@@ -15,8 +15,6 @@
  */
 package com.hpe.caf.services.job.scheduled.executor;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.databind.ObjectMapper;
 import com.hpe.caf.api.Codec;
 import com.hpe.caf.util.ModuleLoader;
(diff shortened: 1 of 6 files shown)


xbreizh:
@dermot-hardy Looking once more at the code, it seems that we are using _report_failure_ only in the event of connection failure with rabbitMQ. If that's so, I believe that we could even go further and remove the concept of a job failing as a whole. That seems odd, so I would prefer to get your opinion on it before engaging in further cleanup

job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/QueueServices.java
```diff
@@ -160,7 +160,7 @@ private static long getStatusCheckIntervalMillis(final String statusCheckInterva
      * Closes the queue connection.
      * @throws Exception thrown if the queue connection cannot be closed.
      */
-    public void close() throws Exception {
+    public void close(){
```
dermot-hardy: - Since you're now implementing `AutoCloseable` you should also be adding the `@Override` annotation to this method.
- I'm not clear why `throws Exception` is being removed.  Either don't remove it, or if you are removing it then also remove the `@throws Exception` line in the JavaDoc (and also add a space before the opening brace).
I'm actually quite surprised this is compiling on the build machine because it's not compiling for me:
```QueueServices.java:161: error: exception not thrown: java.lang.Exception```

File: job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/DatabasePoller.java [hunk 2]
```diff
@@ -118,6 +71,17 @@ public static void pollDatabaseForJobsToRun() {
...
     }
 
+    private static void sendMessageToQueueMessaging(final Codec codec, final JobTaskData jtd, final WorkerAction workerAction)
+    {
+        try (final QueueServices queueServices= QueueServicesFactory.create(jtd.getTaskPipe(), codec)){
+            LOG.debug("Sending task data to the target queue {} ...", workerAction);
+            queueServices.sendMessage(jtd.getPartitionId(), jtd.getJobId(), workerAction);
+            deleteDependentJob(jtd.getPartitionId(), jtd.getJobId());
+        } catch(final Exception ex) {
+            LOG.warn("Failed to send message about dependent jobs {}", ex.getMessage());
```
dermot-hardy: Don't just log the exception message.

```suggestion
            LOG.warn("Failed to send message about dependent jobs", ex);
```

dermot-hardy (approved):

job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/QueueServices.java
xbreizh: I removed as we never throw it in the function, but a RuntimeException instead. I understand that for the unchecked exceptions, there is no need to declare them at method level. 
However, I imagine that's good practice to have them declared, so I ll restore it as a RuntimeException.
I adapted the javadoc to reflect it.

PR #265, opened 2021-09-06
SCMOD-12113: Update swaggerUI

xbreizh:
https://portal.digitalsafe.net/browse/SCMOD-12113


dermot-hardy (changes requested):
Doesn't look right to me.

dermot-hardy (approved):

PR #268, opened 2021-09-16
SCMOD-14090: Upgrade dependencies

xbreizh:
https://portal.digitalsafe.net/browse/SCMOD-14090


xbreizh:
@dermot-hardy I noticed that the postgres version was behind, so added it too

dermot-hardy (approved):

PR #270, opened 2021-10-01
SCMOD-15074 Job Service: Cancellation of complex jobs can fail

simranjeetc:
https://portal.digitalsafe.net/browse/SCMOD-15074

### DIFF
--- a/job-service-container/pom.xml
+++ b/job-service-container/pom.xml
@@ -179,7 +179,7 @@
                 </property>
             </activation>
             <properties>
-                <task.table.deletion.count>10000</task.table.deletion.count>
+                <task.table.deletion.count>30000</task.table.deletion.count>
             </properties>
         </profile>
     </profiles>
--- a/job-service-container/pom.xml
+++ b/job-service-container/pom.xml
@@ -179,7 +179,7 @@
                 </property>
             </activation>
             <properties>
-                <task.table.deletion.count>30000</task.table.deletion.count>
+                <task.table.deletion.count>10000</task.table.deletion.count>
             </properties>
         </profile>
     </profiles>
--- a/job-service-container/src/test/java/com/hpe/caf/services/job/api/JobServiceIT.java
+++ b/job-service-container/src/test/java/com/hpe/caf/services/job/api/JobServiceIT.java
@@ -1092,18 +1092,19 @@ public class JobServiceIT {
 
         try(final java.sql.Connection dbConnection = JobServiceConnectionUtil.getDbConnection())
         {
-            final int totalParentTableCount = Integer.parseInt(System.getProperty("task.table.deletion.count"));
+            final int totalTableCount = Integer.parseInt(System.getProperty("task.table.deletion.count"));
+            final int totalParentTableCount = totalTableCount/96; // Each parent table results in creation of 96 child/sub-child tables.
             LOG.info("Creating tables");
             final Instant startTableCreation = Instant.now();
             IntStream
-                    .range(1, totalParentTableCount)
+                    .rangeClosed(1, totalParentTableCount)
                     .forEach((count) -> {
                         final String jobIdentity = String.valueOf(count);
                         final String parentTableName = "task_" + jobIdentity;
                         deletedOrCancelledJobs.add(parentTableName);
                         createTaskTable(dbConnection, parentTableName);
                         insertRecordsInTaskTable(dbConnection, parentTableName, 20);
-                        createAndPopulateChildTables(dbConnection, parentTableName);
+                        createAndPopulateChildTables(dbConnection, parentTableName); //parentTableName - task_1, task_2 ...
                     });
             final Instant endTableCreation = Instant.now();
             LOG.info("Total time taken to create " + getAllTablesByPattern(dbConnection).size() + " tables in ms. " + 
@@ -1142,12 +1143,21 @@ public class JobServiceIT {
     {
         IntStream.range(1, 20)
                 .forEach((childCount) -> {
-                    //create child task tables
+                    //create child task tables - childTableName - task_1.1, task_1.2...
                     final String childTableName = parentTableName + "." + childCount;
                     createTaskTable(dbConnection, childTableName);
 
-                    //insert records into task table
-                    insertRecordsInTaskTable(dbConnection, childTableName, 1);
+                    final int rowCountInChildTable = 5;
+                    insertRecordsInTaskTable(dbConnection, childTableName, rowCountInChildTable); //childTableName - task_1.1, task_1.2...
+
+                    final int subChildTableCount = rowCountInChildTable;
+                    IntStream.range(1,subChildTableCount)
+                            .forEach(subChildCount -> {
+                                final String subChildTableName = childTableName + "." + subChildCount;
+                                createTaskTable(dbConnection, subChildTableName); // subChildTableName - task_1.1.1, task_1.1.2...
+
+                                insertRecordsInTaskTable(dbConnection, subChildTableName, 1);
+                            });
                 });
     }
 
--- a/job-service-container/src/test/java/com/hpe/caf/services/job/api/JobServiceIT.java
+++ b/job-service-container/src/test/java/com/hpe/caf/services/job/api/JobServiceIT.java
@@ -164,7 +164,7 @@ public class JobServiceIT {
 
     @BeforeTest
     public void setup() throws Exception {
-        connectionString = System.getenv("webserviceurl");
+        connectionString = "http://127.0.0.1:25080/job-service/v1";
 
         //Populate maps for testing    
         taskMessageParams.put("datastorePartialReference", "sample-files");
@@ -190,8 +190,8 @@ public class JobServiceIT {
         workerServices = WorkerServices.getDefault();
         configurationSource = workerServices.getConfigurationSource();
         rabbitConfiguration = configurationSource.getConfiguration(RabbitWorkerQueueConfiguration.class);
-        rabbitConfiguration.getRabbitConfiguration().setRabbitHost(SettingsProvider.defaultProvider.getSetting(SettingNames.dockerHostAddress));
-        rabbitConfiguration.getRabbitConfiguration().setRabbitPort(Integer.parseInt(SettingsProvider.defaultProvider.getSetting(SettingNames.rabbitmqNodePort)));
+        rabbitConfiguration.getRabbitConfiguration().setRabbitHost("127.0.0.1");
+        rabbitConfiguration.getRabbitConfiguration().setRabbitPort(Integer.parseInt("5672"));
         rabbitConn = RabbitUtil.createRabbitConnection(rabbitConfiguration.getRabbitConfiguration());
     }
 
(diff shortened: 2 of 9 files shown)


job-service-container/pom.xml
```diff
@@ -179,7 +179,7 @@
                 </property>
             </activation>
             <properties>
-                <task.table.deletion.count>30000</task.table.deletion.count>
+                <task.table.deletion.count>10000</task.table.deletion.count>
```
simranjeetc: This number now corresponds to number of parent tables. For every parent table, we are creating 20 child tables(child + parent). So approximately 10k * 20 tables are getting created in total.

job-service-container/src/test/java/com/hpe/caf/services/job/api/JobServiceIT.java
```diff
@@ -164,19 +165,19 @@ private NewJob makeRestrictedJob(final String jobId, final String typeId, final
 
     @BeforeTest
     public void setup() throws Exception {
-        connectionString = System.getenv("webserviceurl");
+        connectionString = "http://127.0.0.1:25080/job-service/v1";
```
dermot-hardy: Is this meant to be changing?

job-service-container/src/test/java/com/hpe/caf/services/job/api/JobServiceIT.java
```diff
@@ -164,19 +165,19 @@ private NewJob makeRestrictedJob(final String jobId, final String typeId, final
...
 
         //Populate maps for testing    
         taskMessageParams.put("datastorePartialReference", "sample-files");
         taskMessageParams.put("documentDataInputFolder", "/mnt/caf-datastore-root/sample-files");
         taskMessageParams.put("documentDataOutputFolder", "/mnt/bla");
-        
+
```
dermot-hardy: Pointless whitespace change

job-service-container/src/test/java/com/hpe/caf/services/job/api/JobServiceIT.java
```diff
@@ -164,19 +165,19 @@ private NewJob makeRestrictedJob(final String jobId, final String typeId, final
...
         testDataObjectMap.put("taskClassifier", "*.txt");
         testDataObjectMap.put("batchType", "WorkerDocumentBatchPlugin");
         testDataObjectMap.put("taskMessageType", "DocumentWorkerTaskBuilder");
         testDataObjectMap.put("taskMessageParams", taskMessageParams);
         testDataObjectMap.put("targetPipe", "languageidentification-in");
-        
+
```
dermot-hardy: Pointless whitespace change

job-service-container/src/test/java/com/hpe/caf/services/job/api/JobServiceIT.java
```diff
@@ -190,8 +191,8 @@ public void setup() throws Exception {
         workerServices = WorkerServices.getDefault();
         configurationSource = workerServices.getConfigurationSource();
         rabbitConfiguration = configurationSource.getConfiguration(RabbitWorkerQueueConfiguration.class);
-        rabbitConfiguration.getRabbitConfiguration().setRabbitHost(SettingsProvider.defaultProvider.getSetting(SettingNames.dockerHostAddress));
-        rabbitConfiguration.getRabbitConfiguration().setRabbitPort(Integer.parseInt(SettingsProvider.defaultProvider.getSetting(SettingNames.rabbitmqNodePort)));
+        rabbitConfiguration.getRabbitConfiguration().setRabbitHost("127.0.0.1");
+        rabbitConfiguration.getRabbitConfiguration().setRabbitPort(Integer.parseInt("5672"));
```
dermot-hardy: Halting review... PR clearly not prepared for it.

dermot-hardy (changes requested):

job-service-db/src/main/resources/v5.0.0/changelog.xml
```diff
@@ -0,0 +1,36 @@
...
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext
+                                       http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd
+                                       http://www.liquibase.org/xml/ns/dbchangelog
+                                       http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.3.xsd">
+    <changeSet id="create_deleted_parent_table_log" author="Gamma Team">
+        <createTable tableName="deleted_parent_table_log">
+            <column name="table_name" type="VARCHAR(80)">
```
dermot-hardy: Why is this 80 rather than 63 like the `delete_log`?

job-service-db/src/main/resources/procedures/dropDeletedTaskTables.sql
```diff
@@ -18,18 +18,36 @@
...
+                    $q$SELECT table_name FROM deleted_parent_table_log LIMIT $q$ || commit_limit || $q$ FOR UPDATE SKIP LOCKED$q$;
+    WHILE EXISTS(SELECT 1 FROM deleted_parent_table_log)
+        LOOP
+            FOR parent_table_log_rec IN EXECUTE selected_parent_table_names
+                LOOP
+                    CALL populate_delete_log_table(parent_table_log_rec.table_name, 0);
+                    -- delete the parent table name from parent table.
+                    DELETE FROM deleted_parent_table_log WHERE table_name = parent_table_log_rec.table_name;
+                    COMMIT;
+                END LOOP;
```
dermot-hardy: Did you mean to put the commit after the end loop?

job-service-db/src/main/resources/procedures/dropDeletedTaskTables.sql
```diff
@@ -18,18 +18,36 @@
...
+        END LOOP;
```
dermot-hardy: Pull the whole loop block in 4 spaces and pull the inner loop in 8 spaces so you're using the same spacing convention that is already used in this file.  When in Rome...

job-service-db/src/main/resources/procedures/populateDeleteLogTable.sql
```diff
@@ -0,0 +1,56 @@
...
+--
+
+/*
+ *  Name: populate_delete_log_table
+ *
+ *  Description:
+ *  This procedure populates deleted_log table with the names of all task tables that are to be dropped.
+ */
+
+CREATE OR REPLACE PROCEDURE populate_delete_log_table(
```
dermot-hardy: Prefix procedure name with `internal_`, in line with the convention in this project for identifying routines not intended to be called directly by the code.

job-service-db/src/main/resources/procedures/populateDeleteLogTable.sql
```diff
@@ -0,0 +1,56 @@
...
+    in_task_id VARCHAR(70),
```
dermot-hardy: - As far as I can see you're actually passing in the task table name here - rename parameter to `in_task_table_name`.
- I think this should be `VARCHAR(63)`.

job-service-db/src/main/resources/procedures/populateDeleteLogTable.sql
```diff
@@ -0,0 +1,56 @@
...
+    IF internal_to_regclass(task_table_ident) IS NOT NULL THEN
+        FOR subtask_suffix IN
+            EXECUTE $ESC$SELECT '.' || subtask_id || CASE WHEN is_final THEN '*' ELSE '' END AS subtask_suffix FROM $ESC$ ||
+                    task_table_ident
+            LOOP
+                query_count := query_count + 1;
+                IF query_count >= commit_limit THEN
+                    COMMIT;
+                    query_count = 0;
+                END IF;
```
dermot-hardy: Do we really need any of this `query_count` stuff?  Isn't the fact that you've changed the recursion from functions to procedures enough to ensure that the select statements are not in the same transaction?

dermot-hardy (changes requested):

job-service-db/src/main/resources/procedures/populateDeleteLogTable.sql
simranjeetc: I am not sure if we need to accommodate the scenario where the recursion for one parent task table can result in doing a select on ~15k child(sub-child and so on) tables. If that is a valid scenario then query_count based commit is required otherwise I can get rid of it. This would essentially mean getting rid of the commit statement present at line 47. 

Stored procedures by itself do not provide a transaction boundary but rather provide the ability to use transaction ending statements like `COMMIT`, `ROLLBACK` etc. `BEGIN` and `END` do not serve as transaction boundary in postgres either. Actually, if we remove the commit on line 47, we can choose to change this stored procedure to a function. This would still work because of the presence of [this commit statement](https://github.com/JobService/job-service/blob/118009a2295fc118484d3635f38ca2c819b785cf/job-service-db/src/main/resources/procedures/dropDeletedTaskTables.sql#L47). Of-course that would need to satisfy the above mentioned condition of not doing a select on more than ~15k tables(child + sub-child + ...).

job-service-db/src/main/resources/procedures/dropDeletedTaskTables.sql
simranjeetc: This can indeed be moved after the `END LOOP`. It would depend on the what we do [here](https://github.com/JobService/job-service/pull/270/files#r731768604). Because, if we remove [this commit statement ](https://github.com/JobService/job-service/blob/118009a2295fc118484d3635f38ca2c819b785cf/job-service-db/src/main/resources/procedures/populateDeleteLogTable.sql#L47) and move [this](https://github.com/JobService/job-service/blob/118009a2295fc118484d3635f38ca2c819b785cf/job-service-db/src/main/resources/procedures/dropDeletedTaskTables.sql#L47) after the `END LOOP`, then we should be ok with the scenario where iterating over 10(commit_limit) parent tables + child tables + .. goes over 15k table count collectively. 

If we keep [this commit statement](https://github.com/JobService/job-service/blob/118009a2295fc118484d3635f38ca2c819b785cf/job-service-db/src/main/resources/procedures/populateDeleteLogTable.sql#L47), then this can be moved out without much concern of running into max transactions error.

job-service-db/src/main/resources/procedures/populateDeleteLogTable.sql
dermot-hardy: > I am not sure if we need to accommodate the scenario where the recursion for one parent task table can result in doing a select on ~15k child(sub-child and so on) tables. If that is a valid scenario...

I would assume that we do need to accommodate that scenario and that it is valid.

> Stored procedures by itself do not provide a transaction boundary but rather provide the ability to use transaction

Ok, it's the unbalanced nature of it that I'm wary of but yes that answers this query.

job-service-db/src/main/resources/procedures/populateDeleteLogTable.sql
```diff
@@ -0,0 +1,56 @@
...
+/*
+ *  Name: populate_delete_log_table
+ *
+ *  Description:
+ *  This procedure populates deleted_log table with the names of all task tables that are to be dropped.
+ */
+
+CREATE OR REPLACE PROCEDURE populate_delete_log_table(
+    in_task_id VARCHAR(70),
+    query_count INTEGER DEFAULT 0
```
dermot-hardy: I think the convention in this project is for parameters to be prefixed with `in_`.  (I must admit I hate that convention but when in Rome...)

simranjeetc:
@dermot-hardy Do you know why release builds not executing for this repo? http://sou-jenkins2.swinfra.net:8080/job/simranjeetc/job/simranjeetc~job-service~SCMOD-15074/17/console
image

job-service-container/pom.xml
simranjeetc: With 10k parent table count, this would take ~20-25 mins to populate the data for this test in release build. Should I change this to a lower number say 3k parent tables which would result in creation of approx 60k tables overall?

dermot-hardy:
Set `SUSPENDED` to `FALSE` and try it again.

job-service-container/pom.xml
dermot-hardy: Let's adjust it so that it is roughly the same amount of testing as before.

job-service-container/src/test/java/com/hpe/caf/services/job/api/JobServiceIT.java
```diff
@@ -1087,54 +1088,127 @@ public void testInvalidLabelFormat() {
     public void testDeleteLog() throws SQLException
```
dermot-hardy, line 1088: Maybe we should test another level of depth.

dermot-hardy (approved):

job-service-container/src/test/java/com/hpe/caf/services/job/api/JobServiceIT.java
simranjeetc: Addressed in 990393e

dermot-hardy (approved):

PR #273, opened 2021-11-24
SCMOD-15887: Flyway migration

xbreizh:
https://portal.digitalsafe.net/browse/SCMOD-15887


job-service-db/src/main/resources/db/migration/procedures/R___old.sql
```diff
@@ -0,0 +1,276 @@
+--
```
simranjeetc: As was discussed, we should re-think about the naming the repeatable migration scripts. If I understand correctly, we are at present depending on the filenames to assure the order in which the scripts are executed. Does that mean when there is a requirement to insert a script between the existing scripts, we will have to come up with weird names to guarantee the order of execution?

simranjeetc:
We should also test this with some data populated in the jobservice tables.

job-service-db/src/main/resources/db/migration/procedures/R___old.sql
xbreizh: As per [Flyway documentation](https://flywaydb.org/documentation/concepts/migrations.html#repeatable-migrations): Within a single migration run, repeatable migrations are always applied last, after all pending versioned migrations have been executed. Repeatable migrations are applied in the order of their description.
image

xbreizh: Quick note for that [article](https://flywaydb.org/blog/flyway-timestampsandrepeatables) which could bring an element of response

xbreizh: A repeatable migration in Flyway is a migration that is run whenever its checksum changes. The scripts run in alphabetical order if their checksum is changed. This is similar to what is currently happening in our Liquibase configuration. Only that in Liquibase, we would solve some order issues by grouping a function depending on another into a same changeset so we control that order. We might want to change the filenaming convention in the future, but I could see only a single situation where we have to do that in Liquibase. [Here](https://github.com/JobService/job-service/blob/6817f9dfa8fea2df623a99546a533c6b73c2cca5/job-service-db/src/main/resources/procedures/changelog.xml#L78) 
About the introduction of a ___old.sql file where we could set the obsolete functions and signatures, I believe that it could be a good improvement for readability in the different scripts.
However, it could be error prone as we would have to do with 2 different files when changing a signature.
@dermot-hardy what do you think? I reversed to match the current configuration in order to minimize the changes, but kept the modification in a separate branch if we want to put them in.

xbreizh:
> We should also test this with some data populated in the jobservice tables.

Testing has been performed as follow:

- create an empty jobservice database
- create then run a job-service-db fat jar with liquibase on V0
- create then run the flyway version of job-service-db fat jar
- check for errors then drop the database

The operation has been repeated using the different Liquibase versions
from version V3.5, an insert script has also been run in order to populate data into the different tables and check that there is no issue with it.

job-service-container/src/main/startup.d/setup-database.sh
```diff
@@ -17,7 +17,13 @@
...
 
 cd /database
+echo "Installing Job Service database."
+
 java -jar job-service-db-installer.jar \
      -db.user "${JOB_SERVICE_DATABASE_USERNAME:-${CAF_DATABASE_USERNAME}}" \
      -db.pass "${JOB_SERVICE_DATABASE_PASSWORD:-${CAF_DATABASE_PASSWORD}}" \
-     -db.connection.url "${JOB_SERVICE_DATABASE_URL:-${CAF_DATABASE_URL}}"
+     -db.connection "${JOB_SERVICE_DATABASE_URL:-${CAF_DATABASE_URL}}" \
+     -db.name "${POSTGRES_DB:-jobservice}" \
```
dermot-hardy: Why is this changing?  What does this mean?  There is only one ReadMe update and it doesn't mention anything about a new variable?  Is the intention to pass something different to `JOB_SERVICE_DATABASE_URL` than what was passed before?  Is this a breaking change?

job-service-postgres-container/pom.xml
```diff
@@ -128,7 +128,8 @@
                                     <POSTGRES_DB>${js.database.name}</POSTGRES_DB>
                                 </env>
                                 <cmd>
-                                    ./install_job_service_db.sh -fd -db.connection jdbc:postgresql://${js.database.host}:5432 -db.name ${js.database.name} -db.user ${js.database.username} -db.pass ${js.database.password}
+                                    ./install_job_service_db.sh -fd -db.connection jdbc:postgresql://${js.database.host}:5432/ -db.name
+                                    ${js.database.name} -db.user ${js.database.username} -db.pass ${js.database.password} -log INFO
```
dermot-hardy: What's the change here?  Adding `-log INFO`?  Isn't that the default?

job-service-postgres-container/pom.xml
```diff
@@ -193,7 +194,8 @@
                                     <link>${js.database.host}</link>
                                 </links>
                                 <cmd>
-                                    ./install_job_service_db.sh -db.connection jdbc:postgresql://${js.database.host}:5432 -db.name ${js.database.name} -db.user ${js.database.username} -db.pass ${js.database.password} -log debug
+                                    ./install_job_service_db.sh -db.connection jdbc:postgresql://${js.database.host}:5432/ -db.name
+                                    ${js.database.name} -db.user ${js.database.username} -db.pass ${js.database.password} -log DEBUG
```
dermot-hardy: What's the change here?  Adding `-log INFO`?  Isn't that the default?

job-service-postgres-container/scripts/setup.sh
```diff
@@ -19,9 +19,9 @@
 echo "Installing Job Service database."
 
 ./install_job_service_db.sh \
-  -db.user $POSTGRES_USER \
-  -db.pass $POSTGRES_PASSWORD \
+  -db.user ${POSTGRES_USER} \
+  -db.pass ${POSTGRES_PASSWORD} \
```
dermot-hardy, line 25: Is this change necessary?

job-service-postgres-container/scripts/setup.sh
```diff
@@ -19,9 +19,9 @@
...
   -db.name ${POSTGRES_DB:-jobservice} \
-  -db.connection jdbc:postgresql://127.0.0.1:5432
+  -db.connection jdbc:postgresql://127.0.0.1:5432/
```
dermot-hardy: Is this change necessary?

pom.xml
```diff
@@ -92,11 +92,10 @@
...
                 <scope>import</scope>
             </dependency>
-
             <dependency>
-                <groupId>com.github.cafapi.util</groupId>
-                <artifactId>util-liquibase-installer</artifactId>
-                <version>1.20.0-288</version>
+                <groupId>com.github.cafapi.util.flywayinstaller</groupId>
+                <artifactId>util-flywayinstaller-cli</artifactId>
+                <version>1.0.0-SCMOD-15887-SNAPSHOT</version>
```
dermot-hardy: Don't forget to adjust before merging?

release-notes-5.1.0.md
```diff
@@ -4,5 +4,7 @@
 ${version-number}
 
 #### New Features
+- **SCMOD-15887**: job-service-db migrated from Liquibase to Flyway. If the database does not exist or is empty, it will be created, then upgraded to the latest version.
```
dermot-hardy: The second sentence doesn't sound like a new feature; it sounds like what it did before.

dermot-hardy: We need to try to improve the quality of the release notes.

These are the two examples of minor version updates I gave in the release tasks jira ([SCMOD-15147](https://portal.digitalsafe.net/browse/SCMOD-15147)):
- https://github.com/CAFapi/opensuse-java8-images/releases/tag/v1.1.0
- https://github.com/WorkerFramework/worker-framework/releases/tag/v1.6.0

Note that in all cases there is a few words headline and then an explanatory paragraph underneath.  That is the quality that I expect release notes to be.  It is not an unreasonably high standard.

worker-jobtracking-container/src/main/startup.d/setup-database.sh
```diff
@@ -20,4 +20,6 @@ cd /database
 java -jar job-service-db-installer.jar \
      -db.user "${JOB_SERVICE_DATABASE_USERNAME:-${JOB_DATABASE_USERNAME}}" \
      -db.pass "${JOB_SERVICE_DATABASE_PASSWORD:-${JOB_DATABASE_PASSWORD}}" \
-     -db.connection.url "${JOB_SERVICE_DATABASE_URL:-${JOB_DATABASE_URL}}"
+     -db.connection "${JOB_SERVICE_DATABASE_URL:-${CAF_DATABASE_URL}}" \
+     -db.name "${POSTGRES_DB:-jobservice}" \
+     -log DEBUG
```
dermot-hardy: Why is this changing?  What does this mean?  There is only one ReadMe update and it doesn't mention anything about a new variable?  Is the intention to pass something different to `JOB_SERVICE_DATABASE_URL` than what was passed before?  Is this a breaking change?

job-service-db/src/main/resources/db/migration/V2__job_service.sql
```diff
@@ -0,0 +1,54 @@
...
+ *  Otherwise, the Liquibase specific tables get dropped (databasechangelog, databasechangeloglock)
+ *
+ */
+DO
+$$
+    BEGIN
+
+        IF EXISTS
+            (
+            -- Checks if liquibase table exists
```
dermot-hardy: Indent

job-service-db/src/main/resources/db/migration/V2__job_service.sql
```diff
@@ -0,0 +1,54 @@
...
+                SELECT table_name
```
dermot-hardy: You're only interested in whether or not the row exists.
```suggestion
                SELECT NULL
```

job-service-db/src/main/resources/db/migration/V2__job_service.sql
```diff
@@ -0,0 +1,54 @@
...
+                FROM information_schema.tables
+                WHERE table_name = 'databasechangelog'
+            )
+        THEN
+            IF NOT EXISTS(
```
dermot-hardy:
```suggestion
            IF NOT EXISTS (
```

job-service-db/src/main/resources/db/migration/V2__job_service.sql
```diff
@@ -0,0 +1,54 @@
...
+                    -- Checks that the db checksum is valid
+                    -- The checksum value is from Liquibase V3.0
+                    SELECT *
```
dermot-hardy: Again, you're only interested in whether or not the row exists.
```suggestion
                    SELECT NULL
```

job-service-db/src/main/resources/db/migration/V2__job_service.sql
dermot-hardy: You have two `IF`s embedded.  Why not use `AND`?  Or do you have to do it like this because the `databasechangelog` table may not exist?

job-service-db/src/main/resources/db/migration/V2__job_service.sql
```diff
@@ -0,0 +1,54 @@
...
+                    FROM databasechangelog
+                    WHERE md5sum = '7:6127825a6ab9fb3ed7f2c25e2354f2ce'
+                )
+            THEN
+                RAISE EXCEPTION 'The databasechangelog table is present but does not contain the expected md5sum. Cannot perform migration from Liquibase to Flyway.';
```
dermot-hardy: Line goes over the 138-character guideline.

job-service-db/src/main/resources/db/migration/procedures/R___old.sql
dermot-hardy: > If I understand correctly, we are at present depending on the filenames to assure the order in which the scripts are executed.

We shouldn't be dependent on this.  If we're going to use repeatable migrations then they should be authored so that they can run in any order.  (Otherwise we may as well be using versioned migrations.)

dermot-hardy: I don't know if there's a cleaner way but I think the way I would handle this is to include forward declarations in the modules that use other functions.  The example you gave above is that `internal_does_table_exist` makes use of `internal_to_regclass` internally.  I would declare (not define) the `internal_to_regclass` function in the `R__doesTableExist.sql` file, before the definition of the `internal_does_table_exist` function - i.e. using something like this:

```
DO $$
BEGIN
    CREATE FUNCTION internal_to_regclass(rel_name VARCHAR(63))
    RETURNS regclass
    LANGUAGE plpgsql STABLE
    AS 'BEGIN /* Forward Declaration */ END';
EXCEPTION WHEN duplicate_function THEN
END $$;
```

That way the order that the repeatable migrations are run in should not matter, and if either implementation changes in the future it should not matter either.

job-service-db/src/main/resources/db/migration/V2__job_service.sql
```diff
@@ -0,0 +1,54 @@
+--
```
dermot-hardy: I presume there's a good reason for starting with V2?  I think we discussed this before and there was.  Remind me what it is.

job-service-db/src/main/resources/db/migration/V3__job_service.sql
```diff
@@ -0,0 +1,233 @@
...
+
+CREATE INDEX IF NOT EXISTS idx_partition_id_and_job_id
+    ON public.stowed_task
+        USING btree (partition_id, job_id);
+
+CREATE INDEX IF NOT EXISTS idx_job_partition_id_and_dependent_job_id
+    ON public.job_dependency
+        USING btree (partition_id, dependent_job_id);
+
+CREATE OR REPLACE FUNCTION internal_to_regclass(rel_name VARCHAR(63))
```
dermot-hardy: I just made a [comment](https://github.com/JobService/job-service/pull/273#discussion_r771738095) about adding forward declarations to the modules where functions are used but I guess an alternative would be to forward declare in the versioned migrations.  If we are going to do that then we'd need to declare all internal functions here and not just this one, and also don't put the real implementations in, since we don't want it in two places, and you'd use a create if not exists syntax as with everything else in this migration file.

dermot-hardy (changes requested):

job-service-db/src/main/resources/db/migration/V2__job_service.sql
xbreizh: The first script is for updating an existing Flyway database. As we are just initializing it, we start at V2. Similar to https://github.houston.softwaregrp.net/caf/documentid-service/tree/develop/documentid-service-db/src/main/resources/db/migration/common

job-service-postgres-container/scripts/setup.sh
xbreizh: Removed

job-service-postgres-container/pom.xml
xbreizh: It's the default, I left it here so it's easier to see that we can actually influence that log level. Removed

job-service-postgres-container/pom.xml
xbreizh: It's the default, I left it here so it's easier to see that we can actually influence that log level. Removed

job-service-postgres-container/scripts/setup.sh
xbreizh: Yes, to match the expected patter in the flyway utility as we expect the format `jdbc:postgresql://localhost:3307/` 
Otherwise, I can update the utility to update the url when required

release-notes-5.1.0.md
xbreizh: Updated

job-service-db/src/main/resources/db/migration/V2__job_service.sql
xbreizh: Done

job-service-db/src/main/resources/db/migration/V2__job_service.sql
xbreizh: Done

job-service-db/src/main/resources/db/migration/V2__job_service.sql
xbreizh: Done

job-service-db/src/main/resources/db/migration/V2__job_service.sql
xbreizh: Done

job-service-db/src/main/resources/db/migration/V2__job_service.sql
xbreizh: Corrected

job-service-db/src/main/resources/db/migration/V2__job_service.sql
xbreizh: yes, otherwise we get `SQL Error [42P01]: ERROR: relation "databasechangelog" does not exist
  Where: PL/pgSQL function inline_code_block line 4 at IF`

job-service-db/src/main/resources/db/migration/V3.0.0/V3__job_service.sql
```diff
@@ -0,0 +1,690 @@
...
+    in_ignore_status BOOLEAN
+);
+CREATE OR REPLACE FUNCTION internal_to_regclass(rel_name VARCHAR(63))
+    RETURNS regclass
+    LANGUAGE plpgsql STABLE
+AS $$
+BEGIN
+    SELECT NULL;
+END
+$$;
```
dermot-hardy: When I suggested to create forward declarations in [this](https://github.com/JobService/job-service/pull/273#discussion_r771738095) comment I suggested to write them like this:
```
DO $$
BEGIN
    CREATE FUNCTION internal_to_regclass(rel_name VARCHAR(63))
    RETURNS regclass
    LANGUAGE plpgsql STABLE
    AS 'BEGIN /* Forward Declaration */ END';
EXCEPTION WHEN duplicate_function THEN
END $$;
```

I didn't write it but I was thinking to myself that defining forward declarations like that would allow you to easily detect if any were not properly defined later using a statement like this:
```
SELECT *
FROM information_schema.routines
WHERE routine_definition = 'BEGIN /* Forward Declaration */ END'
```

But you've instead opted to define them like this:
```
CREATE OR REPLACE FUNCTION internal_to_regclass(rel_name VARCHAR(63))
    RETURNS regclass
    LANGUAGE plpgsql STABLE
AS $$
BEGIN
    SELECT NULL;
END
$$;
```

Using `CREATE OR REPLACE FUNCTION` worries me.  I'm not sure why I'd want to replace a legitimate function with a forward declaration.  Why have you not written the forward declarations the way I suggested?  There is no comment in the PR explaining a problem with what I advised you to do.

job-service-container/src/main/startup.d/setup-database.sh
```diff
@@ -17,7 +17,14 @@
...
 
 cd /database
+echo "Installing Job Service database."
+
 java -jar job-service-db-installer.jar \
-     -db.user "${JOB_SERVICE_DATABASE_USERNAME:-${CAF_DATABASE_USERNAME}}" \
-     -db.pass "${JOB_SERVICE_DATABASE_PASSWORD:-${CAF_DATABASE_PASSWORD}}" \
-     -db.connection.url "${JOB_SERVICE_DATABASE_URL:-${CAF_DATABASE_URL}}"
+     -db.host "${JOB_SERVICE_DATABASE_SERVER:-localhost}" \
+     -db.port "${JOB_SERVICE_DATABASE_PORT:-5432}" \
```
dermot-hardy: So I understand from this change that `JOB_SERVICE_DATABASE_URL` is being separated out and the user is expected to supply the relevant segments using separate environment variables.  **This is a breaking change and yet it has not been called out in the release notes breaking changes section.**  If the project didn't already have breaking changes then this would also necessitate it moving to the next major version number.  Ironically the fact that it does already have breaking changes actually makes this slighter harder to roll out.  Rather than being able to simply change the major version number we'll first have to lock down all consuming projects to the latest commit before merging this.  Please discuss this with me before merging this PR.

job-service-container/src/main/startup.d/setup-database.sh
```diff
@@ -17,7 +17,14 @@
 
 
 cd /database
+echo "Installing Job Service database."
+
 java -jar job-service-db-installer.jar \
-     -db.user "${JOB_SERVICE_DATABASE_USERNAME:-${CAF_DATABASE_USERNAME}}" \
-     -db.pass "${JOB_SERVICE_DATABASE_PASSWORD:-${CAF_DATABASE_PASSWORD}}" \
-     -db.connection.url "${JOB_SERVICE_DATABASE_URL:-${CAF_DATABASE_URL}}"
+     -db.host "${JOB_SERVICE_DATABASE_SERVER:-localhost}" \
```
dermot-hardy: I pointed you at https://github.com/CAFapi/opensuse-base-image#pre-installed-utility-scripts just a few days ago and said these are our standard suffixes.  If you're updating to them then why have you chosen to use `JOB_SERVICE_DATABASE_SERVER` instead of `JOB_SERVICE_DATABASE_HOST`?

job-service-container/src/main/startup.d/setup-database.sh
```diff
@@ -17,7 +17,14 @@
...
+     -db.port "${JOB_SERVICE_DATABASE_PORT:-5432}" \
+     -db.user "${JOB_SERVICE_DATABASE_USERNAME:-postgres}" \
+     -db.pass "${JOB_SERVICE_DATABASE_PASSWORD:-root}" \
```
dermot-hardy: So as an example you're changing from:
`-db.user "${JOB_SERVICE_DATABASE_USERNAME:-${JOB_DATABASE_USERNAME}}"`
to:
`-db.user "${JOB_SERVICE_DATABASE_USERNAME:-postgres}"`

and similar for the others.

Two things:
1. I'm fine with dropping support for `JOB_DATABASE_USERNAME` etc. as fallback environment variables but that is **another breaking change that hasn't been called out in the release notes**.  How will someone who was happily using `JOB_DATABASE_USERNAME` know to update to `JOB_SERVICE_DATABASE_USERNAME`?
2. You're introducing `postgres` etc. as hard-coded defaults.  I'm not comfortable with that.  I'd rather these were just mandatory; no defaults.

job-service-container/src/main/startup.d/setup-database.sh
```diff
@@ -17,7 +17,14 @@
...
+     -db.name "${JOB_SERVICE_DATABASE_NAME:-jobservice}" \
+     -log DEBUG
```
dermot-hardy: Don't specify `-log` by default so that it uses whatever log level has been set by the environment variable.

job-service-postgres-container/pom.xml
dermot-hardy: I don't understand you saying it's removed.  I still see it.

job-service-postgres-container/scripts/setup.sh
```diff
@@ -19,9 +19,11 @@
...
 ./install_job_service_db.sh \
-  -db.user $POSTGRES_USER \
-  -db.pass $POSTGRES_PASSWORD \
+  -db.host ${POSTGRES_SERVER:-localhost} \
+  -db.port ${POSTGRES_PORT:-5432} \
+  -db.user ${POSTGRES_USER:-postgres} \
+  -db.pass ${POSTGRES_PASSWORD:-root} \
   -db.name ${POSTGRES_DB:-jobservice} \
-  -db.connection jdbc:postgresql://127.0.0.1:5432
+  -LOG DEBUG
```
dermot-hardy: Do we want to hard-code the log level here?  Shouldn't the consumer choose that by setting the `CAF_LOG_LEVEL` environment variable?

release-notes-6.0.0.md
```diff
@@ -6,6 +6,8 @@ ${version-number}
 #### New Features
 - **359648** taskData no longer sent as Base64.  
     taskData is no longer Base64 encoded in order to reduce message size. 
+- **353257**: Flyway replaces Liquibase for database migrations
+    The job-service-db is now using Flyway for its versioning and migrations.
```
dermot-hardy, line 10: This is how this is rendering:

image

You need to put two spaces after migrations to create a line break, like in the example just above it.

worker-jobtracking-container/src/main/startup.d/setup-database.sh
dermot-hardy: So this has changed again since the last review but the underlying points still haven't been addressed:
- changes to how the worker is configured but no documentation about the changes
  - the [docs](https://github.com/JobService/job-service/tree/develop/docs) project which builds the documentation site hasn't been changed at all
(perhaps nothing in it is effected - I haven't checked)
  - the [deploy](https://github.com/JobService/job-service/tree/develop/deploy) project which builds the [deploy](https://github.com/JobService/job-service/tree/deploy) branch and is supposed to demonstrate how to deploy the Job Service hasn't been changed at all
(it will definitely be effected)
- breaking changes not called out in release notes

job-service-container/pom.xml
```diff
@@ -249,6 +248,8 @@
                         <CAF_WEBSERVICE_URL>${dummyUrl}</CAF_WEBSERVICE_URL>
                         <JOB_SERVICE_DATABASE_APPNAME>Job Service</JOB_SERVICE_DATABASE_APPNAME>
                         <JOB_SERVICE_DATABASE_URL>jdbc:postgresql://${docker.host.address}:${postgres.db.port}/${db.name}</JOB_SERVICE_DATABASE_URL>
+                        <JOB_SERVICE_DATABASE_SERVER>${docker.host.address}</JOB_SERVICE_DATABASE_SERVER>
+                        <JOB_SERVICE_DATABASE_PORT>${postgres.db.port}</JOB_SERVICE_DATABASE_PORT>
```
dermot-hardy: If we're adding new variables specifying the host, port, db-name separately then why do we still have the `JOB_SERVICE_DATABASE_URL` specified here as well.  Do we now need both?

job-service-container/pom.xml
```diff
@@ -502,8 +503,11 @@
                                     <CAF_CONFIG_PATH>/config</CAF_CONFIG_PATH>
                                     <JOB_SERVICE_DATABASE_APPNAME>Job Service</JOB_SERVICE_DATABASE_APPNAME>
                                     <JOB_SERVICE_DATABASE_URL>jdbc:postgresql://jobservice-integrationtests-postgres:5432/${db.name}</JOB_SERVICE_DATABASE_URL>
```
dermot-hardy: Same.  Why do we still have this?

job-service-container/pom.xml
```diff
@@ -502,8 +503,11 @@
...
+                                    <JOB_SERVICE_DATABASE_SERVER>jobservice-integrationtests-postgres</JOB_SERVICE_DATABASE_SERVER>
+                                    <JOB_SERVICE_DATABASE_PORT>5432</JOB_SERVICE_DATABASE_PORT>
                                     <JOB_SERVICE_DATABASE_USERNAME>postgres</JOB_SERVICE_DATABASE_USERNAME>
                                     <JOB_SERVICE_DATABASE_PASSWORD>root</JOB_SERVICE_DATABASE_PASSWORD>
+                                    <JOB_SERVICE_DATABASE_NAME>jobservice</JOB_SERVICE_DATABASE_NAME>
```
dermot-hardy: In `JOB_SERVICE_DATABASE_URL` this was `${db.name}` so why is hardcoded to "jobservice" here instead of being what it was in the url?

job-service-acceptance-tests/pom.xml
```diff
@@ -469,8 +469,11 @@
                                     <CAF_CONFIG_PATH>/config</CAF_CONFIG_PATH>
                                     <JOB_SERVICE_DATABASE_APPNAME>Job Service</JOB_SERVICE_DATABASE_APPNAME>
                                     <JOB_SERVICE_DATABASE_URL>${job.database.url}</JOB_SERVICE_DATABASE_URL>
```
dermot-hardy: If we're splitting this up then do we still need to be passing this and do we still need the `${job.database.url}` property?

job-service-acceptance-tests/pom.xml
```diff
@@ -528,8 +531,11 @@
                                     <CAF_WORKER_THREADS>1</CAF_WORKER_THREADS>
                                     <JOB_SERVICE_DATABASE_APPNAME>Job Tracking Worker</JOB_SERVICE_DATABASE_APPNAME>
                                     <JOB_SERVICE_DATABASE_URL>${job.database.url}</JOB_SERVICE_DATABASE_URL>
```
dermot-hardy: Same question.

worker-jobtracking-container/pom.xml
```diff
@@ -464,8 +464,11 @@
                                     <CAF_WORKER_MAX_BATCH_TIME>1000</CAF_WORKER_MAX_BATCH_TIME>
                                     <CAF_WORKER_THREADS>1</CAF_WORKER_THREADS>
                                     <JOB_SERVICE_DATABASE_URL>${job.database.url}</JOB_SERVICE_DATABASE_URL>
```
dermot-hardy: If we're splitting this up then do we still need to be passing this and do we still need the `${job.database.url}` property?

job-service-postgres-container/pom.xml
```diff
@@ -32,6 +32,7 @@
     </parent>
     
     <properties>
+        <js.database.server>127.0.0.1</js.database.server>
```
dermot-hardy: Why is this needed?

job-service-postgres-container/pom.xml
```diff
@@ -102,6 +103,8 @@
                                     <port>${postgres.db.port}:5432</port>
                                 </ports>
                                 <env>
+                                    <POSTGRES_SERVER>${js.database.server}</POSTGRES_SERVER>
+                                    <POSTGRES_PORT>${postgres.db.port}</POSTGRES_PORT>
```
dermot-hardy: `postgres.db.port` is the external port so this doesn't seem right but why are these two variables that were not required before being added at all?

job-service-postgres-container/pom.xml
```diff
@@ -120,7 +123,7 @@
                         <!-- Run old version of jobservice postgres container against postgres database to install jobservice schema. -->
                         <image>
                             <alias>job-service-postgres-old</alias>
-                            <name>${dockerHubPublic}/jobservice/job-service-postgres:3.0</name>
+                            <name>${dockerHubPublic}/jobservice/job-service-postgres:2</name>
```
dermot-hardy, line 123: Was this a conscious decision to test starting from an older version?

job-service-postgres-container/pom.xml
dermot-hardy: Why is this line changing at all?  Was the syntax different in `:2` to what is used in `:3.0`?

worker-jobtracking-container/src/main/startup.d/setup-database.sh
dermot-hardy: Also same comments as the other `setup-database.sh` about defaults, hard-coded log level, etc.

release-notes-6.0.0.md
```diff
@@ -6,6 +6,8 @@ ${version-number}
...
 
 #### Breaking Changes
```
dermot-hardy, line 12: As discussed in other comments there appear to be a series of breaking changes not called out here.

job-service-db/README.md
```diff
@@ -4,39 +4,43 @@ This is the database schema and provisioning tool for the [Job Service](../job-s
...
-The database is installed using the Liquibase database change management tool and you can choose to install from the database artifact directly or you can use the Docker image that is  also available.
+The database is installed using the Flyway database change management tool and you can choose to install from the database artifact 
+directly or you can use the Docker image that is  also available.
 
 ### PostgreSQL
 The Job Service database requires PostgreSQL 9.4 or later to be installed and configured. 
 
 ### Database Artifact
-With PostgreSQL 9.4 or later installed and configured, download the [job-service-db-2.X.X-XXX-jar-with-dependencies](https://repo1.maven.org/maven2/com/github/jobservice/job-service-db/) jar from Artifactory and then run:
+With PostgreSQL 11 or later installed and configured, download the [job-service-db-2.X.X-XXX-jar-with-dependencies](https://repo1.maven.org/maven2/com/github/jobservice/job-service-db/) jar from Artifactory and then run:
```
dermot-hardy:
```suggestion
With PostgreSQL 11 or later installed and configured, download the [job-service-db-X.X.X-XXX-jar-with-dependencies](https://repo1.maven.org/maven2/com/github/jobservice/job-service-db/) jar from Artifactory and then run:
```

job-service-db/README.md
```diff
@@ -4,39 +4,43 @@ This is the database schema and provisioning tool for the [Job Service](../job-s
...
 
-	java -jar /job-service-db-2.3.0-159-jar-with-dependencies.jar -db.connection jdbc:postgresql://localhost:5432 -db.name jobservice -db.pass root -db.user postgres -fd
+	java -jar /job-service-db-2.3.0-159-jar-with-dependencies.jar -db.host localhost -db.port 5432 -db.name jobservice -db.pass root -db.user postgres
```
dermot-hardy: The jar version number can't be right for these new options.

job-service-db/README.md
```diff
@@ -4,39 +4,43 @@ This is the database schema and provisioning tool for the [Job Service](../job-s
...
 
-the database connection, user and password string arguments will need changed to match your PostgreSQL 9.4 or later setup.
+The database connection, user and password string arguments will need changed to match your PostgreSQL 9.4 or later setup.
```
dermot-hardy: You adjusted to PostgreSQL 11 above but still say 9.4 here.

job-service-db/README.md
```diff
@@ -4,39 +4,43 @@ This is the database schema and provisioning tool for the [Job Service](../job-s
...
+With PostgreSQL 11 or later installed and configured, pull the installer image from artifactory. For example:
 
 	docker pull jobservice/job-service-postgres:2.0.0
 
 then run the image using:
 
 	docker run --rm jobservice/job-service-postgres:2.0.0 \
 	./install_job_service_db.sh \
-	-db.connection jdbc:postgresql://<postgres host>:5432/ \
+	-db.host localhost \
```
dermot-hardy:
```suggestion
	-db.host <postgres host> \
```

job-service-db/README.md
```diff
@@ -4,39 +4,43 @@ This is the database schema and provisioning tool for the [Job Service](../job-s
...
 This is available as a Docker container - see [job-service-postgres-container](../job-service-postgres-container).
 
-With PostgreSQL 9.4 or later installed and configured, pull the installer image from artifactory. For example:
+With PostgreSQL 11 or later installed and configured, pull the installer image from artifactory. For example:
 
 	docker pull jobservice/job-service-postgres:2.0.0
 
 then run the image using:
 
 	docker run --rm jobservice/job-service-postgres:2.0.0 \
```
dermot-hardy: Drop the version number here (and in the instruction to pull above):
```suggestion
	docker run --rm jobservice/job-service-postgres \
```

job-service-db/src/main/resources/db/migration/README.md
```diff
@@ -0,0 +1,34 @@
+# Types of functions
```
dermot-hardy: I very much like that you've added this ReadMe but adding it here (anywhere inside `/src/main/resources/`) will cause it to actually be included in the jar, which is probably not what we want.  Move it up to the `job-service-db` level and rename it to sit alongside the existing ReadMe.

job-service-db/src/main/resources/db/migration/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+There are 2 types of functions that we currently use: internal and external. We use _function_ as a general term for both function and 
+procedures here.
+
+The internal function's name should start with 'internal_'. Ex: internal_has_dependent_jobs()
```
dermot-hardy: Use backticks for code words and phrases:
```suggestion
The internal function's name should start with 'internal_'. Ex: `internal_has_dependent_jobs()`
```

job-service-db/src/main/resources/db/migration/README.md
```diff
@@ -0,0 +1,34 @@
...
+      in_job_id VARCHAR(48)
+    )
+    RETURNS VOID
+    LANGUAGE plpgsql VOLATILE
+    AS $$
+    BEGIN
+    END
+    $$;
+
+The function would be empty and return a dummy value when required. The real implementation would occur in the script corresponding to
```
dermot-hardy: "return a dummy value when required":
Is that actually required?

job-service-db/src/main/resources/db/migration/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Updating a function requiring a Change of signature
+
+If we need to change the function signature, then we need to drop the old one. To proceed, we add the drop function element into the 
+main migration script, then update the function in the corresponding script.
+
+## Updating an internal function that requires a previously declared function that is yet to be created
+
+In the case where we need to use an inner function that hasn't yet be created, we would add it to the main migration script and make 
+sure that it's placed before the function using it.
```
dermot-hardy: I'm not sure what you mean here.  Surely we just forward declare the internal function in the current migration script as we will for all internal functions?  I don't think I understand the extra rule you're calling out here (i.e. I don't understand the "make sure that it's placed before the function using it" bit).

job-service-db/src/main/resources/db/migration/V2.0.0/V2__job_service.sql
```diff
@@ -0,0 +1,54 @@
+--
```
dermot-hardy: I think you had these files directly under `migration` before.  Why are the folders introduced?  We now have the version number in both the folder and the file?  Will the minor and patch versions always be 0?

job-service-db/src/main/resources/db/migration/V3.0.0/V3__job_service.sql
```diff
@@ -0,0 +1,690 @@
...
+    ON public.stowed_task
+        USING btree (partition_id, job_id);
+
+CREATE INDEX IF NOT EXISTS idx_job_partition_id_and_dependent_job_id
+    ON public.job_dependency
+        USING btree (partition_id, dependent_job_id);
+
+/*
+ *
+ * Drop unused functions and forward declare internal functions
```
dermot-hardy: Do you think it would be worth starting a V4 file for these, keeping the V3 with just the base tables, indexes, etc.?

job-service-db/src/main/resources/db/migration/V3.0.0/V3__job_service.sql
```diff
@@ -0,0 +1,690 @@
...
+ *
+ */
+
+DROP FUNCTION IF EXISTS cancel_job(
+    in_partition_id VARCHAR(40),
+    in_job_id VARCHAR(48),
+    in_short_job_id VARCHAR(48)
+);
+CREATE OR REPLACE FUNCTION internal_cleanup_completed_subtask_report(
```
dermot-hardy: Would it be tidier to do all the drops first and then to do the forward declarations rather than mixing them?  Maybe the drops go in a V4 and the forward declares in a V5?  (Of course I understand that going forward they will be mixed chronologically as changes take place but would it be tidier at this starting point?)

dermot-hardy (changes requested):

job-service-db/src/main/resources/db/migration/V3.0.0/V3__job_service.sql
xbreizh: Yes, unsure of the V3, V4, V5, maybe more  `V3__job_service_database_schema.sql` ,   `V3__job_service_functions.sql`, ... Let me know

job-service-db/src/main/resources/db/migration/V3.0.0/V3__job_service.sql
xbreizh: Yes, I thought of that, but should we keep the version aligned with the release versions, here we do a single migration. We could also go with `V3__job_service_database_schema.sql` and  `V3__job_service_functions.sql`.
 This is just for the liquibase -> flyway transition so I guess we can do anything. The splitting would be easier to read for sure. At the moment, I have 3 sections: db schema, drop functions, create internal functions.

job-service-db/src/main/resources/db/migration/V2.0.0/V2__job_service.sql
xbreizh: removed the folders. I wanted to match what was done on liquibase, but realized that it was done that way as each folder would contain a similarly named file `changelog.xml`

job-service-db/src/main/resources/db/migration/README.md
xbreizh: True, shouldn't be required, that's what we re trying to avoid with fwd declarations.

job-service-db/src/main/resources/db/migration/README.md
xbreizh: Done

job-service-db/README.md
xbreizh: Done

job-service-db/README.md
xbreizh: Done

job-service-db/README.md
xbreizh: Done

job-service-db/README.md
xbreizh: Done

job-service-postgres-container/pom.xml
xbreizh: Yes, as the migration script has been adapted to any version (we wanted to avoid a breaking change back then), we re trying to check it from the oldest possible version (couldn't find 1.0)

release-notes-6.0.0.md
xbreizh: Fixed

job-service-postgres-container/scripts/setup.sh
xbreizh: Dropping

job-service-container/src/main/startup.d/setup-database.sh
xbreizh: Dropped

job-service-container/src/main/startup.d/setup-database.sh
xbreizh: Replaced

job-service-postgres-container/pom.xml
xbreizh: Restored

job-service-container/src/main/startup.d/setup-database.sh
xbreizh: - removed variables that differ from https://github.com/CAFapi/opensuse-base-image#pre-installed-utility-scripts
- documented the release notes
- removed the default
- added check for notNull

job-service-container/src/main/startup.d/setup-database.sh
xbreizh: - removed variables that differ from https://github.com/CAFapi/opensuse-base-image#pre-installed-utility-scripts
- documented the release notes
- removed the default
- added check for notNull

job-service-container/pom.xml
xbreizh: Removed

job-service-container/pom.xml
xbreizh: Removed

job-service-container/pom.xml
xbreizh: Fixed

job-service-acceptance-tests/pom.xml
xbreizh: Removed

worker-jobtracking-container/pom.xml
xbreizh: Removed

job-service-postgres-container/pom.xml
xbreizh: Removed

job-service-postgres-container/pom.xml
xbreizh: Removed

release-notes-6.0.0.md
xbreizh: Updated the release notes with the breaking changes

job-service-db/README.md
xbreizh: Updated

job-service-db/src/main/resources/db/migration/README.md
xbreizh: As discussed, depending on the sql language, a value for return is required

job-service-db/src/main/resources/db/migration/README.md
xbreizh: Done

xbreizh:
@dermot-hardy  Hi, there are quite a lot of changes since last review. Those involve:

- restrict job-service database variables to https://github.com/CAFapi/opensuse-base-image#pre-installed-utility-scripts
- I specify those in the [release notes](https://github.com/JobService/job-service/blob/b0141489894cebd264818d1baa907107b0649cc0/release-notes-6.0.0.md)
- As we are not passing the full database url, but are still only expecting postgres, I replaced Properties objects used for the connection with PGSimpleDataSource objects.
- As we don't have any alternative if missing variable now we accept only a single form of variable, I added a check in form of Objects.requireNonNull() to make sure we raise an issue in a fast manner.

It's possibly best if I take you through those, or maybe it's too much change for here, please let me know if you expected something else. Thx

dermot-hardy (approved):

PR #287, opened 2022-02-04
US397417: Job Service: Restore TaskMessage option

xbreizh:
https://internal.almoctane.com/ui/?p=131002%2F6001#/entity-navigation?entityType=work_item&id=397417

### DIFF
--- a/job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/QueueServices.java
+++ b/job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/QueueServices.java
@@ -72,7 +72,6 @@ public final class QueueServices implements AutoCloseable
         //  Generate a random task id.
         LOG.debug("Generating task id ...");
         final String taskId = UUID.randomUUID().toString();
-        final boolean useNewQueueMessageFormat = ScheduledExecutorConfig.useNewQueueMessageFormat();
         final Object taskMessage;
 
         //  Set up string for statusCheckUrl
@@ -89,7 +88,8 @@ public final class QueueServices implements AutoCloseable
                 getStatusCheckIntervalMillis(ScheduledExecutorConfig.getStatusCheckIntervalSeconds()),
                 statusCheckUrl, ScheduledExecutorConfig.getTrackingPipe(), workerAction.getTargetPipe());
 
-        if (useNewQueueMessageFormat) {
+        final String messageFormatVersion = ScheduledExecutorConfig.useNewQueueMessageFormat();
+        if ("v4".equalsIgnoreCase(messageFormatVersion)) {
             taskMessage = getQueueTaskMessage(trackingInfo, workerAction, taskId);
         } else {
             taskMessage = getTaskMessage(trackingInfo, workerAction, taskId);
--- a/job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/QueueServices.java
+++ b/job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/QueueServices.java
@@ -77,9 +77,9 @@ public final class QueueServices implements AutoCloseable
 
         //  Set up string for statusCheckUrl
         final String statusCheckUrl = UriBuilder.fromUri(ScheduledExecutorConfig.getWebserviceUrl())
-                .path("partitions").path(partitionId)
-                .path("jobs").path(jobId)
-                .path("status").build().toString();
+            .path("partitions").path(partitionId)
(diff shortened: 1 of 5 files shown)


job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/ScheduledExecutorConfig.java
```diff
@@ -64,6 +68,10 @@ public static String getTrackingPipe() {
         return getPropertyOrEnvVar("CAF_TRACKING_PIPE");
     }
 
+    public static Boolean useNewQueueMessageFormat() {
+        return Boolean.parseBoolean(getPropertyOrEnvVar("JOB_SERVICE_USE_NEW_QUEUE_MESSAGE_FORMAT"));
```
xbreizh: This will return false if not present or anything different from "TRUE".

job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/QueueServices.java
```diff
@@ -106,15 +89,11 @@ public void sendMessage(
...
-        final QueueTaskMessage taskMessage = new QueueTaskMessage(
-                taskId,
-                workerAction.getTaskClassifier(),
-                workerAction.getTaskApiVersion(),
-                taskData,
-                TaskStatus.NEW_TASK,
-                Collections.<String, byte[]>emptyMap(),
-                targetQueue,
-                trackingInfo);
+        if (useNewQueueMessageFormat) {
```
xbreizh: Using the new variable to switch between message formats

xbreizh:
Hi @dermot-hardy Here, I simply extracted the conversion into taskMessage object into a method then split it to return either TaskMessage or QueueTaskMessage. That decision is made according to the value returned by the variable _JOB_SERVICE_USE_NEW_QUEUE_MESSAGE_FORMAT_ (I couldn't find a better wording but at least it should be descriptive enough). I kept changes to the minimum and haven't touched the tests, only added the variable with TRUE so that they keep on working with the new format. Testing on larry went fine also.

dermot-hardy (approved):
This looks good.  I'm not sure about `JOB_SERVICE_USE_NEW_QUEUE_MESSAGE_FORMAT` as the name of the environment variable though.  It will date if we improve for the format and introduce an even newer one.  Could we change it to `JOB_SERVICE_MESSAGE_OUTPUT_FORMAT=v4`?  And make `v3` the default format if no format is specified?

job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/QueueServices.java
```diff
@@ -106,15 +88,12 @@ public void sendMessage(
...
-        final QueueTaskMessage taskMessage = new QueueTaskMessage(
-                taskId,
-                workerAction.getTaskClassifier(),
-                workerAction.getTaskApiVersion(),
-                taskData,
-                TaskStatus.NEW_TASK,
-                Collections.<String, byte[]>emptyMap(),
-                targetQueue,
-                trackingInfo);
+        final String messageFormatVersion = ScheduledExecutorConfig.useNewQueueMessageFormat();
```
xbreizh, line 91: Note that I moved that next to where it's used. We could even inline it but I thought it would make code easier to read.

job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/ScheduledExecutorConfig.java
```diff
@@ -64,6 +67,11 @@ public static String getTrackingPipe() {
         return getPropertyOrEnvVar("CAF_TRACKING_PIPE");
     }
 
+    public static String useNewQueueMessageFormat() {
+        final String outPutFormatVersion = getPropertyOrEnvVar("JOB_SERVICE_MESSAGE_OUTPUT_FORMAT").toUpperCase(Locale.ROOT);
```
xbreizh, line 71: Updated with String instead of boolean with ["V3", "V4"] as accepted value and V3 as the default one.

xbreizh (commented):
@dermot-hardy  Updated with String instead of boolean with ["V3", "V4"] as accepted value and V3 as the default one.

dermot-hardy (approved):
Remove the first breaking change from the release notes as it is no longer relevant.

xbreizh:
> Remove the first breaking change from the release notes as it is no longer relevant.

Done.

PR #314, opened 2022-11-10
US593021: Updated to run on Java 17

michael-bryson:
https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=593021


dermot-hardy (changes requested):
Actually I don't think this one is a breaking change.  The worker that it includes (i.e. `worker-jobtracking`) is not a Document Worker, so it wouldn't have any Nashorn capabilities.  We don't have many non-Document Workers around so I never thought to mention this subtlety.

dermot-hardy (approved):

PR #321, opened 2022-12-06
US610049: CVE-2022-40149: Update jettison to 1.5.2

raghavendran-gopalakrishnan:
https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=610049


dermot-hardy (changes requested):
Use dependency management to override the version rather than doing it like this.

dermot-hardy (approved):

PR #329, opened 2023-03-08
US618048: updating headers

tbroadbent22:
https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=618048

### DIFF
--- a/deploy/pom.xml
+++ b/deploy/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright 2016-2022 Micro Focus or one of its affiliates.
+    Copyright 2016-2023 Open Text.
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
--- a/docs/_data/footer_links.json
+++ b/docs/_data/footer_links.json
@@ -1,7 +1,7 @@
 {
   "navigation_items": [],
   "feedback_url": "https://github.com/JobService/job-service/issues",
-  "copyright": "© 2016-2021 Micro Focus or one of its affiliates",
+  "copyright": "© Copyright 2016-2023 Open Text",
   "footer_logo": "assets/img/job-service-logo.png",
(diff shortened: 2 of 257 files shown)


frankmccarry (approved):

michael-bryson:
@frankmccarry @caoimhelynch Looks like the update of some sql files has caused [D664108](https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=664108). Normally we do not update sql files as they are excluded from the license check and any updates to them causes db migration validation issues like above. This should have failed the build but it did not so I will take a look at that.

caoimhelynch:
> @frankmccarry @caoimhelynch Looks like the update of some sql files has caused [D664108](https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=664108). Normally we do not update sql files as they are excluded from the license check and any updates to them causes db migration validation issues like above. This should have failed the build but it did not so I will take a look at that.

Makes sense, I expected any sql additions I had made to fail the build which is why this went unnoticed on my part. I will go over these and remove the changes to the sql files. Thanks Michael

michael-bryson:
> > @frankmccarry @caoimhelynch Looks like the update of some sql files has caused [D664108](https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=664108). Normally we do not update sql files as they are excluded from the license check and any updates to them causes db migration validation issues like above. This should have failed the build but it did not so I will take a look at that.
> 
> Makes sense, I expected any sql additions I had made to fail the build which is why this went unnoticed on my part. I will go over these and remove the changes to the sql files. Thanks Michael

No worries, I can see that the reason the build did not fail is because last year the db was migrated to flyway and as such the old sql files were removed for the migration. Therefore the db upgrade test was testing an upgrade from before the migration was done so changes to the sql files did not matter as that version of the db migration had not been applied. On the Moe stooge that migration had been applied before and as such failed the validation once it tried to run again.

I will update the `old` db version in the upgrade test so it picks this up.

PR #338, opened 2023-04-03
US618048: updating headers 

caoimhelynch:
https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=618048

### DIFF
--- a/job-service-internal-client/pom.xml
+++ b/job-service-internal-client/pom.xml
@@ -37,13 +37,13 @@
     <inceptionYear>2016</inceptionYear>
 
     <organization>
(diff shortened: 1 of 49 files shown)


job-service-internal-client/pom.xml
```diff
@@ -37,12 +37,12 @@
     <inceptionYear>2016</inceptionYear>
 
     <organization>
-        <name>EntIT Software LLC, an Open Text company</name>
+        <name>EntIT Software LLC, an Open Text</name>
```
michael-bryson: This doesn't make sense.

michael-bryson (approved):

PR #347, opened 2023-08-09
US749207: Remove okio

jonny-mccoubrey:
https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=749207


jonny-mccoubrey:
Where is okio dependency being consumed?

frankmccarry (changes requested):
I'm still seeing version 2.5 in use in some modules.

frankmccarry (approved):

PR #359, opened 2023-11-20
866042: Update to latest tomcat base image patch SNAPSHOT

rorytorneymf:
Move to the latest tomcat base image patch SNAPSHOT, which replaces the sed command used to replace passwords in setup-tomcat-ssl-cert.sh with awk, as the sed command failed when the password contained the @ symbol.

Note, we're using the 1.6.1 patch SNAPSHOT, which is still on tomcat 9. The latest main branch SNAPSHOT is on tomcat 10, and I don't think this repo is ready to be moved to that yet.

https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=866042


job-service-container/pom.xml
```diff
@@ -33,7 +33,7 @@
 
     <properties>
         <task.table.deletion.count>100</task.table.deletion.count>
-        <tomcatBaseImage>${dockerHubPublic}/cafapi/opensuse-tomcat-jre17:1</tomcatBaseImage>
+        <tomcatBaseImage>${dockerHubPublic}/cafapi/prereleases:opensuse-tomcat-jre17-1.7.0-860022-SNAPSHOT</tomcatBaseImage>
```
rorytorneymf: Update before merging

dermot-hardy (approved):

PR #368, opened 2023-12-22
US771133/US868107: Correctly return responses from new endpoints

jonny-mccoubrey:
https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=771133
https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=868107

### DIFF
--- a/job-service/src/main/java/com/hpe/caf/services/job/api/generated/JobsApiServiceImpl.java
+++ b/job-service/src/main/java/com/hpe/caf/services/job/api/generated/JobsApiServiceImpl.java
@@ -73,7 +73,7 @@ public class JobsApiServiceImpl extends JobsApiService {
     {
         final int successfulDeletions = JobsDelete.deleteJobs(partitionId, jobIdStartsWith, statusType, label, filter);
 
-        return Response.ok(String.format("Successfully deleted %s jobs", successfulDeletions)).build();
+        return Response.ok().entity(String.format("Successfully deleted %s jobs", successfulDeletions)).build();
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


job-service/src/main/java/com/hpe/caf/services/job/api/generated/JobsApiServiceImpl.java
```diff
@@ -89,7 +89,7 @@ public Response cancelJobs(final String partitionId, final String jobIdStartsWit
     {
 
         final int successfulCancellations = JobsCancel.cancelJobs(partitionId, jobIdStartsWith, label, filter);
-        return Response.ok(String.format("Successfully cancelled %s jobs", successfulCancellations)).build();
+        return Response.ok().entity(String.format("Successfully cancelled %s jobs", successfulCancellations)).build();
```
michael-bryson:
```suggestion
        return Response.ok().entity(String.format("Successfully canceled %s jobs", successfulCancellations)).build();
```
This should be US spelling unfortunately

andyreidz (changes requested):
Apologies for not thinking about this in the first review, let's return the number of canceled or deleted jobs instead of a string, that way it will be easier for a caller to validate the response.

jonny-mccoubrey:
This PR was initially correcting the MIME type that the two new endpoints were returning.
Initially we were returning a String response so MIME type `text/plain` was being used.

However, when we decided to return Long response instead, `text/plain` caused an error. For the Long response, we choose MIME type `application/json` and it works as expected.

jonny-mccoubrey:
NOTE: Have requested review but will wait for CI build of final commit to pass before merging

andyreidz (approved):

PR #369, opened 2024-01-05
US771133/US868107: Return JSON Objects from new endpoints

jonny-mccoubrey:
https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=868107


job-service/src/main/java/com/hpe/caf/services/job/api/generated/JobsApiServiceImpl.java
```diff
@@ -71,8 +72,12 @@ public Response deleteJobs(final String partitionId, final String jobIdStartsWit
                                final String filter, final String cAFCorrelationId, final SecurityContext securityContext)
             throws Exception
     {
-        final Long successfulDeletions = JobsDelete.deleteJobs(partitionId, jobIdStartsWith, statusType, label, filter);
-        return Response.ok().entity(successfulDeletions).build();
+        final ObjectMapper mapper = new ObjectMapper();
+        final Long value = JobsDelete.deleteJobs(partitionId, jobIdStartsWith, statusType, label, filter);
+        final Map<String, Long> resultMap = Collections.singletonMap("jobsDeleted", value);
+        final String json = mapper.writeValueAsString(resultMap);
```
dermot-hardy: This is weird looking.  Is it definitely returning a JSON object rather than a JSON-encoded string?

jonny-mccoubrey: ~~@dermot-hardy I am now returning an `ObjectNode` built using the Jackson Object Mapper. Think I was possibly just returning a JSON-encoded string before.~~

jonny-mccoubrey: Sorry, simply removing the `writeValueAsString` function means we are getting JSON object rather than string.

job-service-container/src/test/java/com/hpe/caf/services/job/api/JobServiceIT.java
```diff
@@ -419,15 +421,14 @@ public void testDeleteJobsUsingJobIdsList() throws ApiException
...
-
+        final InlineResponse2001 response = jobsApi.deleteJobs(defaultPartitionId, jobCorrelationId, null, null, filter);
 
         final List<Job> deletedJobs = jobsApi.getJobs(
                 defaultPartitionId, jobCorrelationId,null,null,null,null,null, null, filter);
 
         assertTrue(deletedJobs.isEmpty());
 
-        assertEquals(response, 3);
+        assertEquals((long) response.getJobsDeleted(), 3L);
```
andyreidz, line 431: Very minor point for the future.
Casting esponse.getJobsDeleted() to Long can cause an exception if it's null, making it slightly harder to understand why the test failed, if you cast 3L to a Long you can avoid this.

assertEquals(response.getJobsDeleted(), (Long)3L);

No need to change here.

andyreidz (approved):

dermot-hardy (approved):

PR #379, opened 2024-02-08
I879084: Return empty list instead of null when there are no failures

kusumaghoshdastidar:
https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=879084

### DIFF
--- a/job-service/pom.xml
+++ b/job-service/pom.xml
@@ -248,36 +248,6 @@
                             </scripts>
                         </configuration>
                     </execution>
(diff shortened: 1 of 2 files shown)


dermot-hardy (commented):
Unless it would cause a large amount of changes I would prefer that we deal with this in the code rather than making a dynamic adjustment to the model for it.

dermot-hardy (approved):

PR #387, opened 2024-03-15
US893094: Prevent creation of job if any of its prereq jobs failed

rorytorneymf:
https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=893094


job-service-db/src/main/resources/db/migration/functions/public/R__createJobForDependencies.sql
```diff
@@ -81,6 +83,16 @@ BEGIN
         in_delay = 0;
     END IF;
 
+    -- Raise exception if any prerequisite jobs have failed
+    SELECT ARRAY_AGG(job_id) INTO failed_prerequisite_job_ids
+    FROM job
+    WHERE job_id = ANY(in_prerequisite_job_ids)
+      AND status = 'Failed';
+
+    IF array_length(failed_prerequisite_job_ids, 1) > 0 THEN
```
rorytorneymf: Hi @dermot-hardy , I am following the same pattern as the existing validation in this file (`RAISE EXCEPTION...`), but when I invoked the code, I noticed that part of the the SQL stack trace is also included after the error message - is this a problem?

image

dermot-hardy: It's not ideal but as you rightly say this is the way all of the validations are done in this module so in my opinion you're right to do the same.

dermot-hardy (changes requested):
I can see you've placed this with the other validations but this is not like them as they are only validating the parameters.

Where it is this change will not work:

1. if the prerequisite job fails just after you've made the check (tiny window)
2. if the failure is in the completed subtask table but hasn't been processed yet (much bigger window)

If you look down the function you'll see statements that are:

1. locking the prerequisite job rows
2. processing the outstanding job updates

job-service-db/src/main/resources/db/migration/functions/public/R__createJobForDependencies.sql
```diff
@@ -138,6 +138,20 @@ BEGIN
...
+        -- If any prereqs have a 'Failed' status, the SELECT query will raise an exception and terminate this function ('create_job').
+        -- If no prereqs have a 'Failed' status, the SELECT query won't return anything.
+        UNION
+        SELECT job_id
+        FROM (SELECT uj.job_id, internal_raise_exception_for_failed_prereqs(array_agg(uj.job_id)) AS failed_prereqs
+              FROM updated_jobs uj
+              WHERE uj.partition_id = in_partition_id
+                AND uj.job_id IN (SELECT job_id FROM prereqs)
+                AND uj.status = 'Failed'
+              GROUP BY uj.job_id) AS failed_jobs
```
dermot-hardy: Let's discuss this.  I don't understand the need to introduce arrays for this, and also there are a few things I don't understand about it.

job-service-db/src/main/resources/db/migration/functions/internal/R__internal__raiseExceptionForFailedPrereqs.sql
```diff
@@ -0,0 +1,32 @@
...
+--
+
+/*
+ *  Name: internal_raise_exception_for_failed_prereqs
+ *
+ *  Description:
+ *  Accepts an array of failed prerequisite job ids and raises an exception if the array is not empty.
+ */
+CREATE OR REPLACE FUNCTION internal_raise_exception_for_failed_prereqs(in_failed_prereq_jobs_ids VARCHAR(128)[]) RETURNS BOOLEAN
+    LANGUAGE plpgsql AS
```
dermot-hardy: Can we mark this new function as [`IMMUTABLE`](https://www.postgresql.org/docs/current/xfunc-volatility.html)?

(I'm not sure if it would also need to be updated in the forward declaration.)

dermot-hardy (approved):

PR #409, opened 2024-05-21
US914106: Version Currency: Replace PowerMock

caoimhelynch:
https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=914106


job-service/src/main/java/com/hpe/caf/services/job/api/DatabaseHelperFactory.java
```diff
@@ -0,0 +1,25 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hpe.caf.services.job.api;
+
+import com.hpe.caf.services.configuration.AppConfig;
+
+public class DatabaseHelperFactory {
```
rorytorneymf: Is it possible to use Mockito to mock the constructor to avoid having to add this class and update production code?

https://www.baeldung.com/java-mockito-constructors-unit-testing#mocking-default-constructors-using-mockito

rorytorneymf (approved):

PR #417, opened 2024-06-07
US914108: Drop JUnit 4 dependency

raghavendran-gopalakrishnan:
https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=914108

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -726,11 +726,6 @@
                 <version>2.12.7</version>
             </dependency>
             <dependency>
-                <groupId>junit</groupId>
-                <artifactId>junit</artifactId>
-                <version>4.13.2</version>
-            </dependency>
-            <dependency>
                 <groupId>net.bytebuddy</groupId>
                 <artifactId>byte-buddy</artifactId>
                 <version>1.14.17</version>
--- a/pom.xml
+++ b/pom.xml
@@ -478,6 +478,12 @@
                 <groupId>com.googlecode.json-simple</groupId>
                 <artifactId>json-simple</artifactId>
                 <version>1.1.1</version>
+                <exclusions>
+                    <exclusion>
+                        <groupId>junit</groupId>
+                        <artifactId>junit</artifactId>
+                    </exclusion>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -725,11 +725,6 @@
                 <artifactId>joda-time</artifactId>
                 <version>2.12.7</version>
             </dependency>
-            <dependency>
-                <groupId>junit</groupId>
-                <artifactId>junit</artifactId>
-                <version>4.13.2</version>
-            </dependency>
```
dermot-hardy, line 732: Removing the entry from dependency management does not remove the dependency.  All it does is to remove our control over the version of the dependency that will be used.  It's not a fault in the dependency updating tool that it is adding this entry back in; the tool is effectively telling you that the dependency is not gone.

Looking at the project I can see that it is coming in as a transitive dependency of `json-simple`, which is referenced by the `job-service-acceptance-tests` project:

image

Incidentally I had thought that we were no longer using `json-simple` following the [`US915106`](https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=915106) changes that I made about a month ago.  Obviously I was relying on the dependency reports which are not reporting on the test usages.

dermot-hardy (changes requested):

pom.xml
```diff
@@ -478,6 +478,12 @@
                 <groupId>com.googlecode.json-simple</groupId>
                 <artifactId>json-simple</artifactId>
                 <version>1.1.1</version>
+                <exclusions>
+                    <exclusion>
+                        <groupId>junit</groupId>
+                        <artifactId>junit</artifactId>
+                    </exclusion>
+                </exclusions>
```
raghavgopal, line 486: I have added junit as exclusion for now in scope of this ticket. If this solves the dependency problem, can I raise a separate ticket to remove json-simple, as we need to rewrite test classes.

dermot-hardy (approved):

PR #423, opened 2024-06-12
US914157: Add health checks to check ports are started + open + accepting connections

rorytorneymf:
https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=914157

### DIFF
--- a/job-service-dropwizard/pom.xml
+++ b/job-service-dropwizard/pom.xml
@@ -105,12 +105,6 @@
         <dependency>
             <groupId>org.eclipse.jetty</groupId>
             <artifactId>jetty-server</artifactId>
-            <exclusions>
-                <exclusion>
-                    <groupId> org.eclipse.jetty.toolchain</groupId>
-                    <artifactId>jetty-jakarta-servlet-api</artifactId>
-                </exclusion>
-            </exclusions>
         </dependency>
         <dependency>
             <groupId>org.slf4j</groupId>
--- a/job-service-dropwizard/pom.xml
+++ b/job-service-dropwizard/pom.xml
@@ -103,6 +103,16 @@
             <artifactId>commons-text</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.eclipse.jetty</groupId>
+            <artifactId>jetty-server</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId> org.eclipse.jetty.toolchain</groupId>
+                    <artifactId>jetty-jakarta-servlet-api</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-api</artifactId>
         </dependency>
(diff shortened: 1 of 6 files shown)


job-service-dropwizard/src/main/java/com/hpe/caf/services/job/dropwizard/health/PortsHealthCheck.java
```diff
@@ -0,0 +1,111 @@
...
+                        final String message = getMessage(serverConnectorName, serverConnectorPort, "not open");
+                        LOGGER.error(message);
+                        return Result.unhealthy(message);
+                    }
+
+                    LOGGER.error(getMessage(serverConnectorName, serverConnectorPort, "started and open")); // TODO debug
+
+                    break;
+                case READY:
+                    if (!serverConnector.isAccepting()) {
```
rorytorneymf, line 87: I thought putting the `!serverConnector.isAccepting()` check in the readiness checks would prevent restarting the service in the case where the service was just busy with other requests. 

Looking at the issue described in https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=914157, I am not sure what state the HTTPS connector would have been in when this issue happened.

If it had been not started or closed, then these changes would mean the service would have been restarted, which is what we want.

But if it had been open but not accepting, then putting this in the readiness check and not in the liveness check means we would not be restarting the service if this issue happened again. Putting the `!serverConnector.isAccepting()` check in the liveness check means that we would restart the service if this issue happened again, but also that we may potentially be restarting the service when it is just busy with other requests (killing those requests), so I am not sure what the best thing to do here is?

job-service-dropwizard/pom.xml
```diff
@@ -102,6 +102,16 @@
...
         </dependency>
+        <dependency>
+            <groupId>org.eclipse.jetty</groupId>
+            <artifactId>jetty-server</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId> org.eclipse.jetty.toolchain</groupId>
+                    <artifactId>jetty-jakarta-servlet-api</artifactId>
+                </exclusion>
+            </exclusions>
```
dermot-hardy: There may be exceptions, but generally speaking I think it is better for exclusions to be in dependency management.

job-service-dropwizard/pom.xml
```diff
@@ -102,6 +102,16 @@
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-text</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.eclipse.jetty</groupId>
+            <artifactId>jetty-server</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId> org.eclipse.jetty.toolchain</groupId>
```
dermot-hardy: There is an extra space here.

```suggestion
                    <groupId>org.eclipse.jetty.toolchain</groupId>
```

dermot-hardy (approved):

job-service-dropwizard/src/main/java/com/hpe/caf/services/job/dropwizard/health/PortsHealthCheck.java
dermot-hardy: I'm not sure, but I suspect it was closed, so I think you're changes cover it, hopefully.

PR #451, opened 2024-10-16
634058: Batch Workers should send split messages back out to staging queues

andyreidz:
This logic is being relocated to the job service job definitions.

https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=634058


job-service-scheduled-executor/src/main/java/com/hpe/caf/services/job/scheduled/executor/QueueServicesFactory.java
```diff
@@ -97,36 +79,16 @@ public static QueueServices create(final String targetQueue, final String partit
...
 
         //setting queue properties: durable - true, exclusive - false, autoDelete - false
-        publishChannel.queueDeclare(stagingQueueOrTargetQueue, true, false, false, QUEUE_ARGUMENTS);
+        publishChannel.queueDeclare(targetQueue, true, false, false, QUEUE_ARGUMENTS);
 
         // stagingQueueOrTargetQueue = Queue where message should be published to
         // targetQueue               = Queue that should be set in the 'to' field of the task message
 
-        return new QueueServices(connection, publishChannel, stagingQueueOrTargetQueue, targetQueue, codec);
+        return new QueueServices(connection, publishChannel, targetQueue, targetQueue, codec);
```
rorytorneymf: I think these comments can be removed now as both these will always be the same now:

```
// stagingQueueOrTargetQueue = Queue where message should be published to
// targetQueue = Queue that should be set in the 'to' field of the task message
```

I think QueueServices can be updated to remove the extra queue from its constructor:

```
public QueueServices(
            final Connection connection,
            final Channel publisherChannel,
            final String stagingQueueOrTargetQueue,
            final String targetQueue,
            final Codec codec) 
```

to:

```
public QueueServices(
            final Connection connection,
            final Channel publisherChannel,
            final String targetQueue,
            final Codec codec) 
```

Note though on the original ticket here:

https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=627110 

> FYI:
> 
> Unlike when workers send messages to staging queues, when the Job Service Scheduled Executor sends messages to staging queues, it does *not* change the `to` field of the task message to reference the staging queue.
> 
> The `to` field will contain the original target queue.

But I cannot remember why this was required. With these new changes I think the `to` field will now be set to the staging queue, rather than the target queue as it was before. I am not sure what the implications of this will be.

andyreidz: That's handled by setting  CAF_WORKER_ENABLE_DIVERTED_TASK_CHECKING to false

rorytorneymf (approved):

PR #460, opened 2024-11-12
US975260: Refactor HPE references

raghavendran-gopalakrishnan:
https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=975260


job-service-acceptance-tests/pom.xml
```diff
@@ -374,36 +370,6 @@
...
-                                            <fileSet>
-                                                <directory>${project.build.directory}/worker-batch-plugins-package-unpacked/batch-plugins</directory>
-                                                <outputDirectory>.</outputDirectory>
-                                            </fileSet>
-                                        </fileSets>
-                                    </inline>
-                                </assembly>
-                            </build>
-                        </image>
-
```
dermot-hardy, line 406: Removing this is intentional?

dermot-hardy (approved):

job-service-acceptance-tests/pom.xml
raghavgopal: Yes the worker-batch-plugins doesn't exist since quite a while, also it has references to HPE in the pom. Removing it has no harm to the tests.

image

job-service-unit-tests/src/test/java/com/github/jobservice/core/api/TestUtil.java
```diff
@@ -36,6 +36,8 @@ public static void setSystemEnvironmentFields(Map<String, String> newenv) throws
                 map.putAll(newenv);
             }
         }
+        System.out.println(env);
+        System.out.println(System.getenv("JOB_SERVICE_DATABASE_PASSWORD"));
```
dermot-hardy, line 40: Was leaving these lines in intentional?

raghavgopal: Um, looks like I put this to debug some tests. Its unintentional actually.

PR #466, opened 2024-11-26
D991106: Configure cache with auto eviction

kusumaghoshdastidar:
https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=991106

### DIFF
--- a/job-service-config/src/main/java/com/github/jobservice/configuration/AppConfig.java
+++ b/job-service-config/src/main/java/com/github/jobservice/configuration/AppConfig.java
@@ -28,6 +28,7 @@ import java.util.Locale;
 import java.util.concurrent.ExecutionException;
 import java.util.regex.Pattern;
 
+import com.google.common.base.Strings;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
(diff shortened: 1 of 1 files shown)


job-service-config/src/main/java/com/github/jobservice/configuration/AppConfig.java
```diff
@@ -38,15 +39,23 @@
 @Configuration
 @PropertySource(value = "classpath:${JOB_SERVICE_API_CONFIG_PATH:config.properties}", ignoreResourceNotFound = true)
 public class AppConfig {
+    private static final String JOB_SERVICE_SECRETS_CACHE_COUNT_CONFIG = "JOB_SERVICE_SECRETS_CACHE_COUNT";
+    private static final int JOB_SERVICE_SECRETS_CACHE_COUNT_DEFAULT = 20;
```
kusumaghoshdastidar, line 43: Should probably have a smaller value, just 2?

job-service-config/src/main/java/com/github/jobservice/configuration/AppConfig.java
```diff
@@ -38,15 +39,23 @@
...
 
     private static final LoadingCache<String, String> SECRETS_CACHE = CacheBuilder.newBuilder()
+            .maximumSize(getEnvOrDefault(JOB_SERVICE_SECRETS_CACHE_COUNT_CONFIG, JOB_SERVICE_SECRETS_CACHE_COUNT_DEFAULT))
```
dermot-hardy, line 46: I'll approve but tbh I don't think this needs to be configurable at all.  Looking at the code there are just two possible keys at the minute.  It does no harm to use a slightly higher value in case more secrets are added but I just can't think when or why anyone would ever adjust this configuration entry.

dermot-hardy (approved):

PR #480, opened 2024-12-16
US921241: Add update job progress to scheduled executor

caoimhelynch:
https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=921241

### DIFF
--- a/job-service-contract/src/main/resources/com/github/jobservice/contract/swagger.yaml
+++ b/job-service-contract/src/main/resources/com/github/jobservice/contract/swagger.yaml
@@ -90,6 +90,18 @@ paths:
           - Completed
           - Inactive
           - NotFinished
+      - name: status
+        in: query
+        type: string
+        required: false
+        description: Active - only those with active status will be returned; Cancelled - only those with cancelled status will be returned; Completed - only those results with completed status will be returned; Failed - only those results with failed status will be returned; Paused - only those results with paused status will be returned, Waiting - only those results with waiting status will be returned.
+        enum:
+          - Active
+          - Cancelled
+          - Completed
+          - Failed
+          - Paused
+          - Waiting
       - name: limit
         in: query
         type: integer
--- a/job-service-contract/src/main/resources/com/github/jobservice/contract/swagger.yaml
+++ b/job-service-contract/src/main/resources/com/github/jobservice/contract/swagger.yaml
@@ -459,6 +459,18 @@ paths:
           - Completed
           - Inactive
           - NotFinished
+      - name: status
+        in: query
+        type: string
+        required: false
+        description: Active - only those with active status will be returned; Cancelled - only those with cancelled status will be returned; Completed - only those results with completed status will be returned; Failed - only those results with failed status will be returned; Paused - only those results with paused status will be returned, Waiting - only those results with waiting status will be returned.
+        enum:
+          - Active
+          - Cancelled
+          - Completed
+          - Failed
+          - Paused
+          - Waiting
       - name: filter
         in: query
         type: string
--- a/job-service-db/src/main/resources/db/migration/procedures/public/R__updateJobProgress.sql
+++ b/job-service-db/src/main/resources/db/migration/procedures/public/R__updateJobProgress.sql
@@ -43,7 +43,7 @@ BEGIN
           FOR UPDATE;
 
         -- Perform update on the row
-        PERFORM internal_update_job_progress(rec.partition_id, rec.job_id);
+        PERFORM internal_update_job_progress(task.partition_id, task.job_id);
 
         -- Complete transaction to update row
         COMMIT;
--- a/job-service-db/src/main/resources/db/migration/procedures/public/R__updateJobProgress.sql
+++ b/job-service-db/src/main/resources/db/migration/procedures/public/R__updateJobProgress.sql
@@ -24,16 +24,29 @@ CREATE OR REPLACE PROCEDURE public.update_job_progress(
 	IN in_num_of_tasks_to_update integer DEFAULT 100)
 LANGUAGE plpgsql
 AS $$
+DECLARE
+    task RECORD;
 BEGIN
 
-    PERFORM internal_update_job_progress(tasks_to_update.partition_id, tasks_to_update.job_id)
-    FROM (
+    FOR task IN
         SELECT partition_id, job_id
         FROM completed_subtask_report
         GROUP BY partition_id, job_id
         ORDER BY random()
         LIMIT in_num_of_tasks_to_update
-    ) tasks_to_update;
+    LOOP
+        -- Take out an exclusive update lock on the job row
+        PERFORM NULL
+        FROM job j
+        WHERE j.partition_id = rec.partition_id
+          AND j.job_id = rec.job_id
+          FOR UPDATE;
 
+        -- Perform update on the row
+        PERFORM internal_update_job_progress(rec.partition_id, rec.job_id);
+
+        -- Complete transaction to update row
+        COMMIT;
+    END LOOP;
 END
 $$;
--- a/job-service-db/src/main/resources/db/migration/procedures/public/R__updateJobProgress.sql
+++ b/job-service-db/src/main/resources/db/migration/procedures/public/R__updateJobProgress.sql
@@ -38,8 +38,8 @@ BEGIN
         -- Take out an exclusive update lock on the job row
         PERFORM NULL
         FROM job j
-        WHERE j.partition_id = rec.partition_id
-          AND j.job_id = rec.job_id
+        WHERE j.partition_id = task.partition_id
+          AND j.job_id = task.job_id
           FOR UPDATE;
 
         -- Perform update on the row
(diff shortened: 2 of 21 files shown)


job-service-db/src/main/resources/db/migration/procedures/public/R__update_job_progress.sql
```diff
@@ -0,0 +1,41 @@
...
+AS $$
+BEGIN
+
+    PERFORM internal_update_job_progress(tasks_to_update.partition_id, tasks_to_update.job_id)
+    FROM (
+        SELECT partition_id, job_id
+        FROM (
+            SELECT DISTINCT partition_id, job_id
+            FROM completed_subtask_report
+        ) distinct_tasks
```
dermot-hardy: The inner `SELECT` is not necessary:

```suggestion
        FROM completed_subtask_report
        GROUP BY partition_id, job_id
```

job-service-db/src/main/resources/db/migration/procedures/public/R__update_job_progress.sql
```diff
@@ -0,0 +1,41 @@
...
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+--
+
+/*
+ *  Name: update_job_progress
```
dermot-hardy: Adjust filename to match others.

job-service-db/src/main/resources/db/migration/procedures/public/R__update_job_progress.sql
```diff
@@ -0,0 +1,41 @@
...
+ *  Description:
+ *  This procedure updates the percentage_complete in the job table with data from completed_subtask_report table.
+ */
+CREATE OR REPLACE PROCEDURE public.update_job_progress(
+	IN in_num_of_tasks_to_update integer DEFAULT 100)
+LANGUAGE plpgsql
+AS $$
+BEGIN
+
+    PERFORM internal_update_job_progress(tasks_to_update.partition_id, tasks_to_update.job_id)
```
dermot-hardy: We need to control the transaction here.  We want each job to be updated in its own transaction, even when multiple jobs are being updated.  So for example if 100 jobs are being updated I would like for there to be 100 transactions.  This function should be called once in each transaction.

dermot-hardy (changes requested):

job-service-scheduled-executor/src/main/java/com/github/jobservice/scheduledexecutor/ScheduledExecutor.java
```diff
@@ -48,6 +48,10 @@ public ScheduledExecutor() {
         //  Execute the dropTablesTask periodically.
         scheduler.scheduleWithFixedDelay(new DropTablesTask(), 20, ScheduledExecutorConfig.getDropTablesSchedulerPeriod(),
                 TimeUnit.SECONDS);
+
+        //  Execute the updateJobProgress periodically.
+        scheduler.scheduleWithFixedDelay(new UpdateJobProgress(100), 20,
+                ScheduledExecutorConfig.getJobProgressUpdatePeriod(), TimeUnit.SECONDS);
```
dermot-hardy, line 58: Give this its own single thread scheduler rather than reusing the job dispatch one.  We may as well create a third one for the drop tables when we're in here since it shouldn't be reusing the job dispatch one either.

dermot-hardy (changes requested):

job-service-scheduled-executor/src/main/java/com/github/jobservice/scheduledexecutor/ScheduledExecutor.java
```diff
@@ -48,6 +48,10 @@ public ScheduledExecutor() {
         //  Execute the dropTablesTask periodically.
         scheduler.scheduleWithFixedDelay(new DropTablesTask(), 20, ScheduledExecutorConfig.getDropTablesSchedulerPeriod(),
                 TimeUnit.SECONDS);
+
+        //  Execute the updateJobProgress periodically.
+        scheduler.scheduleWithFixedDelay(new UpdateJobProgress(100), 20,
```
dermot-hardy: I'm fine with the 20 being hard-coded since it is just an initial delay but we should make that 100 configurable.

job-service-contract/src/main/resources/com/github/jobservice/contract/swagger.yaml
```diff
@@ -90,6 +90,18 @@ paths:
...
+        type: string
+        required: false
+        description: Active - only those with active status will be returned; Cancelled - only those with cancelled status will be returned; Completed - only those results with completed status will be returned; Failed - only those results with failed status will be returned; Paused - only those results with paused status will be returned, Waiting - only those results with waiting status will be returned.
+        enum:
+          - Active
+          - Cancelled
+          - Completed
+          - Failed
+          - Paused
+          - Waiting
```
dermot-hardy, line 104: From reading the ticket it seems like the expectation was that this parameter would be added to the `/jobStats/count` endpoint as well.  Was this just missed?  Can it be easily added or should we extract it to a separate ticket?

job-service-scheduled-executor/src/main/java/com/github/jobservice/scheduledexecutor/ScheduledExecutorConfig.java
```diff
@@ -156,4 +156,12 @@ private static String getPropertyOrEnvVar(final String key)
         final String propertyValue = System.getProperty(key);
         return (propertyValue != null) ? propertyValue : System.getenv(key);
     }
+
+    public static int getJobProgressUpdatePeriod() {
+        final String  period = getPropertyOrEnvVar("CAF_JOB_PROGRESS_UPDATE_SCHEDULER_PERIOD");
```
dermot-hardy, line 169:
```suggestion
        final String period = getPropertyOrEnvVar("CAF_JOB_PROGRESS_UPDATE_SCHEDULER_PERIOD");
```

job-service-scheduled-executor/src/main/java/com/github/jobservice/scheduledexecutor/ScheduledExecutorConfig.java
```diff
@@ -156,4 +156,12 @@ private static String getPropertyOrEnvVar(final String key)
...
+        if (null == period || period.isEmpty()) {
+            return 60;
```
dermot-hardy: Let's make this much more aggressive - maybe every 2 seconds by default.

job-service-db/src/main/resources/db/migration/procedures/public/R__updateJobProgress.sql
```diff
@@ -0,0 +1,52 @@
...
+        FROM completed_subtask_report
+        GROUP BY partition_id, job_id
+        ORDER BY random()
+        LIMIT in_num_of_tasks_to_update
+    LOOP
+        -- Take out an exclusive update lock on the job row
+        PERFORM NULL
+        FROM job j
+        WHERE j.partition_id = rec.partition_id
+          AND j.job_id = rec.job_id
```
dermot-hardy: This `rec` is clearly a mistake.  It is concerning that this issue doesn't cause any tests to fail, since presumably it means that no jobs would complete.

dermot-hardy (changes requested):

job-service-db/src/main/resources/db/migration/procedures/public/R__updateJobProgress.sql
```diff
@@ -0,0 +1,52 @@
...
+    LOOP
+        -- Take out an exclusive update lock on the job row
+        PERFORM NULL
+        FROM job j
+        WHERE j.partition_id = task.partition_id
+          AND j.job_id = task.job_id
+          FOR UPDATE;
+
+        -- Perform update on the row
+        PERFORM internal_update_job_progress(rec.partition_id, rec.job_id);
```
dermot-hardy:
```suggestion
        PERFORM internal_update_job_progress(task.partition_id, task.job_id);
```

job-service-contract/src/main/resources/com/github/jobservice/contract/swagger.yaml
caoimhelynch: This was missed but should be easy to add. Will work on adding that in.

job-service-scheduled-executor/src/main/java/com/github/jobservice/scheduledexecutor/ScheduledExecutorConfig.java
```diff
@@ -156,4 +156,20 @@ private static String getPropertyOrEnvVar(final String key)
         final String propertyValue = System.getProperty(key);
         return (propertyValue != null) ? propertyValue : System.getenv(key);
     }
+
+    public static int getJobProgressTasksToUpdate() {
+        final String  numOfTasks = getPropertyOrEnvVar("CAF_JOB_PROGRESS_NUM_TASKS_TO_UPDATE");
+        if (null == numOfTasks || numOfTasks.isEmpty()) {
+            return 2;
```
dermot-hardy: Put this back to 100.

dermot-hardy (approved):

PR #491, opened 2025-01-27
US990352: Correlate specific documents with audit events

jonny-mccoubrey:
https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=990352


jonny-mccoubrey:
## Description of Changes

### Job Service Database
Added new nullable column to _job_task_data_ and _job_ tables named `correlation_id`.
Updated database function scripts _R__createJob.sql_, _R__internal__createJob.sql_, and _R__createJobForDependencies.sql_ with additional parameter correlation id (to be saved to _job_task_data_ and _job_ tables)

Uodated _R__getDependentJobs.sql_ to return correlation id from _job_task_data_ table.

### Job Service
Jobs created now include correlation id from MDC (from incoming request or generated by [caf-correlation-jservlet](https://github.com/CAFapi/caf-correlation/tree/main/caf-correlation-jservlet) filter).

### Job Service Scheduled Executor
Correlation id is now retrieved from database as part of _JobTaskData_ object and set in the _TaskMessage_ that is sent out.

job-service-db/src/main/resources/db/migration/functions/public/R__getDependentJobs.sql
```diff
@@ -20,6 +20,8 @@
  *  Description:
  *  Returns a list of dependent jobs that are now eligible to run.
  */
+DROP FUNCTION IF EXISTS get_dependent_jobs();
```
dermot-hardy: - This should be in the VX__ scripts.
- The other functions whose parameters have changed should also be there.

job-service-db/src/main/resources/db/migration/V8__add_correlation_id_to_job_tables.sql
```diff
@@ -0,0 +1,27 @@
...
+ **************************************************************
+ ******************** Update tables ***************************
+ **************************************************************
+ */
+
+ALTER TABLE job
+    ADD COLUMN IF NOT EXISTS correlation_id VARCHAR(48) NULL;
+
+ALTER TABLE job_task_data
+    ADD COLUMN IF NOT EXISTS correlation_id VARCHAR(48) NULL;
```
dermot-hardy: We need to ensure the upgrade instructions take account of these changes.

See https://gitlab.otxlab.net/cybersecurity-enterprise/cddri/packaging/application-install-package/-/tree/main/upgrade-procedures/24.4.2.HF1?ref_type=heads for previous example.

job-service-scheduled-executor/src/main/java/com/github/jobservice/scheduledexecutor/WorkerAction.java
```diff
@@ -48,6 +48,7 @@ public String toString() {
     private TaskDataEncodingEnum taskDataEncoding = null;
     private String taskPipe = null;
     private String targetPipe = null;
+    private String correlationId = null;
```
dermot-hardy, line 53: Comment above that adding the correlation id to this class but not to the other generated class that is referenced is intentional and not just an oversight.

job-service-scheduled-executor/src/main/java/com/github/jobservice/scheduledexecutor/WorkerAction.java
```diff
@@ -121,6 +122,19 @@ public void setTargetPipe(final String targetPipe) {
...
+    public String getCorrelationId() {
+        return correlationId;
+    }
+
+    public void setCorrelationId(final String correlationId) {
+        this.correlationId = correlationId;
+    }
+
     @Override
     public boolean equals(final Object o) {
```
dermot-hardy, line 141: - The `equals()` method has not been updated for the addition of the correlation id.
- Same with the `hashCode()` method below.
- And `toString()` ...

job-service-scheduled-executor/src/main/java/com/github/jobservice/scheduledexecutor/JobTaskData.java
```diff
@@ -37,6 +39,8 @@ public class JobTaskData
     private String taskPipe;
     @NotNull
     private String targetPipe;
+    @Null
```
dermot-hardy: This means `null`-only, not nullable.  Just get rid of it.  Nullable is the default.

dermot-hardy (changes requested):

job-service-db/src/main/resources/db/migration/V8__add_correlation_id_to_job_tables.sql
jonny-mccoubrey: I have added this to the comments of the ticket

dermot-hardy (approved):

dermot-hardy:
Update release notes about this before merging.

PR #506, opened 2025-03-25
US1008373: Release prep

caoimhelynch:
https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=1008373


raghavendran-gopalakrishnan (approved):

release-notes-9.2.1.md
```diff
@@ -0,0 +1,11 @@
+#### Version Number
+${version-number}
+
+#### New Features
+- None
+
+#### Patch Fixes Included
+- This release includes OS package updates only
```
dermot-hardy, line 8: This is not true.  Don't write that there have only been OS updates when there have clearly been lots of Java dependency updates as well.

PR #513, opened 2025-05-13
US1016047: Support CAF_RABBITMQ_TLS_PROTOCOL_VERSION

rorytorneymf:
https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=1016047

### DIFF
--- a/job-service-core/src/main/java/com/github/jobservice/core/queue/QueueServicesFactory.java
+++ b/job-service-core/src/main/java/com/github/jobservice/core/queue/QueueServicesFactory.java
@@ -16,6 +16,7 @@
 package com.github.jobservice.core.queue;
 
 import com.github.cafapi.common.api.Codec;
+import com.github.workerframework.configs.RabbitConfiguration;
 import com.github.workerframework.util.rabbitmq.RabbitUtil;
 import com.github.jobservice.configuration.AppConfig;
 import com.github.jobservice.configuration.AppConfigException;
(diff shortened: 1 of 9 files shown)


job-service-core/src/main/java/com/github/jobservice/core/queue/QueueServicesFactory.java
```diff
@@ -75,7 +76,17 @@ public static QueueServices create(final AppConfig configuration, final String t
     private static Connection createConnection(AppConfig configuration)
             throws IOException, TimeoutException, URISyntaxException, NoSuchAlgorithmException, KeyManagementException, AppConfigException
     {
-        return RabbitUtil.createRabbitConnection(configuration.getRabbitMQProtocol(), configuration.getRabbitMQHost(),
-                configuration.getRabbitMQPort(), configuration.getRabbitMQUsername(), configuration.getRabbitMQPassword());
+        final RabbitConfiguration rabbitConfiguration = new RabbitConfiguration();
```
rorytorneymf, line 79: I'm directly using `RabbitConfiguration` here and in the [Tag Service PR](https://gitlab.otxlab.net/cybersecurity-enterprise/cddri/app/tag-service/-/merge_requests/525/diffs#fb044c31e42064adbbe4576b683628354559ab37_135_136), 

The alternative to directly using `RabbitConfiguration` would be to add another overloaded method of `RabbitUtil.createRabbitConnection` that takes the TLS protocol version, but that would require another release of the worker framework.

Since `util-rabbitmq` already has a dependency on `worker-configs`, I thought it was ok to do it this way, but if you prefer I can make the change to the Worker Framework instead?

dermot-hardy (approved):

job-service-core/src/main/java/com/github/jobservice/core/queue/QueueServicesFactory.java
dermot-hardy: It is odd but as you say the dependency already existed.

michael-bryson:
@rorytorneymf If this is a new feature then the repo should have been re-versioned to the next minor version.

rorytorneymf:
> @rorytorneymf If this is a new feature then the repo should have been re-versioned to the next minor version.

Yes thanks I will revision this and the Tag Service.

Regarding the workers, I see that it was added as a new feature to the Worker Framework:

https://github.com/WorkerFramework/worker-framework/releases/tag/v9.1.0 

image

but I don't think any workers were re-versioned as part of that or had their release notes updated. Previously, when workers got new functionality from the Worker Framework, it was added to each workers release notes as well, e.g:

image

image

so I think I should probably do the same here for all the workers (move them to a minor release and update their release notes). @dermot-hardy do you agree?

dermot-hardy:
> I think I should probably do the same here for all the workers (move them to a minor release and update their release notes). @dermot-hardy do you agree?

I do agree.  It could be argued that it is part of the workers' release tickets to check on the version and release notes, but doing it now will make it less likely to be missed at that time.

PR #528, opened 2025-07-19
D1038066: Stopped using Spring in the config classes

dermot-hardy:
[D1038066 - Job Service: Memory leak](https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=1038066)

These changes resolve a memory leak resulting from the improper use of Spring &mdash; similar to [spring-framework#12159](https://github.com/spring-projects/spring-framework/issues/12159).  The use of Spring in these classes was overkill so I just removed it.

As a side-effect the changes also remove the defaults that were inadvertently added by #115 changes.  I have incremented the major version number because of this.


raghavendran-gopalakrishnan (approved):
LGTM

PR #531, opened 2025-08-01
D1041027: Improved error handling (Job Service Scheduled Executor)

rorytorneymf:
https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=1041027

Non-transient errors when sending a message to the target queue will now result in the job being marked as failed in the database and the job will not be retried.


job-service-scheduled-executor/src/main/java/com/github/jobservice/scheduledexecutor/QueueServices.java
```diff
@@ -45,18 +45,23 @@ public final class QueueServices implements AutoCloseable
 {
     private static final Logger LOG = LoggerFactory.getLogger(QueueServices.class);
 
+    private final String partitionId;
```
areid2ot: I noticed this draft PR and have an early comment about caching QueueServices objects.
It looks like we'd have one of these for every job and partition, meaning we'd have more connections and channels open at once. 

Check how workers interact with queues, I think they use a single connection and channel.

rorytorneymf: The existing code already created one QueueServices instance per job though (i.e. an entry in the job_task_data table), nothing is changing in that regards.

What's changing is that we keep the QueueServices instance in scope until we get a response. Previously they were closed immediately after sending the message.

Int he normal happy-path case, I would expect these to be very short-lived objects as we should receive an ack from RabbitMQ almost immediately.

I can look at make QueueServices instance shared across jobs, but its going to add a lot of complication (tying messages back to a particular job, a failure to send a message for one job causing other jobs to fail because the shared channel/connection got closed), and I'm not sure its needed?

rorytorneymf: So currently I have:

- 1 connection per job + 1 channel per job

I think to not use a new connection + channel per job I'm probably going to need to implement something similar to what the worker-framework does as you mention (reusing same connection + channel):

https://github.com/WorkerFramework/worker-framework/blob/develop/worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerPublisherImpl.java

however, that just uses one connection + channel, as we might be dealing with multiple queues, we will need to handle multiple channels, which I I think this is probably result in a lot of code being added here, if we do something like:

- 1 shared connection for all jobs + 1 channel **per queue**

Would just reusing the connection, but still create a new channel per job be sufficient:

- 1 shared connection for all jobs + 1 channel **per job**

or is there a worry that we could still exhaust RabbitMQ if we e.g. got 2000 jobs to process, that would result in 2000 channels?

@dermot-hardy would you have any thoughts?

dermot-hardy: I think either option is fine.  Presumably if you go for a channel per job and you run out of channels then the message won't be published and will just be published a few seconds later instead in the next interval.  If that is the case then it's fine.

rorytorneymf: Yes that is what happens, a failure to create a channel is identified as a transient error and retried in the next interval. I will go with 1 shared connection + 1 channel per job for now.

job-service-scheduled-executor/src/main/java/com/github/jobservice/scheduledexecutor/RabbitMqAsyncListener.java
```diff
@@ -0,0 +1,230 @@
...
+                deliveryTag, multiple, partitionId, jobId, targetQueue);
+
+        // A successful ACK means we can safely remove the job from the system.
+        cleanupSuccessfulJob();
+    }
+
+    /**
+     * Handles a message rejection (NACK) from the broker.
+     * NACKs are typically caused by transient issues, such as a queue reaching its capacity,
+     * so they are treated as retryable failures.
```
rorytorneymf: I did not find a way to determine if a nack is transient or not, so assuming non-transient. Maybe we could have something like a job->nack counter, and if we receive N nacks for that job, consider it a non-transient error rather than a transient error?

job-service-acceptance-tests/pom.xml
```diff
@@ -625,6 +625,10 @@
                                     <CAF_RABBITMQ_PASSWORD>guest</CAF_RABBITMQ_PASSWORD>
                                     <CAF_SCHEDULED_EXECUTOR_PERIOD>2</CAF_SCHEDULED_EXECUTOR_PERIOD>
                                     <CAF_WORKER_INPUT_QUEUE>jobservicescheduler-in</CAF_WORKER_INPUT_QUEUE>
+                                    <!-- Can't use quorum as the integration tests (the ones that are not disabled) are
+                                    using an old worker that does not support the 'queueType' prop in its js config:
+                                    cfg_test_worker-batch_RabbitWorkerQueueConfiguration -->
+                                    <CAF_RABBITMQ_QUEUE_TYPE>classic</CAF_RABBITMQ_QUEUE_TYPE>
```
dermot-hardy, line 631: How is this only being added now?  Were the tests silently failing before?

rorytorneymf: They were passing before because the job status was not being changed to 'Failed' when it failed with a non-transient error (such as declaring a queue with the wrong args), and now it does, so these assertions started failing:

```
[ERROR]   JobServiceEndToEndIT.testPausePausedJobIsSuccessful:1260->pauseJob:1531 ▒ Api {"message":"ERROR: job_id {J1754557081903-1911} cannot be paused as it has a status of {Failed}. Only jobs with a status of Active or Waiting can be paused.\n  Where: PL/pgSQL function pause_job(character varying,character varying) line 23 at RAISE"}
[ERROR]   JobServiceEndToEndIT.testPauseWaitingJobIsSuccessful:1251->pauseJob:1531 ▒ Api {"message":"ERROR: job_id {J1754557082073-7842} cannot be paused as it has a status of {Failed}. Only jobs with a status of Active or Waiting can be paused.\n  Where: PL/pgSQL function pause_job(character varying,character varying) line 23 at RAISE"}
[ERROR]   JobServiceEndToEndIT.testResumeCancelledJobIsNotAllowed:1406->cancelJob:1523 ▒ Api {"message":"ERROR: job_id {J1754557082222-4273} cannot be cancelled\n  Where: PL/pgSQL function cancel_job(character varying,character varying) line 24 at RAISE"}
[ERROR]   JobServiceEndToEndIT.testResumePausedJobThatWasPreviouslyWaitingMovesJobToActiveStatus:1310->pauseJob:1531 ▒ Api {"message":"ERROR: job_id {J1754557082493-6876} cannot be paused as it has a status of {Failed}. Only jobs with a status of Active or Waiting can be paused.\n  Where: PL/pgSQL function pause_job(character varying,character varying) line 23
```

job-service-scheduled-executor/src/main/java/com/github/jobservice/scheduledexecutor/DatabasePoller.java
```diff
@@ -142,4 +221,34 @@ private static List<JobTaskData> getDependentJobsToRun() throws ScheduledExecuto
...
+            LOG.info("Calling report_failure() database function with partitionId={}, jobId={}, failureDetails={} ...",
+                    partitionId, jobId, failureDetails);
+            stmt.execute();
+        } catch (final SQLException e) {
+            final String errorMessage = MessageFormat.format(
+                    "Failed in call to report_failure() database function with " +
+                            "partitionId={0}, jobId={1}, failureDetails={2}. {3}",
+                    partitionId, jobId, failureDetails, e.getMessage());
+            LOG.error(errorMessage);
+            throw new ScheduledExecutorException(errorMessage);
```
dermot-hardy, line 249: If this `SQLException` is a transient database issue then this will be re-thrown as a `ScheduledExecutorException` and then the `handleNonTransientFailure()` will log a message saying that the job will be retried, but will it?

job-service-scheduled-executor/src/main/java/com/github/jobservice/scheduledexecutor/QueueServicesFactory.java
```diff
@@ -47,42 +59,100 @@ public final class QueueServicesFactory
...
+
+        Connection connection = SHARED_CONNECTION.get();
+
+        // Check if we need to create or recreate the connection
+        if (connection == null || !connection.isOpen()) {
+            synchronized (QueueServicesFactory.class) {
+                // Double-check pattern to avoid race conditions
+                connection = SHARED_CONNECTION.get();
+                if (connection == null || !connection.isOpen()) {
+                    if (connection != null && !connection.isOpen()) {
```
dermot-hardy, line 141: What's the idea of doing this twice?

job-service-scheduled-executor/src/main/java/com/github/jobservice/scheduledexecutor/DatabasePoller.java
rorytorneymf: Yes the job would be retried since we don't delete from the job_task_data if the DatabasePoller.reportFailure throws, either transient or non-transient:

```
[2025-08-08 09:42:58.957Z #681.049 INFO  -            -   ] c.g.j.s.DatabasePoller: Processing job id 3 ...
[2025-08-08 09:42:58.967Z #681.049 ERROR -            -   ] c.g.j.s.DatabasePoller: {"message":"NON-TRANSIENT failure detected during message publishing. Marking job as failed. [partitionId=1, jobId=3, taskPipe=TaskQueue_542e0a2d-2e6b-4681-b840-f458fb1bfe0e].","exception":"java.lang.IllegalArgumentException: Simulated exception for job 3 in try\n\tat com.github.jobservice.scheduledexecutor.DatabasePoller.sendMessageToQueueMessaging(DatabasePoller.java:95)\n\tat com.github.jobservice.scheduledexecutor.DatabasePoller.pollDatabaseForJobsToRun(DatabasePoller.java:71)\n\tat com.github.jobservice.scheduledexecutor.ScheduledExecutor.runAvailableJobs(ScheduledExecutor.java:79)\n\tat com.github.jobservice.scheduledexecutor.ScheduledExecutor.lambda$poke$1(ScheduledExecutor.java:63)\n\tat java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:572)\n\tat java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)\n\tat java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304)\n\tat java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)\n\tat java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)\n\tat java.base/java.lang.Thread.run(Thread.java:1583)\n"}
[2025-08-08 09:42:58.973Z #681.049 ERROR -            -   ] c.g.j.s.DatabasePoller: Failed in call to report_failure() database function with partitionId=1, jobId=3, failureDetails={"failureId":"ADD_TO_QUEUE_FAILURE","failureTime":"2025-08-08T09:42:58.967Z","failureSource":"Job Service Scheduled Executor for job id 3","failureMessage":"Exception during message publishing: Simulated exception for job 3 in try"}. simulated SQLException for job 3 in reportFailure
[2025-08-08 09:42:58.973Z #681.049 ERROR -            -   ] c.g.j.s.JobFailureHandler: {"message":"Failed to mark job as failed in the database. Job will be retried. [partitionId=1, jobId=3, error=Failed in call to report_failure() database function with partitionId=1, jobId=3, failureDetails={\"failureId\":\"ADD_TO_QUEUE_FAILURE\",\"failureTime\":\"2025-08-08T09:42:58.967Z\",\"failureSource\":\"Job Service Scheduled Executor for job id 3\",\"failureMessage\":\"Exception during message publishing: Simulated exception for job 3 in try\"}. simulated SQLException for job 3 in reportFailure]","exception":"com.github.jobservice.scheduledexecutor.ScheduledExecutorException: Failed in call to report_failure() database function with partitionId=1, jobId=3, failureDetails={\"failureId\":\"ADD_TO_QUEUE_FAILURE\",\"failureTime\":\"2025-08-08T09:42:58.967Z\",\"failureSource\":\"Job Service Scheduled Executor for job id 3\",\"failureMessage\":\"Exception during message publishing: Simulated exception for job 3 in try\"}. simulated SQLException for job 3 in reportFailure\n\tat com.github.jobservice.scheduledexecutor.DatabasePoller.reportFailure(DatabasePoller.java:257)\n\tat com.github.jobservice.scheduledexecutor.JobFailureHandler.handleNonTransientFailure(JobFailureHandler.java:107)\n\tat com.github.jobservice.scheduledexecutor.DatabasePoller.sendMessageToQueueMessaging(DatabasePoller.java:137)\n\tat com.github.jobservice.scheduledexecutor.DatabasePoller.pollDatabaseForJobsToRun(DatabasePoller.java:71)\n\tat com.github.jobservice.scheduledexecutor.ScheduledExecutor.runAvailableJobs(ScheduledExecutor.java:79)\n\tat com.github.jobservice.scheduledexecutor.ScheduledExecutor.lambda$poke$1(ScheduledExecutor.java:63)\n\tat java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:572)\n\tat java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)\n\tat java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304)\n\tat java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)\n\tat java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)\n\tat java.base/java.lang.Thread.run(Thread.java:1583)\n"}
[2025-08-08 09:43:37.187Z #681.049 INFO  -            -   ] c.g.c.c.u.m.ModuleLoader: Detected component implementation JsonCodec
[2025-08-08 09:43:37.187Z #681.049 INFO  -            -   ] c.g.j.s.DatabasePoller: Processing job id 3 ...
```

For a non-transient SQLException, this will mean the job will be retried, if it fails again, and the report_failure function also fails again,  it means we will never be able to mark the job as failed in the db. 

I am not sure what our options are here if we are unable to mark the job as failed in the db other that to keep retrying, and hope we stop getting non-transient SQLExceptions eventually?

If we don't mark the job as failed and delete the row from job_task_data the job message will be lost and the job would be stuck in the 'Waiting' state forever.

job-service-scheduled-executor/src/main/java/com/github/jobservice/scheduledexecutor/QueueServicesFactory.java
```diff
@@ -47,42 +59,100 @@ public final class QueueServicesFactory
...
         final Channel publishChannel = connection.createChannel();
+        if (publishChannel == null) {
+            throw new IOException(
+                    MessageFormat.format(
+                            "Failed to create a new channel for job [partitionId={0}, jobId={1}, queue={2}]. " +
+                            "Job will be retried.", partitionId, jobId, targetQueue));
+        }
+
+        publishChannel.confirmSelect();
+        RabbitMqAsyncListener.registerListeners(publishChannel, partitionId, jobId, targetQueue);
```
dermot-hardy, line 115: Doesn't the `.confirmSelect()` enable synchronous acknowledgments, defeating the purpose of the async-listening on the next line?

dermot-hardy: Looking at this a bit more you also require `.waitForConfirmsOrDie()` for synchronous listening, like we do in [`job-service-core/.../QueueServices.java:148`](https://github.com/JobService/job-service/blob/v10.0.0/job-service-core/src/main/java/com/github/jobservice/core/queue/QueueServices.java#L148).  Is that something we could do here or do we require the complexity of asynchronous listening for performance reasons?

rorytorneymf: I am looking at this now. 

My initial though was that if we did something like:

```
publisherChannel.basicPublish("", targetQueue, true, MessageProperties.PERSISTENT_TEXT_PLAIN, taskMessageBytes);
publisherChannel.waitForConfirmsOrDie(10000);
```

then because we send one message per job in a for loop, and if we had 50 jobs to run:

```
publisherChannel.basicPublish(job1)
publisherChannel.basicPublish(job2)
...
publisherChannel.basicPublish(job50)
```

and there was an issue sending `publisherChannel.basicPublish(job2)` that caused it to block for the timeout period (10 seconds), then that would hold up sending all the remaining 48 messages behind it.

But in all likelihood, if there is something causing  `publisherChannel.basicPublish(job2)` to block, its maybe likely that the same problem would happen for the remaining 48 messages, so the fact that they are not sent until job2 unblocks is maybe ok, and could possibly be preferable? i.e. if its a transient RabbitMQ issue, only job2 would fail, and by the time the remaining jobs come to run 10 seconds later, maybe the transient issue may have resolved itself?

dermot-hardy: If it's a RabbitMQ issue then that sounds right to me &mdash; no point publishing new messages if RabbitMQ is unhealthy.  On the other hand if it is something specific about the job2 message that is causing RabbitMQ to actively reject it (too large for example), then we shouldn't hold up other messages over something like that.

rorytorneymf:
Closing this MR. I have created a new MR that uses a simpler approach here:

https://github.com/JobService/job-service/pull/535

PR #532, opened 2025-08-06
US1009117: Remove reliance on large messages being supported

DavidGerald-Milligan:
https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=1009117


job-service-scheduled-executor-container/src/main/config/cfg~caf~worker~DocumentWorkerConfiguration.js
```diff
@@ -17,6 +17,8 @@
     workerName: "JobServiceScheduler",
     workerVersion: "${project.version}",
     outputQueue: getenv("CAF_WORKER_OUTPUT_QUEUE") || undefined,
+    invalidQueue: getenv("CAF_WORKER_INVALID_QUEUE")
+        || (getenv("CAF_WORKER_BASE_QUEUE_NAME") || getenv("CAF_WORKER_NAME") || "worker") + "-invalid",
```
dermot-hardy: For the Document Workers I think it would be preferable to fallback to the fallback logic in the code:

```suggestion
    invalidQueue: getenv("CAF_WORKER_INVALID_QUEUE") || undefined,
```

dermot-hardy (changes requested):

dermot-hardy (approved):

PR #545, opened 2025-12-04
US1059166: Database Creation: Allow tablespace to be specified

raghavendran-gopalakrishnan:
https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=1059166


michael-bryson (approved):

release-notes-10.1.0.md
```diff
@@ -0,0 +1,10 @@
+!not-ready-for-release!
+
+#### Version Number
+${version-number}
+
+#### New Features
+- US1059166: Database Creation: Allow tablespace to be specified  
+  `-db.tablespace` optional argument can be now be passed to specify the tablespace for the Job service database creation. If not set, the database default tablespace will be used.
```
michael-bryson: The `JOB_SERVICE_DATABASE_TABLESPACE` env var can also be used which is what we use in the product.

michael-bryson (changes requested):

release-notes-10.1.0.md
raghavgopal: [Database creation](https://github.com/JobService/job-service/blob/develop/job-service-db/README.md) seems to be independent of the main service. It looks to me that in the product, we are just using the `_DATABASE_* ` env vars for connecting to the db. Please correct me if I miss the point, thanks.

michael-bryson: I am only pointing out that the env var can be used as well so I think you could call it out as well in the release notes and other documentation where appropriate.

michael-bryson (approved):

job-service-db/README.md
```diff
@@ -36,6 +40,7 @@ then run the image using:
 	-db.name jobservice \
 	-db.user postgres \
 	-db.pass root
```
dermot-hardy, line 42: Line continuation character missing

PR #561, opened 2026-03-11
US1071286: Update to `json-schema-validator` v3.0.0

anshuman0799:
https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=1071286

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -156,67 +156,72 @@
             <dependency>
                 <groupId>com.github.cafapi.common</groupId>
                 <artifactId>bootstrapconfig-system</artifactId>
-                <version>4.0.1-694</version>
+                <version>4.1.0-999</version>
             </dependency>
             <dependency>
                 <groupId>com.github.cafapi.common</groupId>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -1202,6 +1212,11 @@
                 <artifactId>jackson-databind</artifactId>
                 <version>3.1.0</version>
             </dependency>
+            <dependency>
+                <groupId>tools.jackson.dataformat</groupId>
+                <artifactId>jackson-dataformat-yaml</artifactId>
+                <version>3.0.3</version>
```
dermot-hardy: Why would you be introducing this at `3.0.3` when the other Jackson 3.x components are already using `3.1.0`?

dermot-hardy (changes requested):

pom.xml
anshuman0799: Addressed

dermot-hardy (approved):

